PR #3, opened 2015-06-03
Added methods to Graph for obtaining an iterator yielding mutable access to all Node/Edge weights.

mitchmindtree:
I came across a use-case for this in the `dsp-chain` crate. We need to visit every node (in no particular order) with mutable access in order to pre-allocate sample buffers given some stream format. Forgive me if there's already a better way to do this!

Also added a .gitignore (hope you don't mind!).

Also incremented the version in Cargo.toml in preparation for a `cargo publish` in the case that the PR gets merged.


bluss:
Hi. I assume you want to access the node and edge “weights” i.e the **N** and **E** values? Please tell me if this is wrong.

I don't want to give this access because it allows invalidating the whole datastructure if you swap positions of elements inside the node or edge slices. If I give this up in a safe function, we lose the possibility of taking care of a known consistent state in future optimizations of Graph.

You can already access all the weights, if you use `node_weight_mut` and `edge_weight_mut`. If you need a high performing version, I would suggest we add iterators for mutable access to all node weights and all edge weights. The iterators would not permit mutable access to the Node and Edge structs themselves.

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 name = "petgraph"
-version = "0.1.5"
+version = "0.1.6"
```
bluss: I think it's best to save this until I release a new version. I will do directly after this is merged, I promise.

mitchmindtree:
Ahh of course, that makes sense, I hadn't considered swaps, etc!

Perhaps an `all_node_weights_mut` method that returns an `AllNodeWeightsMut` Iterator yielding `&mut N`s? I'll make some changes and grab your thoughts again :+1:

I'll also revert my over-eager version increment :)

mitchmindtree:
Ok, I just replaced the `raw_nodes_mut` and `raw_edges_mut` methods with:

- `all_node_weights_mut`
- `all_edge_weights_mut`

Along with their respective Iterators:

- `AllNodeWeightsMut`
- `AllEdgeWeightsMut`

Let me know your thoughts when you get a chance :+1:

bluss:
I'll look closer at this soon. I prefer a different name maybe node_weights_mut? I also will improve the iterators with size hints etc.. But minor things I'll fixup myself, don't need to do it.

bluss:
Thanks, I'm going to include this with some cleanup. Excuse my pedantry (if any).

Can I add you to the authors?

bluss:
Done thank you!

mitchmindtree:
> maybe node_weights_mut?

Good call!

> Excuse my pedantry (if any).

Your pedantry probably contributes to why I love petgraph so much!

> Can I add you to the authors?

T'would be my pleasure!

Thank you!

PR #8, opened 2015-08-16
Update for well-formedness (Rust RFC 1214)

bluss:
Update for well-formedness (Rust RFC 1214)

https://github.com/rust-lang/rfcs/pull/1214


bluss:
cargo-crusader says dependents don't break — so we go for a minor version bump

PR #9, opened 2015-10-01
Topological order visitor

bluss:


bluss:
@mitchmindtree This feature landed at least, so it's a small improvement to incremental cycle detection (but you're right, no nice API to wrap the cycle detect part). In 0.1.14.

mitchmindtree:
Nice one, thanks @bluss :+1: Will adapt to this in daggy next I get the chance.

PR #12, opened 2015-10-12
Add a `Dag` type - a directed acyclic graph data structure.

mitchmindtree:
I thought it might make sense to consolidate the `Dag` type with the rest of petgraph, considering it's really just a wrapper around the `Graph` type. Thoughts?

No problems at all though if you think it's best to keep them separate, I'm fine either way :+1:

Edit: Ahh, just saw your reply on reddit. I'll leave this here for now in case you think it's fine to add, otherwise feel free to close if you'd prefer to re-evaluate in the future when you have more time (or if you just don't want it in petgraph) :+1:


bluss:
Yep, as I said, I don't want to be the bottleneck for this, much better if the author (you) develops it without having to go through me.

Issue #16, opened 2015-11-09
Iterator over the all edges of a GraphMap


contradictioned:
If I don't miss anything it is only possible to call `nodes()` to get an iterator over all nodes of a graph or ` `edges(from: N)` in order to get the edges incident to N.

Is there a special reason, that there is no iterator over *all* edges of a graph (in no specified order)?

mitchmindtree:
@contradictioned there is the [`raw_edges` method](http://bluss.github.io/petulant-avenger-graphlibrary/doc/petgraph/graph/struct.Graph.html#method.raw_edges) - you could use that to produce a slice upon which you could call `.iter()` to produce an iterator over all edges. Does that suit your need?

contradictioned:
@mitchmindtree unfortunateley this does not exist for a `GraphMap`...

mitchmindtree:
Agh, apologies.

contradictioned:
No problem, this hint was inspirational :D

contradictioned:
Solved by #18

PR #17, opened 2015-11-09
Adding raw access to the internal data structure of GraphMap.

contradictioned:
This is analogous to the implementation in `graph.rs`.

Reason for this PR: #16

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -241,6 +241,18 @@ impl<N, E> GraphMap<N, E> where N: NodeTrait
     {
         self.edges.get_mut(&edge_key(a, b))
     }
+
+    /// Access the internal nodes HashMap.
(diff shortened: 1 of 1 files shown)


bluss:
Not sure we should expose "raw" versions here. The Graph type is somewhat of a special case being relatively open to inspection of the whole implementation of the data structure. Something corresponding to the existing `nodes` would be nice.

contradictioned:
You mean like:

    fn all_edges<'a>(&'a self) -> Edges<'a, N, E>

which would return an iterator over ((N,N), E) ?

I'll try that :)

bluss:
Yes. Not so fond of the all_edges name, but I can't find a better one — just edges is taken.

contradictioned:
Instead of this, #18 is better.

PR #18, opened 2015-11-11
Iterator over all edges in a GraphMap.

contradictioned:
Added an iterator for all edges in a GraphMap.
Added test for this iterator.

Based on the discussion in #17

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -246,9 +246,9 @@ impl<N, E> GraphMap<N, E> where N: NodeTrait
     /// Return an iterator over all edges of the graph with their weight in arbitrary order.
     ///
     /// Iterator element type is `((N,N), &'a E)`
-    pub fn all_edges<'a>(&'a self) -> EdgeIter<'a, N, E>
+    pub fn all_edges<'a>(&'a self) -> AllEdges<'a, N, E>
     {
-        EdgeIter {
+        AllEdges {
             inner: self.edges.iter()
         }
     }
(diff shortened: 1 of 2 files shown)


src/graphmap.rs
```diff
@@ -327,6 +338,23 @@ impl<'a, N, E> Iterator for Edges<'a, N, E>
     }
 }
 
+pub struct EdgeIter<'a, N, E: 'a> where N: 'a + NodeTrait {
+    pub inner: HashmapIter<'a, (N, N), E>
```
bluss: Ah this should not be `pub`

src/graphmap.rs
```diff
@@ -327,6 +338,23 @@ impl<'a, N, E> Iterator for Edges<'a, N, E>
     }
 }
 
+pub struct EdgeIter<'a, N, E: 'a> where N: 'a + NodeTrait {
```
bluss: The struct should be named after the method that creates it.

bluss:
Thanks! Just fix up some small things and it's ready to go

contradictioned:
Btw: I made the inner iterator `pub` such that it fits the style of `Edges` [some lines above](https://github.com/bluss/petulant-avenger-graphlibrary/blob/master/src/graphmap.rs#L302-L307); maybe these should also be not-pub then.

bluss:
What in the world, that can't be intentional. De-pubbing edges will be put on the todolist for breaking changes.

bluss:
Thank you

contradictioned:
You're welcome :)

bluss:
Published in 0.1.13.

PR #30, opened 2015-11-25
Add node_capacity and edge_capacity methods to Graph and GraphMap

mitchmindtree:


bluss:
Thanks. Why are these needed? We have no reserve :)  (And I'm a bit unsure that it's needed)

mitchmindtree:
It might be useful if another buffer needs to be reserved with a capacity dependent upon the capacity with which some Graph was initialised.

I've just come across a case where this would be nice, as I don't have the value used to initialise the Graph's capacity available locally, though the buffer i'm reserving should have the same capacity as the Graph's node Vec (the buffer will be re-used and it is possible that the Graph will grow closer to its capacity between re-use, and i'd prefer not to have to re-allocate the buffer each time the Graph grows a bit bigger).

bluss:
ok. I guess we will end up with reserve later too then.

I think I'd prefer the paired style of with_capacity, so `fn capacity(&self) -> (usize, usize)`, isn't that ok?  (The eventual reserve would be the same).

mitchmindtree:
Ah yes, I guess that's ok - My decision to separate them was influenced by the separation of `node_count` and `edge_count`, but I see how it might make more sense to follow `with_capacity`.

Hmmm are you sure that's a good idea for `reserve`? I guess it would be ok for a user to just pass `0` if they did not need to reserve any more space for either of the `Vec`s.

Bout to go for a coffee but can do make the fix when I get back :+1:

bluss:
We don't need to add reserve now if we don't need it.

bluss:
Thanks!

Remember there is no notification for just newly pushed commits, needs a comment on the PR to draw attention to it. No big deal of course.

mitchmindtree:
No worries! I wasn't in too much of a rush for this one and forgot after waiting for trav heheh :smile_cat:

PR #39, opened 2015-12-02
neighbors().detach(), WalkNeighbors, and misc changes in preparation for 0.2; Part IV

### LINKED ISSUE #27: Breaking changes for version 0.2
- [x] Fix #19
- [x] Fix element type of `GraphMap::all_edges()`
-  ~~Look at panic vs return option behavior of Graph methods?~~
- [x] ~~Remake Graph's edges() iterator methods, maybe #13, also #39~~ Actual change is #39 
- [x] Remove MinScored?
- [x] Remove IndexType::one, zero.
- [x] Fix neighbors #31 
- [x] Mark IndexType unsafe
- [x] Revise ESI and DEI for Graphmap's iterators

.

- ~~GraphMap::add_node, change return type to `()`, maybe?~~
- [x] GraphMap::add_edge, change return type to `Option<E>`
- [x] ~~Rename~~ Remove Graph::edges_both to edges_undirected
- [x] Rename Graph::without_edges -> externals
- [x] Undirected, Directed should be enums.

bluss:
Introduce `.neighbors(a).detach()` and so on to be the replacement for `walk_edges_directed` which will be deprecated.

Closes #27

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,6 +19,34 @@ __ http://bluss.github.io/petulant-avenger-graphlibrary/
 Recent Changes
 --------------
 
+- 0.2.0
+
+  - New Features
+
+    - Add Graph::neighbors().detach() to step edges without borrowing.
+      This is more general than, and replaces now deprecated
+      walk_edges_directed. (#39)
+    - Implement Default for Graph, GraphMap
+    - Add method EdgeDirection::opposite()
+
+  - Breaking changes
+
+    - Graph::neighbors() for undirected graphs and Graph::neighbors_undirected
+      for any graph now visit self loop edges once, not twice. (#31)
+    - Renamed Graph::without_edges to Graph::externals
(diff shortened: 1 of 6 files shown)


bluss:
@mitchmindtree This is my draft so far of small parts that are Walker-related. I know it's not much, but it's enough to explore the area. neighbor and edges iterators stay, I want to have them and I don't want to break the single feature that every Graph user uses; I think the .detach() model is nice and logical.

Since this part doesn't need any breaking changes, it doesn't need to be elaborated more before 0.2! Will post an update on the stable indices graph for comments soon..

mitchmindtree:
Nice! I think `.detach()` looks fine and is a solid compromise :+1:

bluss:
Thank you for the feedback! Chooo chooo here comes the 0.2.0 train. Just a lot of cruft gone, but the future is bright.

PR #42, opened 2015-12-02
StableGraph, a graph where indices don't invalidate on deletion

### LINKED ISSUE #38: Add StableGraph / a stable indices after deletion graph

bluss:
Closes #38


bluss:
@mitchmindtree This is the draft for “StableGraph”, i.e. Graph but it doesn't invalidate unrelated indices on deletion.

It's using the Option enum, and @eddyb's device for tracking the free slots via a linked chain of indices. We don't need an usize in the enum since we have lots of free index fields in the Node and Edge structs.

This is a bit less clean that I wanted to show it, but it's mostly a full featured directed graph.

I don't know who to ask about the tricky issues of creating a nice and versatile interface, ideally this would be a parameter in graph: `Graph<N, E, Ty, Ix, StableIndices>`, but there seems to be no way to do this in a nice and composable way.

- We want an internal trait for the backend part, but rust requires this to be public,
- We need to duplicate type parameters to simulate HKT, which for a four-parameterized struct is bad
- Abusing private types and public type aliases could work, but it's not composable (can't be re-encapsulated by for example daggy).

So instead of this "interior genericity" of being able to switch storage policy, it looks like we have to(?) use external methodology: Graph and StableGraph can implement the same traits.

Relevant to issue RustAudio/dsp-chain/issues/103

mitchmindtree:
Nice :+1: hopefully one day we can work out a nice way of working something like a `NodeIndices` type param into `Graph`, but until then this should be very useful.

I guess now I'd do well to make a `StableDag` in daggy that uses this. Tis a pity to have to make a whole new type, but it will be useful.

bluss:
I got pretty far into trying out the selectable backend approach before I went with this instead, see this comment https://github.com/bluss/petulant-avenger-graphlibrary/issues/38#issuecomment-162184198

PR #50, opened 2015-12-11
impl Clone for iterators & document edge order

### LINKED ISSUE #49: Investigate what we can guarantee about edge order
W.r.t order of insertion vs order in neighbors()

bluss:
Fixes #49

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ matrix:
       env:
        - FEATURES='test quickcheck'
        - BENCH=1
+branches:
+  only:
+    - master
 script:
   - |
       cargo build --verbose --features "$FEATURES" &&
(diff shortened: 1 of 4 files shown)


bluss:
cc @mitchmindtree This is the order that's always present. Removals don't affect order of the remaining edges (in the neighbors() iteration).

mitchmindtree:
Ah yep, I was using [this](https://github.com/bluss/petulant-avenger-graphlibrary/pull/50/files#diff-a0c94d0c1540ede7afa6f3cbdb8c8948R668) as a heuristic for the instantiation order of widgets in conrod, but ended up opting to explicitly track the order as the heuristic didn't work out for the use case in conrod.

bluss:
Aha I see, what behavior would you need? I'm not sure much else could be provided. The edge lists are basically singly linked lists, so we add by appending in front.

mitchmindtree:
Yeah I think it's probably not petgraph's issue to worry too much about ordering etc, I think what you've done here by explicitly stating the behaviour in the docs is plenty enough :+1:

bluss:
ok, great. :cherries:

PR #58, opened 2016-02-22
visit.rs typo

BurntPizza:
Noticed this little copy-paste error while browsing.
There's also calling the Bfs queue field 'stack' but I'm sure you're aware of that.

### DIFF
--- a/src/visit.rs
+++ b/src/visit.rs
@@ -540,7 +540,7 @@ impl<N, VM> Bfs<N, VM>
         }
     }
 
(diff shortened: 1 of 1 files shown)


bluss:
Thank you. The `stack` thing could actually be changed as a breaking change.

Since this small fix would conflict with a larger WIP branch I am working on, I will decline to merge it, but thanks for bringing this to attention anyway!

BurntPizza:
No problem. I saw that it conflicted on the other branch so I kind of expected that.

PR #60, opened 2016-04-04
Fix Kosaraju

Kha:
I'm afraid my previous fix wasn't quite sufficient. Unfortunately, the only test case I have contains about 10,000 nodes...

### DIFF
--- a/src/algo.rs
+++ b/src/algo.rs
@@ -11,6 +11,7 @@ use super::{
     Graph,
     Directed,
     Undirected,
+    EdgeDirection,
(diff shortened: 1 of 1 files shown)


bluss:
Thank you! This is going to be a part of PR #61. Keep this PR open, it should automatically close once #61 merges I think..

Kha:
Ooh, that is a very nice test setup!

bluss:
Released 0.2.4 with the fix. Thanks a lot for reporting & fixing.

PR #62, opened 2016-04-04
Add SCC condensation algorithm

Kha:
I'm not sure whether the API is reasonable, and the test case isn't very satisfying yet... but at the very least I'm successfully using the implementation in my project :) .

### DIFF
--- a/src/algo.rs
+++ b/src/algo.rs
@@ -184,6 +184,45 @@ pub fn scc<N, E, Ty, Ix>(g: &Graph<N, E, Ty, Ix>) -> Vec<Vec<NodeIndex<Ix>>>
     sccs
 }
 
+/// Condense every strongly connected component into a single node and return the result.
+///
+/// If `make_acyclic` is true, self-loops and multi edges are ignored, guaranteeing that
+/// the output is acyclic.
+pub fn condensation<N, E, Ty, Ix>(g: Graph<N, E, Ty, Ix>, make_acyclic: bool) -> Graph<Vec<N>, E, Ty, Ix>
+    where Ty: EdgeType,
+          Ix: IndexType,
+{
+    let sccs = scc(&g);
+    let mut condensed: Graph<Vec<N>, E, Ty, Ix> = Graph::with_capacity(sccs.len(), g.edge_count());
+
+    // Build a map from old indices to new ones.
+    let mut node_map = vec![NodeIndex::end(); g.node_count()];
+    for comp in sccs {
+        let new_nix = condensed.add_node(Vec::new());
+        for nix in comp {
+            node_map[nix.index()] = new_nix;
+        }
+    }
+
+    // Consume nodes and edges of the old graph and insert them into the new one.
+    let (nodes, edges) = g.destructure();
+    for (nix, node) in nodes.into_iter().enumerate() {
+        condensed.node_weight_mut(node_map[nix]).unwrap().push(node.weight);
+    }
+    for edge in edges {
+        let source = node_map[edge.source().index()];
+        let target = node_map[edge.target().index()];
+        if make_acyclic {
+            if source != target {
+                condensed.update_edge(source, target, edge.weight);
+            }
+        } else {
+            condensed.add_edge(source, target, edge.weight);
+        }
+    }
+    condensed
+}
+
 /// Return the number of connected components of the graph.
 ///
 /// For a directed graph, this is the *weakly* connected components.
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -883,6 +883,12 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
         &self.edges
     }
 
+    /// Return the internal arrays, consuming the graph.
+    pub fn destructure(self) -> (Vec<Node<N, Ix>>, Vec<Edge<E, Ix>>)
+    {
+        (self.nodes, self.edges)
+    }
+
     /// Accessor for data structure internals: the first edge in the given direction.
     pub fn first_edge(&self, a: NodeIndex<Ix>, dir: EdgeDirection) -> Option<EdgeIndex<Ix>>
     {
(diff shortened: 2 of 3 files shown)


bluss:
This is great! I won't review this today though.

src/algo.rs
```diff
@@ -177,6 +177,45 @@ pub fn scc<N, E, Ty, Ix>(g: &Graph<N, E, Ty, Ix>) -> Vec<Vec<NodeIndex<Ix>>>
...
+/// the output is acyclic.
+pub fn condensation<N, E, Ty, Ix>(g: Graph<N, E, Ty, Ix>, make_acyclic: bool) -> Graph<Vec<N>, E, Ty, Ix> where
+    Ty: EdgeType,
+    Ix: IndexType,
+{
+    let sccs = scc(&g);
+    let mut condensed: Graph<Vec<N>, E, Ty, Ix> = Graph::with_capacity(sccs.len(), g.edge_count());
+
+    // Build a map from old indices to new ones.
+    let mut node_map = vec![NodeIndex::end(); g.node_count()];
```
bluss: This and the tarjan scc makes me think petgraph should offer a vecmap that's indexed by NodeIndex / EdgeIndex. It would be a simple convenience, but it comes up all the time.

src/algo.rs
```diff
@@ -177,6 +177,45 @@ pub fn scc<N, E, Ty, Ix>(g: &Graph<N, E, Ty, Ix>) -> Vec<Vec<NodeIndex<Ix>>>
     sccs
 }
 
+/// Condense every strongly connected component into a single node and return the result.
+///
+/// If `make_acyclic` is true, self-loops and multi edges are ignored, guaranteeing that
+/// the output is acyclic.
+pub fn condensation<N, E, Ty, Ix>(g: Graph<N, E, Ty, Ix>, make_acyclic: bool) -> Graph<Vec<N>, E, Ty, Ix> where
```
bluss: just style, but where should be hanging on the next line. Maybe the library is not 100% consistent with this.

src/graph.rs
```diff
@@ -883,6 +883,12 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
         &self.edges
     }
 
+    /// Return the internal arrays, consuming the graph.
+    pub fn destructure(self) -> (Vec<Node<N, Ix>>, Vec<Edge<E, Ix>>)
```
bluss: This is my main hold up, since it seems to make the data structure more naked (less encapsulation).

Yet that's not really the case, we could make Node and Edge vectors from any graph representation. The existing raw_nodes etc accessor is a real encapsulation blocker in comparison. So this is good.

src/algo.rs
Kha: Yes, that indeed sounds like a nice abstraction.

bluss:
Thanks!

src/graph.rs
Kha: I should mention that I wasn't even sure whether `condensation` should consume the graph. An implementation with signature `&Graph<N, E> -> Graph<Vec<&N>, &E>` may be equally interesting (on the other hand, it could be obtained by composing the consuming implementation with a function `&Graph<N, E> -> Graph<&N, &E>`...).

bluss: I was wondering about that, I'm assuming you needed it in the consuming form. We can't really offer both in the same function unfortunately..

src/algo.rs
```diff
@@ -184,6 +184,45 @@ pub fn scc<N, E, Ty, Ix>(g: &Graph<N, E, Ty, Ix>) -> Vec<Vec<NodeIndex<Ix>>>
     sccs
 }
 
+/// Condense every strongly connected component into a single node and return the result.
+///
+/// If `make_acyclic` is true, self-loops and multi edges are ignored, guaranteeing that
+/// the output is acyclic.
+pub fn condensation<N, E, Ty, Ix>(g: Graph<N, E, Ty, Ix>, make_acyclic: bool) -> Graph<Vec<N>, E, Ty, Ix>
```
Kha: I just noticed a small issue with the signature (slightly too late, I guess :smile:): Ignoring self-loops and multi edges are really two orthogonal tasks. In particular, multi edges do not affect acyclicity, of course.

I mainly introduced the parameter because, in my use case, I have unweighted edges and am not interested in self-loops or multi edges. On the other hand, I also wanted to provide a way to preserve all weights from the input graph.

bluss: It's not too late until the next version is released. Breaking changes in master is no problem.

bluss: What I think is a bit confusing in general is to preserve all edges of a component that was condensed. They turn into self-loops right?

src/algo.rs
```diff
@@ -184,6 +184,45 @@ pub fn scc<N, E, Ty, Ix>(g: &Graph<N, E, Ty, Ix>) -> Vec<Vec<NodeIndex<Ix>>>
...
+        let new_nix = condensed.add_node(Vec::new());
+        for nix in comp {
+            node_map[nix.index()] = new_nix;
+        }
+    }
+
+    // Consume nodes and edges of the old graph and insert them into the new one.
+    let (nodes, edges) = g.destructure();
+    for (nix, node) in nodes.into_iter().enumerate() {
+        condensed.node_weight_mut(node_map[nix]).unwrap().push(node.weight);
```
bluss: By the way, node_weight().unwrap() is the same as using indexing notation (same for mut).

`condensed[node_map[nix]].push(node.weight)`  thus works.

src/algo.rs
bluss: We could either add more arguments to the function, or if it should be extensible to other algo options in the future, use a more general kind of argument that carries all the options.

src/graph.rs
Kha: I suppose both variants would work for my use case, the consuming one just felt more natural (and, apparently, more general).

src/algo.rs
Kha: Yes, edges inside a component are turned into self-loops. I'm not sure even a two-argument signature would be needed for actual use cases. If you think that self-loops are generally uninteresting, we can just discard them and rename `make_acyclic` to `ignore_multi_edges`.

bluss: I think it's fine as it is now.

PR #64, opened 2016-04-08
Added single node topological traversal and quickcheck tests.

### LINKED ISSUE #63: Topological iterator for subgraph.
It would be nice if there was way to traverse directed graph topologically starting from certain node.

WaDelma:
Resolves #63

### DIFF
--- a/tests/quickcheck.rs
+++ b/tests/quickcheck.rs
@@ -622,7 +622,7 @@ fn sub_topo() {
             }
         }
         let mut index = 0;
-        let mut topo = SubTopo::from_node(&gr, graph_index);
+        let mut topo = SubTopo::new(&gr, graph_index);
         while let Some(nx) = topo.next(&gr) {
             gr[nx] = index;
             index += 1;
@@ -630,7 +630,7 @@ fn sub_topo() {
 
         let mut order = Vec::new();
         index = 0;
-        let mut topo = SubTopo::from_node(&gr, graph_index);
+        let mut topo = SubTopo::new(&gr, graph_index);
         while let Some(nx) = topo.next(&gr) {
             order.push(nx);
             assert_eq!(gr[nx], index);
@@ -644,7 +644,7 @@ fn sub_topo() {
 
         {
             order.clear();
-            let mut topo = SubTopo::from_node(&gr, graph_index);
+            let mut topo = SubTopo::new(&gr, graph_index);
             while let Some(nx) = topo.next(&gr) {
                 order.push(nx);
             }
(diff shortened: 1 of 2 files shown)


bluss:
This algorithm should work well. (I've never read about it, I wonder if it's the usual way to do it?)

I'm reluctant to add these extra datastructures to Topo. It's better as a separate walker. You can name it SubgraphTopo if you want. Or SubTopo or so?

WaDelma:
Yeah I was also iffy about adding more stuff to `Topo`, but splitting is also not ideal as well as it creates code dublication.    
I tried to search if there was some other way to do it, but didn't find anything.

tests/quickcheck.rs
```diff
@@ -453,3 +457,194 @@ fn graph_sccs() {
     }
     quickcheck::quickcheck(prop as fn(_) -> bool);
 }
+
+#[derive(Debug, Clone)]
+struct DAG<N: Default + Clone + Send + 'static>(Graph<N, ()>);
+
+impl<N: Default + Clone + Send + 'static> quickcheck::Arbitrary for DAG<N> {
```
bluss: We don't need to fix it now, but N should probably just be Arbitrary and generated by qc.

bluss:
Quickcheck too, that's nice!

Can we cut down the number of constructors for SubTopo?

tests/quickcheck.rs
WaDelma: Well the nodes value is used by the quicktest for both generic `Topo` and `SubTop` currently.

WaDelma:
Shoulde the constructor then be just named `new` and take the node as parameter?

bluss:
from node is good, then we can expand later if it's needed.

WaDelma:
Well I already renamed it :P Will I change it back?

bluss:
please use from_node

bluss:
Just for future reference, github does not notify for new commits, needs comments on the PR.

bluss:
Oops, sorry, my merging of the other PR conflicted with this branch. Didn't think of it.

bluss:
Hm. This algorithm really has some interesting behavior for cyclic graphs. Unlike Topo, it will traverse nodes that participate in a cycle, where the whole cycle is reachable from the start node. That's a bit problematic. Can we solve that with documentation alone?

WaDelma:
Rebased.    
Yeah should add that to documentation.

WaDelma:
Whoops

WaDelma:
I added note that cycles are included in `SubTopo`.

bluss:
Thank you! Let's gooo. I'll look at this locall and maybe edit something, then petgraph is ready for a new release.

PR #70, opened 2016-05-01
fixed some clippy warnings

llogiq:
This is mostly minor readability stuff, the code is already quite nice. :+1:

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -1,5 +1,4 @@
-//! `Graph<N, E, Ty, Ix>` is a graph datastructure using an adjacency list 
-//! representation.
+//! `Graph<N, E, Ty, Ix>` is a graph datastructure using an adjacency list representation.
 
 use std::cmp;
 use std::fmt;
@@ -542,7 +541,9 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
     /// endpoint in the displaced node.
     pub fn remove_node(&mut self, a: NodeIndex<Ix>) -> Option<N>
     {
-        if self.nodes.get(a.index()).is_none() { return None }
+        if self.nodes.get(a.index()).is_none() {
+            return None
+        }
         for d in &DIRECTIONS {
             let k = *d as usize;
 
@@ -923,7 +924,7 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
     /// - `Directed`, `Incoming`: All edges to `a`.
     pub fn walk_edges_directed(&self, a: NodeIndex<Ix>, dir: EdgeDirection) -> WalkEdges<Ix>
     {
-        let first_edge = self.first_edge(a, dir).unwrap_or_else(EdgeIndex::end);
+        let first_edge = self.first_edge(a, dir).unwrap_or(EdgeIndex::end());
         WalkEdges { next: first_edge, direction: dir }
     }
 
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -1,4 +1,5 @@
-//! `Graph<N, E, Ty, Ix>` is a graph datastructure using an adjacency list representation.
+//! `Graph<N, E, Ty, Ix>` is a graph datastructure using an adjacency list 
+//! representation.
 
 use std::cmp;
 use std::fmt;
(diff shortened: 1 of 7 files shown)


src/graphmap.rs
```diff
@@ -451,14 +451,9 @@ impl<N, E> Default for GraphMap<N, E>
 /// in particular in the combination where the node type for
 /// `GraphMap` is something of type for example `Ptr(&Cell<T>)`,
 /// with the `Cell<T>` being `TypedArena` allocated.
+#[derive(Copy, Clone)]
```
bluss: This one is wrong, deriving inserts bounds on `T`. Is that a bug in clippy?

src/scored.rs
```diff
@@ -33,22 +33,15 @@ impl<K: PartialOrd, T> Ord for MinScored<K, T> {
...
-            // these are the NaN cases
-            if a != a && b != b {
-                Ordering::Equal
-            } else if a != a {
-            // Order NaN less, so that it is last in the MinScore order
-                Ordering::Less
-            } else {
-                Ordering::Greater
-            }
+        match b.partial_cmp(a) {
```
bluss: It doesn't matter much, but I don't think this one is wanted -- unless  I misremember, calling partial_cmp on floats is pretty terrible performance wise (and the explicit operators better).

bluss:
Thanks. There's lots of great small fixes there and some I don't care for. I don't agree with for example changing match to if let just because it's possible.

src/graphmap.rs
llogiq: Yep, that's a false positive. Though on a fairly uncommon use case.

src/scored.rs
llogiq: TIL. Interesting. I wonder why this is the case.

llogiq:
Always good to learn something new. Feel free to pick the changes you like (of if I find the time I'll roll back the unwanted ones).

Replacing `match` with `if let` is deemed a readability oomprovememt, because a) less syntactic elements to read and b) it implies one indentation less (which in a good number of cases leads to less line wrapping / over-long lines).

src/graph.rs
```diff
@@ -1,4 +1,5 @@
-//! `Graph<N, E, Ty, Ix>` is a graph datastructure using an adjacency list representation.
+//! `Graph<N, E, Ty, Ix>` is a graph datastructure using an adjacency list 
+//! representation.
```
bluss: no

src/graph.rs
```diff
@@ -498,12 +497,9 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
...
         if let Some(ix) = self.find_edge(a, b) {
-            match self.edge_weight_mut(ix) {
-                Some(ed) => {
-                    *ed = weight;
-                    return ix;
-                }
-                None => {}
+            if let Some(ed) = self.edge_weight_mut(ix) {
+                *ed = weight;
+                return ix;
```
bluss: Ok, I do come around to this being easier to read.

src/graph.rs
```diff
@@ -546,11 +542,8 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
     /// endpoint in the displaced node.
     pub fn remove_node(&mut self, a: NodeIndex<Ix>) -> Option<N>
     {
-        match self.nodes.get(a.index()) {
-            None => return None,
-            _ => {}
-        }
-        for d in DIRECTIONS.iter() {
+        if self.nodes.get(a.index()).is_none() { return None }
```
bluss: I'd prefer if this was not a oneliner but a regular if with line break after { and ; after return None.

src/graph.rs
```diff
@@ -930,7 +923,7 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
     /// - `Directed`, `Incoming`: All edges to `a`.
     pub fn walk_edges_directed(&self, a: NodeIndex<Ix>, dir: EdgeDirection) -> WalkEdges<Ix>
     {
-        let first_edge = self.first_edge(a, dir).unwrap_or(EdgeIndex::end());
+        let first_edge = self.first_edge(a, dir).unwrap_or_else(EdgeIndex::end);
```
bluss: no

src/graphmap.rs
```diff
@@ -117,7 +117,7 @@ impl<N, E> GraphMap<N, E>
 
     /// Add node `n` to the graph.
     pub fn add_node(&mut self, n: N) -> N {
-        self.nodes.entry(n).or_insert_with(|| Vec::new());
+        self.nodes.entry(n).or_insert_with(Vec::new);
```
bluss: no

src/visit.rs
```diff
@@ -340,7 +340,7 @@ pub trait GetAdjacencyMatrix : Graphlike {
     fn is_adjacent(&self, matrix: &Self::AdjMatrix, a: Self::NodeId, b: Self::NodeId) -> bool;
 }
 
-/// The **GraphMap** keeps an adjacency matrix internally.
+/// The `GraphMap? keeps an adjacency matrix internally.
```
bluss: typo

bluss:
Ok, now it has a full review. The unmentioned changes are all very good.

src/graph.rs
llogiq: I blame my editor :smile:

llogiq:
Glad you like them. :smile:

bluss:
Thanks!

llogiq:
You're welcome.

PR #74, opened 2016-08-02
Allow changing capacity of underlying Vecs in Graph

### LINKED ISSUE #73: Expose basic memory management methods in Graph's API
Hi,
I'm implementing genome assembler using `petgraph`, and i really miss functions that would allow me to actively change `Graph`'s capacity:

- `shrink_to_fit_nodes()`
- `shrink_to_fit_edges()`
- `reserve_nodes(usize)`
- `reserve_edges(usize)`

Often i find myself in the situation in which `Graph`'s capacity is orders of magnitude higher than what i'm actually storing in the graph (it happens when huge graph is getting aggressively pruned).
Likewise when i create such huge graphs i would like to have some control over allocation sizes, hence `reserve`s. 
Are there any specific reasons why these methods aren't available in the current API (as they are available in the underlying Vecs)? I would gladly implement these methods if you agreed on merging them.

Thanks for this awesome crate!

fuine:
As described in #73 . I'm not really sure if the style / layout of the commited code is on a par with petgraph's implementation so feel free to point out what should be corrected.

Fixes #73

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -387,6 +387,64 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
         (self.nodes.capacity(), self.edges.capacity())
     }
 
+    /// Reserves capacity for at least `additional` more nodes to be inserted in
+    /// the graph. Graph may reserve more space to avoid frequent reallocations.
+    ///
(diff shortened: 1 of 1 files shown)


bluss:
Summer's over!

fuine:
Thanks!

PR #77, opened 2016-09-28
Fix a bug in SubTopo

bluss:
It did not correctly limit the count of incoming edges to those only in
the subgraph. This was discovered by the existing quickcheck test
failing (so it was pretty rare).

Fix it by introducing another visitor map ("discovered") to keep track
of which nodes participate in the subgraph.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ debug = true
 
 [dependencies]
 fixedbitset = "0.1.0"
-quickcheck = { optional = true, version = "0.2.24" }
+quickcheck = { optional = true, version = "0.3" }
 
 [dev-dependencies]
 rand = "0.3"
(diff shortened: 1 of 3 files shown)


bluss:
cc @WaDelma There was a bug, the existing quickcheck test found it (it just failed very seldomly).

You have to run quickcheck sub_topo in a while loop in bash (exit on test failure) to find it.

Do you think this is the simplest fix? It needs another visit map (discovered).

WaDelma:
I haven't looked at it closely, but it would be good if the failing case would be added as a normal test too (especially when it only rarely fails with quickcheck).
I will look at it more closely later.

bluss:
Diff by ignoring whitespace https://github.com/bluss/petgraph/pull/77/commits/b52c10f89a8accaa4ed21db40c3d3dec6dbf40d0?w=1

bluss:
This makes the SubTopo eager, by eagerly discovering the whole subgraph on the first .next() call.

bluss:
Merging this so I can get on with other work on petgraph. Your revisit of this is very welcome, maybe there's a more efficient way to resolve it.

PR #82, opened 2016-09-30
Reduced subtopo test case.

WaDelma:
This reduces the previously failing test case for subgraph topological traversal.

I also realised that there isn't really better way to implement it lazily: The failing case can be generalized so that algorithm cannot know which node to choose when there is incoming edge from outside of the subgraph.

Now that the visit is eager is there any point for SubTopo?

I was thinking about it and realised that it would be nice if there were way to have views into the graph which would also work as graph. Basically have visit map that marks which nodes are in the view and then implement basic graph stuff so that only those nodes are taken into account.

This way you could run the normal topoglogical traversal. And it could be used with any other algorithms.

Then there could be method for turning view to actual graph.


bluss:
Thanks. I had the same thought! I haven't tried if it's feasible, but I think it is.

With this discussion we'll post a patched 0.2.x version but will consider removing SubTopo in favour of a more exciting subgraph / subset thing in the coming 0.3

bluss:
mockup using the new (0.3.x) graph traits

```rust
pub struct SubGraph<G>
    where G: Visitable
{
    graph: G,
    map: G::Map,
}

impl<G> GraphBase for SubGraph<G>
    where G: GraphBase + Visitable,
{
    type NodeId = G::NodeId;
    type EdgeId = G::EdgeId;
}

impl<'a, G: 'a> IntoNeighbors for &'a SubGraph<G>
    where G: IntoNeighbors + Visitable,
{
    type Neighbors = SubGraphFilter<'a, G, G::Neighbors>;
    fn neighbors(self, n: G::NodeId) -> Self::Neighbors {
        SubGraphFilter {
            map: &self.map,
            iter: self.graph.neighbors(n),
        }
    }
}

pub struct SubGraphFilter<'a, G, I>
    where G: 'a + Visitable,
{
    map: &'a G::Map,
    iter: I,
}

impl<'a, G, I> Iterator for SubGraphFilter<'a, G, I>
    where G: Visitable,
          I: Iterator<Item=G::NodeId>,
{
    type Item = I::Item;
        let map = self.map;
        (&mut self.iter).filter(|n| map.is_visited(n)).next()
    }
}

```

PR #83, opened 2016-10-01
Move `tarjan_scc` from tests to algo

### LINKED ISSUE #79: Lift the tarjan scc algorithm from tests into petgraph::algo
It needs to be generalized and then it's good to go

BurntPizza:
Fixes #79.
Would like thoughts on the API, is it suitably general? Or is there a better way?

Can also amend to replace `scc` with renamed `tarjan_scc`. Note that the APIs are slightly different.

### DIFF
--- a/src/algo.rs
+++ b/src/algo.rs
@@ -21,6 +21,7 @@ use scored::MinScored;
 use super::visit::{
     Visitable,
     VisitMap,
+    IntoNeighbors,
     IntoNeighborsDirected,
     IntoNodeIdentifiers,
     IntoExternals,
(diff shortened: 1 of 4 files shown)


BurntPizza:
Something I came across: should `GraphBase::NodeId` be `PartialEq`? If it was, the `NodeIndexable` bound could be removed I think, although the time complexity would suffer a bit (`NodeIndexable` allows the `node!` macro to be O(1)), so it probably isn't worth it.

bluss:
NodeIndexable is used instead of Visitable here, you could say. I'm happy that it looks like the existing traits handled that well

bluss:
What more traits to require of G::NodeId is a great question, I didn't consider it. Individual algos can place those requirements themselves, though, and until now none of them do.

Thanks for this contribution, it's solid, the changes from the test version to generalized are easy to read. And it looks like you already tested on StableGraph too, and the algorithm should work fine with a non-compact index space from the looks of it.

PR #88, opened 2016-10-02
Remove SubTopo

bluss:
Remove it, because it is eagerly exploring the whole subgraph anyway.

cc @WaDelma 

I haven't been able to write a nice subgraph restriction graph combinator yet, though.


bluss:
New try is welcome.

PR #103, opened 2016-10-19
WIP: .edge_entry() for Graph

bluss:


fuine:
I have a question regarding this API - is it possible to add methods that would mutate the edge in-place? I am unfamiliar with the exact underlying representation of edges, and so my question might be completely wrong. What i would like to do is  given `OccupiedEntry` change this edges target and/or source (change the target/source of the `Edge` to different `NodeIndex`, NOT mutate nodes themselves). Is this possible? If so i am willing to implement such functionality. If not is there any way to achieve this with current implementation or should i resort to adding completely new edge like so:
```
// assuming edge_idx, new_target are known at this time and are present in the graph
let w = graph.edge_weight(edge_idx).unwrap();
let (s, _) = graph.edge_endpoints(edge_idx).unwrap();
// remove first to assure that index of the new edge won't get changed upon removal
graph.remove_edge(edge_idx);
edge_idx = match graph.add_edge(s, new_target, w) {
    Some(e) => e,
    None => // do something accordingly
}
```
unwraps are used because i assume that edge_idx and new_target are already present in the graph

bluss:
While changing an edge's target or source is cheaper than re adding it entirely, I think the latter is about as easy.

If there is a nice way to write the update source in place, I'd accept it. Note that it requires walking g the linked lists of edges.

fuine:
Ok, i will most probably take a look at this, as i will be running a lot of such operations in the algorithm i'm implementing. I will let you know if i came with better solution than the one proposed

bluss:
Note that if your graph is dense enough (too many edges per node), none of the petgraph Graph data structures are a good fit. Maybe Csr, yes, but it's not really made for mutation. This was just a random thought.

By looking at `Graph::remove_edge` I think you can figure out how to do it. I guess `change_edge_links` needs to be split into two parts, incoming and outgoing.

fuine:
Sure, do you want me to include it in the `Entry` API or rather open a standalone PR for the `change_edge_links` method? Also i'm currenty using `petgraph` for sparse, large graph computations and it works really well.

bluss:
Hm I don't know. It doesn't look like it fits edge_entry perfectly when looking at its data structures. Yet I think it sounds like it fits there w.r.t workflow

TedDriggs:
@bluss I found myself looking for this exact API today when trying to make a graph that has `Vec<_>` for its edge weights. Apart from merge conflicts, is there anything else that has to be addressed for this to merge?

PR #104, opened 2016-10-21
Add support for computing dominators

fitzgen:
This is an implementation of the engineered ["Simple, Fast Dominance Algorithm"][0] discovered by Cooper et al.

This algorithm is `O(v^2)`, and therefore has slower theoretical running time than the Lenguaer-Tarjan algorithm (which is `O(e * log(v)`). However, Cooper et al found it to be faster in practice on control flow graphs of up to ~30,000 vertices. Additionally, the algorithm is much easer to understand than LT and, and easier to implement as well.

[0]: http://www.cs.rice.edu/~keith/EMBED/dom.pdf

### DIFF
--- a/src/algo/dominators.rs
+++ b/src/algo/dominators.rs
@@ -1,0 +1,265 @@
+//! # The Dominance Relation
+//!
+//! In a directed graph with a root node `R`, a node `A` is said to *dominate* a
+//! node `B` iff every path from `R` to `B` contains `A`.
+//!
+//! The node `A` is said to *strictly dominate* the node `B` iff `A` dominates
+//! `B` and `A != B`.
+//!
+//! The node `A` is said to be the *immediate dominator* of a node `B` iff it
+//! strictly dominates `B` and there does not exist any node `C` where `A`
+//! dominates `C` and `C` dominates `B`.
+
+use std::collections::{HashMap, HashSet};
+use std::hash::Hash;
+
+use visit::{DfsPostOrder, GraphBase, IntoNeighbors, Visitable, Walker};
+
+/// The dominance relation for some graph and root.
+#[derive(Debug, Clone)]
+pub struct Dominators<N>
+    where N: Copy + Eq + Hash
+{
+    root: N,
+    dominators: HashMap<N, N>,
+}
+
+impl<N> Dominators<N>
+    where N: Copy + Eq + Hash
+{
+    /// Get the root node used to construct these dominance relations.
+    pub fn root(&self) -> N {
+        self.root
+    }
+
+    /// Get the immediate dominator of the given node.
+    ///
+    /// Returns `None` for any node that is not reachable from the root, and for
+    /// the root itself.
+    pub fn immediate_dominator(&self, node: N) -> Option<N> {
+        if node == self.root {
+            None
+        } else {
+            self.dominators.get(&node).cloned()
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


bluss:
:heart:

I'm half inclined to just hit merge (looks like literate Rust code), but I'll do some comments inline.

src/dominators.rs
```diff
@@ -0,0 +1,277 @@
...
+                node: Some(node),
+            })
+        } else {
+            None
+        }
+    }
+}
+
+/// Iterator for a node's dominators.
+pub struct Iter<'a, G>
```
bluss: A good rule for iterators is that they are named after the function that created them (You'll find this is true of the `Iter` you find in libstd!). This one looks like it should be something closer to `Dominators`.

src/dominators.rs
```diff
@@ -0,0 +1,277 @@
...
+//! dominate `B` in turn.
+
+use std::collections::{HashMap, HashSet};
+use std::hash::Hash;
+
+use visit::{DfsPostOrder, GraphBase, IntoNeighbors, Visitable, Walker};
+
+/// The dominance relation for some graph and root.
+#[derive(Debug, Clone)]
+pub struct Dominators<G>
```
bluss: I think this should be `Dominators<N>` (where `N` is the `G::NodeId`).

The reason is that the `G` as used in this file is a *reference-to-graph* (due to how lifetimes + iterators + traits are limited in Rust, that's what we use).  We don't want to tie the Dominators struct to a borrow of a graph, when no borrowed data is actually needed.

src/dominators.rs
```diff
@@ -0,0 +1,277 @@
...
+        .enumerate()
+        .map(|(idx, &node)| (node, idx))
+        .collect();
+
+    // Maps a node's `post_order` index to its set of predecessors's indices
+    // into `post_order` (as a vec).
+    let idx_to_predecessor_vec =
+        predecessor_sets_to_idx_vecs(&post_order, &node_to_post_order_idx, predecessor_sets);
+
+    let mut dominators: Vec<_> = (0..length).map(|_| UNDEFINED).collect();
```
bluss: `let mut dominators = vec![UNDEFINED; length];` is short and sweet.

src/dominators.rs
bluss: Rust has that nice thing that `Dominators` would hold the graph borrowed, so that they are guaranteed to be valid together (because the graph can't change), but at the end of the day, the redundant borrowing decreases the flexibility of the program.

bluss:
What do you think about putting this under the algo submodule? Obviously still in its own file.

src/dominators.rs
fitzgen: I *always* do this and can never remember to use that form of `vec!`, not sure why it won't stick...

src/dominators.rs
fitzgen: Will do "DominatorsIter".

src/dominators.rs
fitzgen: Will do.

fitzgen:
> What do you think about putting this under the algo submodule? Obviously still in its own file.

SGTM

src/dominators.rs
bluss: That's the convetion from pre Rust 1.0-alpha or so? :smile:

fitzgen:
> ❤️

😸

PR #111, opened 2016-10-30
Clarify edge direction for GraphMap

sbjorn:
Clarify the direction of edges created by the `add_edge` function.


bluss:
Hm this does not seem so clear. In an undirected graph, the edges don't have a direction, so I'm not sure what this explanation says. It should say that for an undirected edge, the edge connects a and b.

sbjorn:
Duh, of course! It seems i documented `edge_key` rather than this function. Sorry!

bluss:
Thanks. Sorry for the delay, indeed.

PR #113, opened 2016-11-03
Implement Walker for &mut W where W: Walker

bluss:
Closes #205


bluss:
Breaking change additions have started, so here we go.

PR #114, opened 2016-11-13
Initial working version of the push-relabel algorithm for max-flow/min-cut.

jneem:
This isn't ready for merging, but I have a question about how to proceed: obviously, I want this to be generic over the type of edge weights, but I really want exact arithmetic and checked overflow (neither of which are provided by the `Measure` trait). The `num-traits` crate has traits for that -- are there any objections to add that as a dependency? Do you have any long-term thoughts about the right traits for weights in weighted graph algorithms?


bluss:
Hi! I think I would recommend having this algorithm in a separate crate. I feel I have a lot to do, and I can't offer much guidance. For the other open PR to petgraph, I've (un)fortunately already committed myself.

PR #118, opened 2016-12-31
Implement PartialEq for algo::Cycle

jmcomets:
I added in the `PartialEq` autogenerated implementation in order to compare the return value of `toposort`.


bluss:
Thank you. It's not always clear how to handle equality in errors, but the default one is ok here, even if it can lead to logic bugs.

Issue #28, opened 2015-11-24
Add a dense adjacency matrix graph


bluss:

jmcomets:
I'd like to look into this, do you have some tips on the various traits the graph should implement to be complete?

RaoulLuque:
Isn't this revolved in #173? :) @starovoid

bluss:
Great

PR #120, opened 2017-01-07
Initial version of an adjacency matrix based graph

jmcomets:
This version adds a `DenseGraph` struct, which implements the basic
graph functionality. The underlying structure is two weight maps and an
adjacency matrix.

The following traits are currently implemented:
- Default
- Index/IndexMut
- Clone

As well as basic node/edge addition.

The DenseMapping/IdGenerator structs allow fast indexing of weights,
combined with automatic index generation. This comes at the cost of
manual and regular resizing of a slice (`std::vec::Vec`). This slice can
probably be replaced with a `std::collections::HashMap` (since it uses
open-addressing under the hood), [stash](https://github.com/Stebalien/stash-rs) or even [vec_map](https://github.com/contain-rs/vec-map).

I was largely inspired from [src/graph.rs](https://github.com/bluss/petgraph/blob/master/src/graph.rs) for most of the implementation, so there's probably a bunch of documentation errors to fix.

__Important note:__ I'm aware there's still a bunch of work to be done and I'll probably need some pointers (#28) concerning the methodology for implementing neighbor iterators/visiting.

### DIFF
--- a/src/dense_graph.rs
+++ b/src/dense_graph.rs
@@ -1,0 +1,741 @@
+use std::collections::hash_set;
+use std::marker::PhantomData;
+use std::ops::{Index, IndexMut};
+use std::cmp;
+
+use dense_mapping::{DenseMapping, IdIterator};
+use fixedbitset::FixedBitSet;
+
+use {
+    Directed,
+    Direction,
+    EdgeType,
+    Incoming,
+    IntoWeightedEdge,
+    //Outgoing,
+    Undirected,
+};
+
+pub use graph::{
+    DefaultIx,
+    EdgeIndex,
+    IndexType,
+    NodeIndex,
+};
+
+use visit::{
+    GetAdjacencyMatrix,
+    GraphBase,
+    IntoNeighbors,
+    IntoNeighborsDirected,
+    IntoNodeIdentifiers,
+    NodeCount,
+    Visitable
+};
+
+pub use graph::{node_index, edge_index};
+
+type OptionVec<T> = Vec<Option<T>>;
+
+const INITIAL_RESIZE: usize = 10;
+
+fn ensure_len<T>(vec: &mut OptionVec<T>, min_len: usize) where T: Clone {
+    if vec.len() <= min_len {
+        let next_power_of_2_len = (min_len + 2) & !1;
+        vec.resize(cmp::max(next_power_of_2_len, INITIAL_RESIZE), None);
+    }
+}
+
+type Matrix = OptionVec<OptionVec<EdgeList>>;
+
+type EdgeList = hash_set::HashSet<usize>;
+type EdgeIter<'a> = hash_set::Iter<'a, usize>;
+
+pub struct DenseGraph<N, E, Ty = Directed, Ix = DefaultIx> {
+    matrix: Matrix,
+    nodes: DenseMapping<N>,
+    edges: DenseMapping<E>,
+    ty: PhantomData<Ty>,
+    ix: PhantomData<Ix>,
+}
+
+pub type DenseDiGraph<N, E, Ix = DefaultIx> = DenseGraph<N, E, Directed, Ix>;
+pub type DenseUnGraph<N, E, Ix = DefaultIx> = DenseGraph<N, E, Undirected, Ix>;
+
+impl<N, E, Ty, Ix> DenseGraph<N, E, Ty, Ix>
+    where Ty: EdgeType,
+          Ix: IndexType,
+          N: Clone,
+          E: Clone
+{
+    /// Create a new `DenseGraph` with estimated capacity.
+    pub fn with_capacity(nodes: usize, edges: usize) -> Self {
+        DenseGraph {
+            matrix: Matrix::with_capacity(nodes),
+            nodes: DenseMapping::with_capacity(nodes),
+            edges: DenseMapping::with_capacity(edges),
+            ty: PhantomData,
+            ix: PhantomData,
+        }
+    }
+
+    /// Remove all nodes and edges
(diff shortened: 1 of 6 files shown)


jmcomets:
For reference, I wrote an implementation using stash, available [here](https://github.com/jmcomets/petgraph/commit/9d5c2f871f302c1212907eff0924d8f62de8c872). It's significantly shorter and much easier to understand, although it does imply adding the stash dependency. I also suspect it's faster by looking at stash's source code.

bluss:
Thanks will look at this later. Don't add a stash dep. I suspect it's about the same as StableGraph's freelist thing

jmcomets:
I've just squashed everything, could you have a look? I'm planning to add a bunch of unit-tests, but the implementation should be fairly complete.

bluss:
thanks, the clean version is easy to read

src/dense_graph.rs
```diff
@@ -0,0 +1,741 @@
...
+
+pub use graph::{node_index, edge_index};
+
+type OptionVec<T> = Vec<Option<T>>;
+
+const INITIAL_RESIZE: usize = 10;
+
+fn ensure_len<T>(vec: &mut OptionVec<T>, min_len: usize) where T: Clone {
+    if vec.len() <= min_len {
+        let next_power_of_2_len = (min_len + 2) & !1;
```
bluss: I don't understand how this works.

I would probably elect to let the vec decide pace to grow, and just ask it to grow to the min_len, and fill it with None until its resulting capacity.

jmcomets:
Yeah I was expecting this...

In these use-cases, I like to grow vecs to the next power of two, instead
of simply growing to "len + 1". Thanks to this, the vec grows
exponentially, therefore performing much less allocations.

bluss:
That's the default allocation pattern for a vec, we don't need to reimplement that though.

jmcomets:
I've just run some benchmarks, and while looking at it, I noticed that I screwed up when writing this. It simply goes to the next *multiple* of 2, not the next power...

Here are my results:
```rust
// with:
// test dense_graph_add_edges_from_root ... bench:      15,340 ns/iter (+/- 988)
//
// without:
// test dense_graph_add_edges_from_root ... bench:      14,963 ns/iter (+/- 4,461)
#[bench]
fn dense_graph_add_edges_from_root(bench: &mut Bencher)
{
    bench.iter(|| {
        let mut gr: DenseGraph<(), ()> = DenseGraph::new();
        let a = gr.add_node(());

        for _ in 0..100 {
            let b = gr.add_node(());
            gr.add_edge(a, b, ());
        }   
    }); 
}

// with:
// test dense_graph_add_adjacent_edges ... bench:      40,118 ns/iter (+/- 26,898)
//
// without:
// test dense_graph_add_adjacent_edges ... bench:      34,736 ns/iter (+/- 28,208)
#[bench]
fn dense_graph_add_adjacent_edges(bench: &mut Bencher)
{
    bench.iter(|| {
        let mut gr: DenseGraph<(), ()> = DenseGraph::new();
        let mut prev: Option<NodeIndex> = None;
        for _ in 0..100 {
            let b = gr.add_node(());

            if let Some(a) = prev {
                gr.add_edge(a, b, ());
            }   

            prev = Some(b);
        }   
    }); 
}
```

And here's the patch:
```diff
@@ -41,8 +41,8 @@ const INITIAL_RESIZE: usize = 10;

fn ensure_len<T>(vec: &mut OptionVec<T>, min_len: usize) where T: Clone {
     if vec.len() <= min_len {
-        let next_power_of_2_len = (min_len + 2) & !1;
-        vec.resize(cmp::max(next_power_of_2_len, INITIAL_RESIZE), None);
+        let next_len = min_len + 1;
+        vec.resize(cmp::max(next_len, INITIAL_RESIZE), None);
     }
 }
```


I'll attach the benchmarks and the patch to this in a follow-up commit.

bluss:
Well that explains why I didn't understand that computation.

src/dense_graph.rs
```diff
@@ -41,8 +41,8 @@ const INITIAL_RESIZE: usize = 10;
 
 fn ensure_len<T>(vec: &mut OptionVec<T>, min_len: usize) where T: Clone {
     if vec.len() <= min_len {
-        let next_power_of_2_len = (min_len + 2) & !1;
-        vec.resize(cmp::max(next_power_of_2_len, INITIAL_RESIZE), None);
+        let next_len = min_len + 1;
+        vec.resize(cmp::max(next_len, INITIAL_RESIZE), None);
```
bluss: I'd use .reserve(1) and then fill with None up until its new capacity

bluss: oh reserve just enough to reach min_len I guess, but still fill up to capacity (which may be larger).

jmcomets: Looking at the [implementation of Vec](https://doc.rust-lang.org/src/collections/vec.rs.html#1170-1249), it seems that there's already the optimization of using reserve + fill when calling `resize`.

bluss: I've worked on that exact code, so I know it well.

jmcomets: Why do you want to call `reserve(1)` then? I can't tell the difference with calling `resize(next_len)`.

bluss: Use the vec's default growth algorithm instead of anything else

jmcomets:
I'm going to close this PR. I'm not satisfied with the benchs I've run on it. I have another version which is almost ready in [`adjacency-matrix-v2`](https://github.com/jmcomets/petgraph/tree/adjacency-matrix-v2), which uses more cache-friendly structures.

PR #123, opened 2017-02-04
Add index_twice_mut to StableGraph

christolliday:
Just copied straight from Graph, not sure if there's a way this should be shared


christolliday:
good point, I placed it above capacity() because that was the closest common method I could find, but it definitely shouldn't be seen first in the docs.

bluss:
Thanks for the help :)

PR #124, opened 2017-02-06
Generalize `IntoWeightedEdge` from `Copy` to `Clone`

pthariensflame:


bluss:
Hm why is this? The ix types are always Copy

pthariensflame:
It's a first step towards a plan to generalize the whole library in the same way.  Is that not desired?

bluss:
I don't think so, I went from clone to copy for `GraphBase::NodeId` in 0.3 or 0.4.

pthariensflame:
Oh, ok!  I'll close this then.  :)

PR #126, opened 2017-02-23
Adding retain for stable_graph, function returns removed weights

Rupsbant:


Rupsbant:
I'll use the same signature as Graph::retain_nodes. (With Frozen<Self>). The current interface is not general enough.

PR #129, opened 2017-03-06
AllEdges additions

### LINKED ISSUE #128: Mutable equivalent to `AllEdges`
I am not sure how to get an iterator with mutable access to all edge weights in a `GraphMap`. Is this even possible?

Binero:
Implemented a mutable equivalent of AllEdges (AllEdgesMut), and added some function overrides to make use of increased iterator performance the OrderMap iterators provide.

(closes #128)

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -3,7 +3,10 @@
 
 use std::cmp::Ordering;
 use std::hash::{self, Hash};
-use std::iter::Cloned;
+use std::iter::{
+    Cloned,
+    DoubleEndedIterator,
+};
 use std::slice::{
     Iter,
 };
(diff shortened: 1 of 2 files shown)


Binero:
Seems like there is some failing tests on the master branch in general. Not caused by this PR.

bluss:
Indeed, the tests are failing due to bug rust-lang/rust/issues/40235 which has been fixed.

bluss:
To be clear, the good performance of ordermap iteration is innate to its .next() method and not related to the other methods.

Binero:
Yes, but OrderMap is able to do a lot more than just getting the next() element faster. It can get count() etc in O(1) time.

bluss:
Thanks! There's nothing more to do than merge.

I'd reuse .next_back() in last but that's too trivial to hold us up. I spy another interesting warning in the travis log though, that's the next thing to investigate.

PR #130, opened 2017-03-06
Fix compilation issue in tests on nightly

jmcomets:
I noticed that a recent update to the nightly toolchain on Travis broke the compilation of `tests/graph.rs`. The issue is simply a naming clash because `node_index` was aliased to `n`, which was also used as a variable in one of the test bodies.


bluss:
I wanted to not fix this, since it is a bug in the nightly. It's already fixed upstream. It doesn't matter much either way, thanks for dedicating your time to it.

bluss:
Thanks. Nightly is now fixed, finally, so we don't need to work around it.

PR #142, opened 2017-04-17
Add opt-in serde support

generalelectrix:
Adds opt-in serde support for Graph and StableGraph, using serde's best practice for feature naming.  GraphMap support depends on OrderMap adding support for serde, so I've left that derive out for now.

I wrote one unit test to exercise this for Graph but it isn't making any meaningful assertions because Graph doesn't provide any simple way to compare two graphs (and really we'd want to be comparing the exact details of their internals anyway).  It could just as well be removed, since all we're doing is deriving traits and any bugs that we catch would almost certainly be Serde bugs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ rand = "0.3"
 odds = { version = "0.2.19" }
 itertools = { version = "0.5" }
 defmac = "0.1"
+serde_json = "*"
 
 [features]
 default = ["graphmap", "stable_graph"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,8 @@ debug = true
 fixedbitset = { version = "0.1.4" }
 quickcheck = { optional = true, version = "0.4", default-features = false }
 ordermap = { version = "0.2.2", optional = true }
+serde = { version = "0.9.13", optional = true }
+serde_derive = { version = "0.9.13", optional = true }
 
 [dev-dependencies]
 rand = "0.3"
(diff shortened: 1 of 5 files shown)


bluss:
Hi.

Don't you think using derived serialization will be very inflexible? Can't all graphs serialize into essentially the same format instead?

generalelectrix:
I'm assuming you mean to create one standard representation of a serialized graph such that one could serialize a Graph and deserialize it later as, say, StableGraph?

I'm personally not interested in that level of detail and don't have the time to commit to doing such a thing.  Here I just want the ability to pass these entities across process boundaries as well as persist them to disk and back again.  I'm also not sure how such a format would work for, say, StableGraph, where I'd want a round-trip through the serialized form to produce a graph that is exactly identical so that any indices into that graph are preserved.  If there were a single serialized format for Graph and friends, it seems like it would need to carry more information in some cases than others.

bluss:
Ok, that's reasonable. Since this is effectively smashing the encapsulation of the implementation, I'll have to think carefully about it. Also needs a doc to point out the new feature flag, and maybe consider moving the serialization tests to a subcrate. Though the fully external serde implementation strategy is interesting too (See https://github.com/bluss/rust-ndarray/issues/274 for a description of that).

generalelectrix:
I'll leave the PR open for now, but you're right: it would be significantly better to Do It Right. Especially considering that this is an important data structure library for the language, keeping the implementation well hidden is particularly important.  A proper implementation probably won't even be that hard and would spur development of the aforementioned "turn this graph into an iterator of all the elements".

The only downside I can see there is that serializing/deserializing might be more expensive due to the extra abstraction layer, but I would hazard no guess as to where the real life bottleneck would be there anyway.

I'm a bit busy this week but might have some time to put into in next week, let me know if you're poking at it so I can avoid duplicating effort.

bluss:
serde is 1.0 :100: which is a big relief..

bluss:
This was fixed in #166

PR #143, opened 2017-04-21
Add A* pathfinding algorithm

### LINKED ISSUE #138: astar algorithm
It would be nice to have  [A*](https://en.wikipedia.org/wiki/A*_search_algorithm) implementation in petgraph library. For path finding and similar stuff.

jmcomets:
## Summary

I've written the current A* to return the full path from start to finish, unlike the Dijkstra that simply computes all costs, stopping if the goal is found. Additionally, I could add in the computed cost of the path.

I've added a test that validates a basic usage, although there should be a proof that the heuristic drives the pathfinding.

Performance-wise, the tracking of previous nodes to later rebuild the path from start to finish adds on the Dijkstra's performance cost, otherwise it's roughly equivalent.

## Sidenotes

Some work could be done to allow more flexibility in terms of data-structures used by the algorithm, as it currently requires that node IDs be `Eq` + `Hash` when we could have another version only requiring `Ord` which would use trees instead of hashmaps.

## Questions

- [x] Should the path returned be the list of node IDs or the actual edge IDs?
- [x] Should the pathfinding algorithms be reworked to share some common "flexible" interface?
- [ ] Should there be bench tests on the A* vs the Dijkstra when `H(x) = 0` to measure the cost of tracking the path?

(fixes #138)

### DIFF
--- a/src/astar.rs
+++ b/src/astar.rs
@@ -22,16 +22,20 @@ use algo::Measure;
 
 /// [Generic] A* shortest path algorithm.
 ///
-/// Compute the length of the shortest path from `start` to `finish`.
+/// Computes the shortest path from `start` to `finish`, including the total path cost.
 ///
 /// `finish` is implicitly given via the `is_goal` callback, which should return `true` if the
 /// given node is the finish node.
 ///
-/// The function `edge_cost` should return the cost for a particular edge, which is used to compute
-/// path costs.
+/// The function `edge_cost` should return the cost for a particular edge. Edge costs must be
+/// non-negative.
 ///
 /// The function `estimate_cost` should return the estimated cost to the finish for a particular
-/// node, also used to compute path costs.
+/// node. For the algorithm to find the actual shortest path, it should be admissible, meaning that
+/// it should never overestimate the actual cost to get to the nearest goal node. Estimate costs
+/// must also be non-negative.
+///
+/// The graph should be `Visitable` and implement `IntoEdges`.
 ///
 /// ```
 /// use petgraph::Graph;
@@ -57,8 +61,6 @@ use algo::Measure;
 /// let path = astar(&g, a, |finish| finish == f, |e| *e.weight(), |_| 0);
 /// assert_eq!(path, Some((6, vec![a, d, e, f])));
 /// ```
-///
-/// The graph should be `Visitable` and implement `IntoEdges`. Edge costs must be non-negative.
 ///
 /// Returns the total cost + the path of subsequent `NodeId` from start to finish, if one was
 /// found.
--- a/src/astar.rs
+++ b/src/astar.rs
@@ -55,15 +55,16 @@ use algo::Measure;
 /// ]);
 ///
 /// let path = astar(&g, a, |finish| finish == f, |e| *e.weight(), |_| 0);
-/// assert_eq!(path, Some(vec![a, d, e, f]));
+/// assert_eq!(path, Some((6, vec![a, d, e, f])));
 /// ```
 ///
 /// The graph should be `Visitable` and implement `IntoEdges`. Edge costs must be non-negative.
 ///
-/// Returns a Path of subsequent `NodeId` from start to finish, if one was found.
+/// Returns the total cost + the path of subsequent `NodeId` from start to finish, if one was
+/// found.
 pub fn astar<G, F, H, K, IsGoal>(graph: G, start: G::NodeId, mut is_goal: IsGoal,
                                      mut edge_cost: F, mut estimate_cost: H)
-    -> Option<Vec<G::NodeId>>
+    -> Option<(K, Vec<G::NodeId>)>
     where G: IntoEdges + Visitable,
           IsGoal: FnMut(G::NodeId) -> bool,
           G::NodeId: Eq + Hash,
@@ -82,7 +83,9 @@ pub fn astar<G, F, H, K, IsGoal>(graph: G, start: G::NodeId, mut is_goal: IsGoal
 
     while let Some(MinScored(_, node)) = visit_next.pop() {
         if is_goal(node) {
-            return Some(path_tracker.reconstruct_path_to(node));
+            let path = path_tracker.reconstruct_path_to(node);
+            let cost = scores[&node];
+            return Some((cost, path));
         }
 
         // Don't visit the same node several times, as the first time it was visited it was using
--- a/src/astar.rs
+++ b/src/astar.rs
@@ -1,0 +1,86 @@
+use std::collections::{
+    HashMap,
+    BinaryHeap,
+};
+use std::collections::hash_map::Entry::{
+    Occupied,
+    Vacant,
+};
+
+use std::hash::Hash;
+
+use scored::MinScored;
+use super::visit::{
+    Visitable,
+    VisitMap,
+    IntoEdges,
+    EdgeRef,
+};
+
+use algo::Measure;
+
+/// [Generic] A* shortest path algorithm.
+///
+/// Compute the length of the shortest path from `start` to the goal node.
+///
+/// The graph should be `Visitable` and implement `IntoEdges`. The function `edge_cost` should
+/// return the cost for a particular edge, which is used to compute path costs.  The function
+/// `estimate_cost` should return the estimated cost to the goal for a particular node, also used
+/// to compute path costs. Edge costs must be non-negative.
+///
+/// Returns a `HashMap` that maps `NodeId` to path cost.
+pub fn astar<G, F, H, K>(graph: G, start: G::NodeId, goal: Option<G::NodeId>,
+                         mut edge_cost: F, mut estimate_cost: H)
+    -> HashMap<G::NodeId, K>
+    where G: IntoEdges + Visitable,
+          G::NodeId: Eq + Hash,
+          F: FnMut(G::EdgeRef) -> K,
+          H: FnMut(G::NodeId) -> K,
+          K: Measure + Copy,
+{
+    let mut visited = graph.visit_map();
+    let mut visit_next = BinaryHeap::new();
+    let mut scores = HashMap::new();
+
+    let zero_score = K::default();
+    scores.insert(start, zero_score);
+    visit_next.push(MinScored(estimate_cost(start), start));
+
+    while let Some(MinScored(_, node)) = visit_next.pop() {
+        if visited.is_visited(&node) {
+            continue
+        }
+
+        if goal.as_ref() == Some(&node) {
+            break
+        }
+
+        for edge in graph.edges(node) {
+            let next = edge.target();
+            if visited.is_visited(&next) {
+                continue
+            }
+
+            let node_score = *scores.get(&node).unwrap();
+            let mut next_score = node_score + edge_cost(edge);
+
+            // TODO: understand this section
+            match scores.entry(next) {
+                Occupied(ent) => if next_score < *ent.get() {
+                    *ent.into_mut() = next_score;
+                } else {
+                    next_score = *ent.get();
+                },
+                Vacant(ent) => {
+                    ent.insert(next_score);
+                }
+            }
+
+            let next_estimate_score = next_score + estimate_cost(next);
+            visit_next.push(MinScored(next_estimate_score, next));
+        }
+        visited.visit(node);
+    }
+
+    scores
+}
--- a/src/astar.rs
+++ b/src/astar.rs
@@ -11,27 +11,28 @@ use std::hash::Hash;
 
 use scored::MinScored;
 use super::visit::{
-    Visitable,
-    VisitMap,
-    IntoEdges,
     EdgeRef,
+    GraphBase,
+    IntoEdges,
+    VisitMap,
+    Visitable,
 };
 
 use algo::Measure;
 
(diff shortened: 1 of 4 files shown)


bluss:
This looks really good. I've worked down PR queues in other repos but I'm sorry I haven't had much progress in petgraph. It's going to happen, but a bit late.

bluss:
This is where I think of something like the inverse of an iterator (Think instead of Iterator::next we have a generalization of Vec::push, something that eats elements) -- so that the user could submit both a vector for path and for edges, if they wanted to save both those.

jmcomets:
Thanks for the update!

Concerning the generalization of `Vec::push`, since the path is reconstructed from finish to start, I need to reconstruct the path otherwise the user would have to handle each path component in reverse order.

Maybe `Path<G>` can be structured with a few iterators on things like: node ids, edge ids? I feel like there should be some kind of all-purpose `Path` type that would be returned by the various pathfinding functions.

src/astar.rs
```diff
@@ -0,0 +1,140 @@
...
+            return Some(path_tracker.reconstruct_path_to(finish));
+        }
+
+        // Don't visit the same node several times, as the first time it was visited it was using
+        // the shortest available path.
+        if visited.is_visited(&node) {
+            continue
+        }
+
+        visited.visit(node);
```
bluss: We can combine the is_visited check with this visit call -- using the return value from visit. It returns false if it's already visited.

bluss:
Ah the path idea, do you mean we use a bit more opaque type? 

Say if we store the node ids, we can take the path and the graph and make an iterator that gives the edge ids and so on?

jmcomets:
(replying to your comment about the `Path` type)

Precisely. Only problem with that is that although both `G::NodeId` and `G::EdgeId` are `Copy`, the opaque type would have to borrow a `G`.

bluss:
Can you elaborate on that problem?

I guess parallel edges existing means that we can't give an unambiguous path in terms of edges indices, too. Unless we save them, which sounds simple to do.

jmcomets:
NVM, I forgot about the `IntoEdges::edges()` method.

Would something like this work?

```rust
impl<G:GraphBase> Path<G> {
  fn node_ids(&self) -> impl Iterator<Item=G::NodeId> {
    unimplemented!();
  }

  fn edge_ids(&self) -> impl Iterator<Item=&[G::EdgeId]> {
    unimplemented!();
  }
}
```

I'm also thinking that yielding `&[G::EdgeId]` will be awkward to use, as people generally expect only a unique path from source to target. Not to mention that 90% of the time people don't use parallel edges.

Another option instead of yielding a `&[G::EdgeId]` is to yield an enum `NodeConnection` (apologies for the awful name), which would look something like this:

```rust
enum NodeConnection<G: GraphBase> {
  OneEdge(G::EdgeId),
  MultipleEdges(Vec<G::EdgeId>),
}

impl<G: GraphBase> NodeConnection<G> {
  fn as_one_edge(&self) -> Option<G::EdgeId> {
    if let NodeConnection::OneEdge(edge_id) = *self { Some(edge_id) } else { None }
  }

  fn as_multiple_edges(&self) -> Option<&[G::EdgeId]> {
    if let NodeConnection::MultipleEdge(ref edge_ids) = *self { Some(edge_ids) } else { None }
  }
}
```

src/astar.rs
jmcomets: *(re-posting the comment to close the damn Github review)*

Nice catch. I'll edit the comment to something like *"Don't explore the same node's children several times [...]"*, as visiting implies that there's a visitor other than the `VisitMap`.

jmcomets:
One important note: when re-reading my implementation, I noticed that the heuristic's signature is `FnMut(G::NodeId) -> K`. Shouldn't it take a `G::EdgeRef`, as it's called `edge_cost`?

In that case, we should only have to handle parallel edges *if* two edges have the same cost, and are therefore both visited during the A*.

Here's the problem with that: the `Visitable` trait only defines a `Map` that takes `G::NodeId` and not `G::EdgeId`. Therefore there's no way we'll encounter parallel edges if we use `Visitable::Map`.

bluss:
I don't think parallel edges should be a problem for the algorithm, only a problem if one wants to use nodes to identify the edge ids that were in the path.

Isn't the heuristic function standard? Like wikipedia says, it should be a function that takes a node *x* and returns an estimate of the distance from *x* to the goal.

Btw, I saw a different Rust astar algo (would link if I had the URL handy) that was using a callback to determine if a given node was the goal, instead of taking a goal node. It is more flexible that way..

bluss:
The usual rule applies: Don't overengineer it :smile:

jmcomets:
Ok I just re-checked the wikipedia article and re-read some things, you're 100% right about `H` taking a node.

Using a callback is much more flexible. I've added one in the latest version. I'm still having trouble with handling parallel edges and like you said, things are getting a little over-engineered now 😅.

Would you be ok with the current version, simply returning a list of node ids?

src/astar.rs
```diff
@@ -0,0 +1,168 @@
...
+/// ```
+///
+/// The graph should be `Visitable` and implement `IntoEdges`. Edge costs must be non-negative.
+///
+/// Returns a Path of subsequent `NodeId` from start to finish, if one was found.
+pub fn astar<G, F, H, K, IsGoal>(graph: G, start: G::NodeId, is_goal: IsGoal,
+                                     mut edge_cost: F, mut estimate_cost: H)
+    -> Option<Path<G>>
+    where G: IntoEdges + Visitable,
+          IsGoal: Fn(G::NodeId) -> bool,
```
bluss: any reason is goal is Fn instead of FnMut?

src/astar.rs
```diff
@@ -0,0 +1,168 @@
...
+/// let d = g.add_node((0., 2.));
+/// let e = g.add_node((3., 3.));
+/// let f = g.add_node((4., 2.));
+/// let ab = g.add_edge(a, b, 2);
+/// let ad = g.add_edge(a, d, 4);
+/// let bc = g.add_edge(b, c, 1);
+/// let bf = g.add_edge(b, f, 7);
+/// let ce = g.add_edge(c, e, 5);
+/// let ef = g.add_edge(e, f, 1);
+/// let de = g.add_edge(d, e, 1);
```
bluss: Using `.extend_with_edges(&[(a, b, 2),` ... `])` to add all the edges would maybe improve this example

src/astar.rs
```diff
@@ -0,0 +1,168 @@
...
+
+            let next_estimate_score = next_score + estimate_cost(next);
+            visit_next.push(MinScored(next_estimate_score, next));
+        }
+    }
+
+    None
+}
+
+pub type Path<G: GraphBase> = Vec<G::NodeId>;
```
bluss: probably rewrite this without the trait bound and with `<G as GraphBase>::NodeId` instead. That's the form the compiler doesn't warn about right now at least..

bluss: Possibly this should just be a private type alias. Then it's just a Vec in the public docs.

bluss:
This looks good. I want to include this in the 0.4.x release series. I might put out a point release before this merges, though.

bluss:
You might have seen my other recent fixups, seeing that I prioritize non-breaking changes. New algo is good..

src/astar.rs
jmcomets: Nice catch, I'll switch that.

src/astar.rs
jmcomets: I removed the `Path` alias for now, since it's just a `Vec`.

src/astar.rs
```diff
@@ -0,0 +1,168 @@
...
+        // the shortest available path.
+        if !visited.visit(node) {
+            continue
+        }
+
+        // This lookup can be unwrapped without fear of panic since the node was necessarily scored
+        // before adding him to `visit_next`.
+        let node_score = *scores.get(&node).unwrap();
+
+        for edge in graph.edges(node) {
```
bluss: I believe this handles undirected and directed graphs both correctly, by virtue of how `.edges(..)` works. Everything should just work.

src/astar.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        // Don't visit the same node several times, as the first time it was visited it was using
+        // the shortest available path.
+        if !visited.visit(node) {
+            continue
+        }
+
+        // This lookup can be unwrapped without fear of panic since the node was necessarily scored
+        // before adding him to `visit_next`.
+        let node_score = *scores.get(&node).unwrap();
```
bluss: Ah by the way, *foo.get(&k).unwrap() can just be `foo[&k]`.

bluss:
I think this should return the path cost as well. Either the whole score map, or just the cost for the found path.

Sorry for not looking at this with my full attention. Low complexity is good, but we need to align it a bit with dijkstra.

Also for inspiration we can alsways look in boost::graph: http://www.boost.org/doc/libs/1_46_0/libs/graph/doc/astar_search.html  Looking there it's very clear that petgraph is not boost::graph, in terms of type parameters.

jmcomets:
> Sorry for not looking at this with my full attention.

No problem, I was actually glad this was getting some attention. 😄 

I think returning the predecessor map is a bit heavy API-wise, the user is likely uninterested in the algorithm's intermediate steps and more interested in the final result (eg. the path cost). Do you think a return type of `(Vec<G::NodeId>, K)` would do the trick?

bluss:
Yes. Just an impulse, cost should probably be field 0 in the tuple. I imagine we can add a more configurable astar later.

src/astar.rs
```diff
@@ -0,0 +1,171 @@
...
+/// Compute the length of the shortest path from `start` to `finish`.
+///
+/// `finish` is implicitly given via the `is_goal` callback, which should return `true` if the
+/// given node is the finish node.
+///
+/// The function `edge_cost` should return the cost for a particular edge, which is used to compute
+/// path costs.
+///
+/// The function `estimate_cost` should return the estimated cost to the finish for a particular
+/// node, also used to compute path costs.
```
bluss: This sentence needs clarification. I think I can do that, but I wonder what you think too. I think the "also used to compute path costs" is confusing and should be removed. Instead I'd expect it to clarify us what the requirements of the heuristic are, and as wikipedia says:

> For the algorithm to find the actual shortest path, the heuristic function must be admissible, meaning that it never overestimates the actual cost to get to the nearest goal node

jmcomets: I agree with you 100% on this. I haven't checked the docstring for quite some time, I need to give it another read to see what needs to be rewritten.

tests/graph.rs
```diff
@@ -342,6 +343,62 @@ fn dijk() {
...
+    let d = g.add_node((0., 2.));
+    let e = g.add_node((3., 3.));
+    let f = g.add_node((4., 2.));
+    g.add_edge(a, b, 2.);
+    g.add_edge(a, d, 4.);
+    g.add_edge(b, c, 1.);
+    g.add_edge(b, f, 7.);
+    g.add_edge(c, e, 5.);
+    g.add_edge(e, f, 1.);
+    g.add_edge(d, e, 1.);
```
bluss: Edge weights don't seem to be the manhattan distance here?

tests/graph.rs
```diff
@@ -342,6 +343,62 @@ fn dijk() {
...
+
+    let path = astar(&g, a, |finish| finish == f, |e| *e.weight(),
+    |node| {
+        let (x1, y1): (f32, f32) = g[node];
+        let (x2, y2): (f32, f32) = g[f];
+
+        let (x, y) = (x2 + x1, y2 + y1);
+
+        x.powi(2) + y.powi(2)
```
bluss: Euclidean distance is sqrt((x2 - x1)² + (y2 - y1)²) and is that what this is going for?

tests/graph.rs
```diff
@@ -342,6 +343,62 @@ fn dijk() {
...
+    g.add_edge(b, f, 15);
+    g.add_edge(c, f, 11);
+    g.add_edge(e, f, 6);
+
+    let path = astar(&g, a, |finish| finish == e, |e| *e.weight(), |_| 0);
+    assert_eq!(path, Some((23, vec![a, d, e])));
+
+    let path = astar(&g, e, |finish| finish == b, |e| *e.weight(), |_| 0);
+    assert_eq!(path, None);
+}
```
bluss: Looks like a good enough test. The exhaustive approach is to use dijkstra and check that every path and path cost for *a* to *any node* agrees in both algorithms with the trivial heuristic.

We've had success with the exhaustive approach, checking algorithms against each other in petgraph before. It's a good way to gain confidence in the implementation. (See bugs in this crate's history)

bluss: That kind of setup https://github.com/bluss/petgraph/pull/61

src/astar.rs
bluss: I don't want to give you more to do though!! :-) I have a comment on tests so I can't stop idly trickling in comments.

bluss:
Thanks!

gorn:
This is alrady implemented in https://docs.rs/petgraph/0.4.12/petgraph/algo/fn.astar.html, could anyone sumup what is the relation to petgraph crate?

PR #144, opened 2017-04-22
Add generic implementation of FromElements

jmcomets:
Just a small code duplication I stumbled upon when hacking around the codebase.
Basically, any graph `G: Create + NodeIndexable` can automatically have an
implementation of `FromElements` that uses `from_elements_indexable`.

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -275,7 +275,6 @@ pub trait FromElements : Create {
         }
         gr
     }
-        
 }
 
 fn from_elements_indexable<G, I>(iterable: I) -> G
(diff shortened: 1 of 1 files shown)


bluss:
Thanks.

The existing code is deliberate (but it's of course not perfect).

The difference is the API. Should `impl<G: Create + NodeIndexable> FromElements for G` be something that petgraph puts out? The original code thought no, we expose only that some certain types implement FromElements, there's no blanket impl.

I understand the impulse here. I don't want to be too boring, but are the NodeIndexable + Create traits complete in how they are documented to work, so that this works for all such graphs?

jmcomets:
I completely understand. I've been trying to grok the various traits that petgraph has and this bit triggered a bad habit of mine: refactoring without giving enough thought.

> Should `impl<G: Create + NodeIndexable> FromElements for G` be something that petgraph puts out?
Now that I've thought of it, it seems like the `NodeIndexable` is too much of an implementation detail to put out as the public API. It adds a level of indirection that makes it harder to discover and would need an example to make things more clear.

> are the NodeIndexable + Create traits complete in how they are documented to work, so that this works for all such graphs?

Here's my understanding: the `NodeIndexable` trait provides a bidirectional mapping between nodes and indices, the `Create` trait a hinted constructor and the inherited `Build` trait graph construction.

The only unclear behavior in this blanket impl is the handling of parallel edges. If I'm not mistaken, they're currently ignored when the user may want to quit the construction of the graph, pick which edge should stay or even merge both edge weights somehow.

bluss:
I'm sorry that I'm not so good at answering questions. I'd answer if I knew. I think NodeIndexable should be enough, just need to think about how graphs with holes behave (StableGraph) and if there are other kinds of graphs with holes. We already assume that the indices are close to zero and relatively compact. (An indexable graph would not use indices *n* to *n + 100* for some big *n*).

PR #146, opened 2017-04-30
Minor fixes for doc comments

cmbrandenburg:


bluss:
Thank you, if ever so late.

PR #147, opened 2017-05-11
Add support for pruning nodes from a DFS.

khuey:


khuey:
@bluss is there any chance of getting this PR merged?  It's pretty straightforward and would let me stop maintaining a fork :)

bluss:
Hi, sorry for letting this slip. It's not a big deal but it's also a breaking change so it's not straightforward: it is hindering if it is merged without release planning. It can wait until we have enough to warrant a breaking upgrade or we try to make it nonbreaking.

bluss:
This is a worthwhile feature, but I'll close the PR for inactivity. Now, soon, should be a good time to merge this. I think a few more changes to the support traits and types are needed -- this change means that Control is now explicitly a Dfs visit control enum. I would revisit the Result impl of ControlFlow too (Maybe it can be `impl<C, E> ControlFlow for Result<C, E> where C: ControlFlow` ?)

PR #149, opened 2017-06-02
Deconstruct a graph into Elements and rebuild it as another graph.

Rupsbant:
This request implements the inverse of FromElements. It deconstructs a graph (Graph or StableGraph) into an Iterator of Elements of Cloned weights, ready to be rebuild. 

GraphMap is not compatible because it doesn't implement the necessary interfaces, but it has the `into_graph()`method.

The signature of the trait interface `FromElements::from_elements` was relaxed from a moved `IntoIterator` to a reference to an `Iterator`.


src/into_elements.rs
```diff
@@ -0,0 +1,73 @@
...
+        Box::new(all)
+    }
+}
+
+/// Rebuild a graph with Cloneable weights into another graph.
+pub fn rebuild<'a, G, H>(g: &'a G) -> H
+    where G: 'a + DataMap + Data,
+          &'a G: IntoNodeIdentifiers + IntoEdgeReferences,
+          &'a G: GraphBase<NodeId = G::NodeId, EdgeId = G::EdgeId>,
+          &'a G: Data<EdgeWeight = G::EdgeWeight>,
```
bluss: Here petgraph would normally use a type parameter `G`, for the reference to Graph or in general a `GraphRef` (trait)

bluss:
Hi, thank you.

I disagree with this, it is not a relaxation, it is less generic this way. Mut ref to iterator is definitely already an `IntoIterator`.

> The signature of the trait interface FromElements::from_elements was relaxed from a moved IntoIterator to a reference to an Iterator.

Rupsbant:
You're right, I'll change it. I didn't understand enough of rust yet.

src/into_elements.rs
```diff
@@ -0,0 +1,65 @@
...
+pub struct IntoElements<'a, G: 'a>(pub &'a G);
+
+
+//Todo: ids do not need to be cloneable. I couldn't get the lifetimes of EdgeWeight to work.
+/// Graphs that have a Cloneable weigths and implement the necessary traits can be deconstructed into elements.
+impl<'a, G> IntoIterator for IntoElements<'a, G>
+    where &'a G: GraphBase + DataMap + IntoNodeIdentifiers + IntoEdgeReferences,
+          <&'a G as Data>::NodeWeight: Clone,
+          <&'a G as Data>::EdgeWeight: Clone,
+          <&'a G as GraphBase>::NodeId: Eq + Hash + Clone,
```
bluss: Please convert this into using `G` as a GraphRef instead of hardcoding it as `&G`. Same in the struct definition.

src/into_elements.rs
```diff
@@ -0,0 +1,65 @@
...
+{
+    type Item = Element<<&'a G as Data>::NodeWeight, <&'a G as Data>::EdgeWeight>;
+    type IntoIter = Box<Iterator<Item = Self::Item> + 'a>;
+    fn into_iter(self) -> Self::IntoIter {
+        use std::iter;
+        let nodes = self.0.node_identifiers();
+        let edge_refs = self.0.edge_references();
+        let map = Rc::new(RefCell::new(HashMap::<<&'a G as GraphBase>::NodeId, usize>::new()));
```
bluss: Why is there an Rc'd map here? Now I'm surprised.

bluss: We'll find a way to resolve this without Rc's and `Box<Iterators>` before we can merge it.

bluss: Feel free to rebase and rewrite commits while working on this, if it makes the commit history cleaner.

PR #151, opened 2017-06-05
Add transitive_closure to algo

BurntPizza:
Might want to add more description, but I feel this is useful.


bluss:
Thanks for the interesting suggestion and sorry for the delay.

I would generally encourage more motivation around the PR, tell why is it useful, why this design and so on. The return type of a bitmap surprises me and will others as well I think. It's not ideal that we don't have a type for an adjacency matrix, but that's clearly what's missing here.

Dylan-DPC:
@BurntPizza can you update the description?

XVilka:
@waywardmonkeys this might be interesting to include into the upcoming release too, it is self-contained and transitive closure is kind of famous task for graph theory:
- https://algowiki-project.org/en/Transitive_closure_of_a_directed_graph
- https://en.wikipedia.org/wiki/Transitive_closure

waywardmonkeys:
Maybe this should take the changes in PR #173 into account? (I know ... this was opened before PR #173, and this is all done long long ago...)

jmcomets:
I did some hacking around [on this branch](https://github.com/jmcomets/petgraph/tree/transitive-closure), both to compare the algorithm's performance with one based on [Floyd-Warshall](https://en.wikipedia.org/wiki/Floyd%E2%80%93Warshall_algorithm) & to find an appropriate interface to return to the user.

Here's the basic interface I think would be easier to use:
```rust
impl<G: NodeIndexable> TransitiveClosure<G> {                                                                       
    pub fn has_relation(&self, a: G::NodeId, b: G::NodeId) -> bool { /* ... */ }
}
```
When comparing the current implementation with Floyd-Warshall I get similar performance, slightly in favor of this PR's version:
```rust
#[bench] // PR version
fn transitive_closure_dfs(bench: &mut Bencher) {                                                                                                   
    let g = digraph().bigger();                                                                                                                    
    bench.iter(|| transitive_closure(&g));                                                                                                         
}                                                                                                                                             
                                                                                                                                                   
#[bench] // Floyd-Warshall
fn transitive_closure_dp_using_fixed_bitset(bench: &mut Bencher) {                                                                                 
    let g = digraph().bigger();                                                                                                                    
    bench.iter(|| TransitiveClosure::using_fixed_bitset(&g));                                                                                      
}
                                                                                                                                                  
#[bench] // Floyd-Warshall, swapping out the `FixedBitSet` with a `MatrixGraph<(), ()>`
fn transitive_closure_dp_using_matrix_graph(bench: &mut Bencher) {                                                                                 
    let g = digraph().bigger();                                                                                                                    
    bench.iter(|| TransitiveClosure::using_matrix_graph(&g));
}
```
Here are the results:
```
test transitive_closure_dfs                   ... bench:      20,319 ns/iter (+/- 577)
test transitive_closure_dp_using_fixed_bitset ... bench:      25,912 ns/iter (+/- 262)
test transitive_closure_dp_using_matrix_graph ... bench:      23,823 ns/iter (+/- 380)
```
There's some unneeded branching to take into account so it's not 100% accurate (in favor of this PR's version), but overall there's no surprises. I'd be in favor of merging this PR if it provides a cleaner return value.

XVilka:
So, since there is no consensus on this, maybe it's better to postpone this one after the 0.5 version, and release 0.5 as soon as possible? There are many important fixes were merged in, in my opinion waiting more is not the best strategy.

symphorien:
#263 also provides an implementation of transitive closure.

koivunej:
Can't see why this PR was never picked up but I just rebased it on top of the latest master. The two conflicts are not anything difficult to handle. The rebased commit can be found at: https://github.com/koivunej/petgraph/commit/50f57676c2e1d116636806666c61b9e9b035a206

XVilka:
I am closing this since https://github.com/petgraph/petgraph/pull/263 was successfully merged, thanks to @symphorien !

PR #152, opened 2017-06-12
Modify bellman_ford to work with undirected graphs, and add quickcheck test

### LINKED ISSUE #150: bellman_ford does not handle Undirected graphs correctly
`bellman_ford` fails when called on certain nodes in an undirected graph:
```
extern crate petgraph;
use petgraph::prelude::*;
use petgraph::algo;
fn main() {
    let mut graph: GraphMap<usize, f32, Undirected> = GraphMap::new();
    graph.add_edge(0, 1, 1f32);
    algo::bellman_ford(&graph, 1).unwrap();
}
```

```
> cargo run
   Compiling pgtest v0.1.0 (file:///home/***************)
    Finished dev [unoptimized + debuginfo] target(s) in 0.78 secs
     Running `target/debug/pgtest`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NegativeCycle(())', /checkout/src/libcore/result.rs:859
```

This is because `bellman_ford` uses `Graph::edge_references` in the main pass of the algorithm, and so treats undirected edges as though they had a single direction. This leaves nodes that should be reachable marked as unreachable. In the test pass of the algorithm, the function uses `Graph::edges`, which correctly checks the edges in both directions, but incorrectly assumes it has found a negative cycle when it then visits a node previously thought to be unreachable.

I believe one of two things needs to change to fix this:
 1.  `edge_references` should return edges in both forward and backward directions (when called on an undirected graph), or
 2. `bellman_ford` needs some way of checking whether `G` is `Directed` or `Undirected`.

I'm not particularly familiar with this project, so I don't know how feasible the potential solutions are.

carrutstick:
Fixes #150:
 * Adds a check in bellman_ford that applies edges in both directions if the graph is undirected.
 * Adds a quickcheck test to verify that bellman_ford returns results for undirected graphs with non-negative edge weights.

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -33,7 +33,7 @@ use super::unionfind::UnionFind;
 use super::graph::{
     IndexType,
 };
-use visit::{Data, NodeRef, IntoNodeReferences, GraphProp};
+use visit::{Data, NodeRef, IntoNodeReferences};
 use data::{
     Element,
 };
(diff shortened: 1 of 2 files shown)


bluss:
Looks good. Just expect some delay since it needs to go into petgraph 0.5

bluss:
I have been looking at ways to include this without a breaking change, but I can't come up with anything short of hacks. Hacks like for example adding methods like "edges_directed() -> bool" to the IntoEdges trait with a default impl.  Any other ideas?

bluss:
I think it could have a significant effect yes. The global edge scan should be much faster than edge per node iteration with the graph types in petgraph (except Csr). This is not meant as a final decision in either way.

It probably would be good with a trait - data structures - algorithms crate split, so that they can version separately. One person is too few to keep it all organized.

bluss:
I think we should fix the bug, and fix the performance issues in a breaking change later.

bluss:
Thanks for the solid work. I'd like to squash merge this PR, then history is really neat, and it's small enough to just be one commit.

carrutstick:
No problem, happy to help!

Issue #139, opened 2017-04-16
Support Serialization using serde


generalelectrix:
The backing collections behind Graph and friends are trivially compatible with serde if the node and edge weights implement Serialize and Deserialize.  As a client of this library, however, there's no way to do this, and instead one must resort to a complex/hacky scheme to do this.  Is there a simple way to provide this functionality without forcing all clients of this crate to take serde as a dependency?

Otherwise it seems that I'm left with forking this project and adding it myself, or implementing a hacky way to restore a serialized graph.

Dushistov:
@generalelectrix You can just use http://stackoverflow.com/questions/42551113/is-it-possible-to-conditionally-derive-with-features and add suitable feature to `petgraph`, so it works for you and not touch users that not use `serde`.

bluss:
Support for serde in some way would be great. Using their externalization feature sounds good, especially before they go 1.0.

I'm reminded of this feature (from elements); we have a basic way to build a graph from a basic representation of nodes and edges here https://docs.rs/petgraph/0.4.4/petgraph/data/trait.FromElements.html  but I don't think we have the opposite yet in a nice function — from graph into an elements iterator.

generalelectrix:
Cool, it looks like it should be clean and easy to add serde support via a feature.  I'll try to get this working in my fork with forthcoming PR.

fintelia:
Has there been any progress on this issue?

generalelectrix:
I haven't dedicated any time to it as I've been focusing on other areas of the project that required it.  That project may also move away from using petgraph as it turns out a custom data structure may prove more suitable.

bluss:
Let's get going. We need a good solution.

Proposed Design Criteria

- Graph and StableGraph have the same serialized representation. The stable distinction only matters while we are removing nodes and edges.
- No funky nodeindex etc wrapper types. Indices are integers

Ideas and Alternatives

- Support both "adjacency list" or "adjacency matrix" based representation
  - Counterargument: Graph/StableGraph are not good for dense graphs anyway
- Support eliding node/edge weights if weight types are `()`.

bluss:
- Preserve graph indices across serialization
  - This one is important. Means that stable graph and graph are different.
- Preserve directionality across serialization (no weak typing / silent conversion)

bluss:
I'm working on a proposal. Examples below.

Graph (Debug output)

```rust
```

Graph (JSON serialization)
```json
```

bluss:
[...]

PR #156, opened 2017-06-29
Derive Serialize and Deserialize for some common types

fintelia:
A first pass at addressing https://github.com/bluss/petgraph/issues/139.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ debug = true
 fixedbitset = { version = "0.1.4" }
 quickcheck = { optional = true, version = "0.4", default-features = false }
 ordermap = { version = "0.2.2", optional = true }
+serde = { version = "1.0", optional = true, features = ["derive"] }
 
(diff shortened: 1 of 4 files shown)


bluss:
Hi, thank you for this.

Deriving serialization traits is convenient but it also exposes all private fields, so we lose encapsulation. Do you see where I'm coming from? Say if we want to make changes to private fields of StableGraph, then it's a breaking change. This is the reason I think the graphs need deliberate ser/de implementations.

bluss:
See also #142.

fintelia:
That makes a lot of sense

bluss:
I even think Graph and StableGraph should have the same serialized repr. (Of course something discovered during implementation could derail this goal.)

bluss:
Please see issue #139 for discussing a design

bluss:
serde support for Graph, StableGraph has been merged in #166

PR #159, opened 2017-08-27
Add minimal node support for Csr graphs

jmcomets:
I noticed when trying to use the Csr implementation that there wasn't any
support for node weights. I gave it a go and added the following:

- `add_node`
- `Index` and `IndexMut` impls

Unrelated, but I noticed that there was no `new` constructor or `Default` impl,
I added them as they're straightforward. I needed them for a test, but of
course, I can always remove them and use `with_nodes(0)`.

I noticed that there's a lot missing in this impl, notable `clear`,
`remove_node` and `remove_edge`, which are all three basic graph operations.
I'll have a try at implementing them in another PR.

### DIFF
--- a/src/csr.rs
+++ b/src/csr.rs
@@ -1,9 +1,8 @@
-
 //! Compressed Sparse Row (CSR) is a sparse adjacency matrix graph.
 
 use std::marker::PhantomData;
 use std::cmp::max;
-use std::ops::Range;
+use std::ops::{Range, Index, IndexMut};
 use std::iter::{Enumerate, Zip};
 use std::slice::Windows;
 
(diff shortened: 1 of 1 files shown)


bluss:
Thanks a lot. It makes sense. We'll document how to best use Csr -- it is still going to revolve mostly around creating it in one go. I'm thinking of adding an "extend" version of the from sorted edges function.

bluss:
Damn it, I'm not sure why it's stuck on squash merge. I'll remember to check that button again on the next PR.

bluss:
Ah I'll also change with_nodes so that it's not a breaking change

PR #166, opened 2017-09-10
New optional feature serde-1: serde serialization for Graph, StableGraph

bluss:
- Implement serde serialization for Graph, StableGraph

  - Both graphs use the same representation; we can always convert Graph
    -> StableGraph but only the other way if it has no holes.

  - The graphs use a manual serialization method so that we encode
    graphs portably and without redundant information. Supports both
    bincode style and json style (self describing) serialization.

  - We use serde deriving as far as possible, but via intermediate data
    structures that we create during de/serialization. This is done to
    reach all the goals of portable serialization and additionally,
    avoiding excess copying or reallocation. Serialization is zero
    allocation.

  - An important property of the serialization is that node and edge
   indices remain valid across de/serialization. Integer size issues
   are checked in deserialization. Binary format serializers will
   need to decode using the same integer size as was used for encode.
   (As an example scenario: we encode two graphs that use identical
   node indices and a single extra node index for a "starting position";
   all remain compatible after deserialization.)

- Other graphs can learn serialization down the line. Getting Graph
  & StableGraph finished is a major milestone.

- Depend on itertools 0.6.2 for some utilities we had copied from them before


bluss:
Adresses #139 in important parts (Graph, StableGraph)

bluss:
cc @dtolnay if you want to look it through and sanity check, or make note of better ways to implement this.

I have sort of split de/serialization into separate structs, this is to meet the goal of no cloning for serialization

The utilities in [src/serde_utils.rs](https://github.com/bluss/petgraph/blob/cefbedd84b6638cede080455675a7a58ad46980f/src/serde_utils.rs), especially MappedSequenceVisitorhave been helpful to lazily munge elements while they are deserialized and before they end up in a Vec.

bluss:
Graph serialization implementation in file [src/graph_impl/serialization.rs](https://github.com/bluss/petgraph/blob/e32d0f65fca374154706178fc188e4b9d60af8ff/src/graph_impl/serialization.rs)

dtolnay:
I can't right now, but will try to look through this by tomorrow night.

src/serde_utils.rs
```diff
@@ -0,0 +1,92 @@
...
+
+
+
+/// Serde combinator. A sequence visitor that maps deserialized elements
+/// lazily; the visitor can also emit new errors if the elements have errors.
+pub struct MappedSequenceVisitor<T, R, F>
+    where F: Fn(T) -> Result<R, &'static str>
+{
+    f: F,
+    marker: PhantomData<fn() -> T>
```
dtolnay: Is this correct? There are no methods that return T.

src/serde_utils.rs
```diff
@@ -0,0 +1,92 @@
...
+}
```
dtolnay: How come this doesn't fail with unused type parameter R?

bluss: It is used in a projection of the assocoated type of the closure trait. Same rule w asc. types just a bit more hidden when it comes to closures.

bluss: It is used in a projection of the assocoated type of the closure trait. Same rule w asc. types just a bit more hidden when it comes to closures.

bluss:
Thanks for looking through it, much appreciated! Derive is very powerful, it's a nice design.

src/serde_utils.rs
bluss: It is used in the Fn to T sense, I don't think it matters that there is no method with the same signature.

It is also used in the Fn Of T to R sense, so the result must be that it is invariant in T. It is not mandatory to specify this exactly for any reason, so I'm content with the approximation here. It does manage to avoid the implications of phantom data of T, at least.

fintelia:
It would be preferable if the feature were called "serde" instead of "serde-1". You can do this by changing the dependencies to `serde` and `serde_derive` to the single line:
```toml
serde = { version = "1.0", optional = true, features = ["derive"] }
```
(The "derive" feature was added to serde to support this use case)

bluss:
The "serde-1" name was my conclusion from previous serde experiences (1) separate dependency names and crate feature names and (2) versioning a crate feature where pressure will exist to support new versions.

Maybe the pre 1.0 serde era was especially painful, but if serde 2.0 is released it will be back the same way.

I don't know the best way -- the best way is probably something that's using a cargo feature that doesn't exist yet.

fintelia:
The [Rust API Guidelines](https://rust-lang-nursery.github.io/api-guidelines/interoperability.html#crate-has-a-serde-cfg-option-that-enables-serde-c-serde-cfg) suggest just calling the feature serde. However, I'm not quite sure what the plan is if there's ever a serde 2.

bluss:
Yes the API guidelines are good, but I'm not sure they have thought this through. I've been burnt by serde versioning and need to choose to either learn from that and adapt my code, or to trust it doesn't happen again.

PR #168, opened 2017-09-16
Fix StableGraph::add_edge to error when using vacant nodes

bluss:
.add_edge() would sometimes accept adding edges between interior vacant nodes;
fix it by not using any Graph code.


bluss:
Apparently this fix didn't get into the previous version for some reason.

PR #172, opened 2017-10-07
Use OrderMap 0.3

bluss:
- No impact for public api
- Only used by graphmap (feature graphmap)


bluss:
sigh, squash and merge again. Maybe watch that button better, self.

PR #173, opened 2017-10-07
Add adjacency matrix implementation

jmcomets:
*Equivalent to #120, implements #28.*

This pull-request adds a `MatrixGraph` structure which implements all graph traits
using a dense adjacency matrix for its underlying structure.

The matrix itself is represented as a flat array where each entry is an edge. The size of the array varies, depending on the graph being directed (`~V^2`) or undirected (`V*(V+1)/2`). See below for more details.

**Directed Graph:**

Layout is a flattened 2D array, where `(i, j)` maps to an array index `i * capacity + j`.

```
 0  1  2  3
 4  5  6  7
 8  9 10 11
12 13 14 15
```

The node capacity is greater than the number of nodes, and is maintained to a power of two, reducing the number of resizes that need to happen as they involve `O(V)` buffer copies.

**Undirected Graph:**

Layout is a lower triangular matrix, where the array indices are mapped like this:

```
0
1 2
3 4 5
6 7 8 9
```

Resizing this type of matrix is at worst a `Vec` resize.

**Traits:**

- [x] `Default`
- [x] `Clone`
- [x] `GraphBase`
- [x] `GraphProp`
- [x] `IntoNeighborsDirected`
- [x] `IntoNodeIdentifiers`
- [x] `IntoNodeReferences`
- [x] `IntoNeighbors`
- [x] `IntoEdges`
- [x] `IntoEdgeReferences`
- [x] `Visitable`
- [x] `GetAdjacencyMatrix`
- [x] `NodeCount`
- [x] `NodeIndexable`
- [x] `NodeCompactIndexable`
- [x] `Data`
- [x] `Build`

**Questions:**

- [x] Any suggestions for better naming are welcome. ~Maybe something like
`AdjacencyMatrixGraph` would work? (Although it is much more verbose.)~ Settled on `MatrixGraph`.

**Misc:**

- Edge weights are directly stored in the matrix, which whould probably appear clearly in the doc because the underlying array could get quite large if the user isn't careful.
- Parallel edges are not supported but aren't too difficult to support, although the iterators are a bit tricky to get right.
- To support the `Build` trait, the `add_node` impl method should return an `Option<NodeIndex>`, returning `None` if the edge already exists, as parallel edges aren't currently supported. This is entirely feasible and I'll get to this when I have some time on my hands, but that may not be before closing this PR.

### DIFF
--- a/src/matrix_graph.rs
+++ b/src/matrix_graph.rs
@@ -36,6 +36,8 @@ use crate::visit::{
     Visitable,
 };
 
+use crate::data::Build;
+
 pub use crate::graph::IndexType;
 
 // The following types are used to control the max size of the adjacency matrix. Since the maximum
@@ -902,7 +904,7 @@ impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> GraphProp for MatrixGraph<N,
     type EdgeType = Ty;
 }
 
-impl<'a, N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> Data for &'a MatrixGraph<N, E, Ty, Null> {
+impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> Data for MatrixGraph<N, E, Ty, Null> {
     type NodeWeight = N;
     type EdgeWeight = E;
 }
@@ -956,7 +958,7 @@ impl<'a, N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> IntoEdges for &'a Matrix
 
 impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> NodeIndexable for MatrixGraph<N, E, Ty, Null> {
     fn node_bound(&self) -> usize {
-        MatrixGraph::node_count(self)
+        self.node_count()
     }
     fn to_index(&self, ix: NodeIndex) -> usize {
         ix.index()
@@ -967,6 +969,26 @@ impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> NodeIndexable for MatrixGrap
 }
 
 impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> NodeCompactIndexable for MatrixGraph<N, E, Ty, Null> {
+}
+
+impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> Build for MatrixGraph<N, E, Ty, Null> {
+    fn add_node(&mut self, weight: Self::NodeWeight) -> Self::NodeId {
+        self.add_node(weight)
+    }
+
+    fn add_edge(&mut self, a: Self::NodeId, b: Self::NodeId, weight: Self::EdgeWeight) -> Option<Self::EdgeId> {
+        if !self.has_edge(a, b) {
+            MatrixGraph::update_edge(self, a, b, weight);
+            Some((a, b))
+        } else {
+            None
+        }
+    }
+
+    fn update_edge(&mut self, a: Self::NodeId, b: Self::NodeId, weight: Self::EdgeWeight) -> Self::EdgeId {
+        MatrixGraph::update_edge(self, a, b, weight);
+        (a, b)
+    }
 }
 
 #[cfg(test)]
--- a/src/matrix_graph.rs
+++ b/src/matrix_graph.rs
@@ -45,6 +45,137 @@ type Ix = u16;
 
 pub type NodeIndex = GraphNodeIndex<Ix>;
 
+mod private {
+    pub trait Sealed {}
+
+    impl<N> Sealed for super::NotZero<N> {}
+    impl<N> Sealed for Option<N> {}
+}
+
+/// Wrapper trait for an `Option`, allowing user-defined structs to be input as containers when
+/// defining a null element.
+///
+/// Note: this trait currently is sealed and cannot be implemented for types outside this crate.
+pub trait Nullable: Default + Into<Option<<Self as Nullable>::Wrapped>> + private::Sealed {
+    #[doc(hidden)]
+    type Wrapped;
+
+    #[doc(hidden)]
+    fn new(value: Self::Wrapped) -> Self;
+
+    #[doc(hidden)]
+    fn as_ref(&self) -> Option<&Self::Wrapped>;
+
+    #[doc(hidden)]
+    fn as_mut(&mut self) -> Option<&mut Self::Wrapped>;
+
+    #[doc(hidden)]
+    fn is_null(&self) -> bool {
+        self.as_ref().is_none()
+    }
+}
+
+impl<T> Nullable for Option<T> {
+    type Wrapped = T;
+
+    fn new(value: T) -> Self {
+        Some(value)
+    }
+
+    fn as_ref(&self) -> Option<&Self::Wrapped> {
+        self.as_ref()
+    }
+
+    fn as_mut(&mut self) -> Option<&mut Self::Wrapped> {
+        self.as_mut()
+    }
+}
+
+/// Struct used to optimize the memory usage of edge weights `E` in a
+/// [`MatrixGraph`](struct.MatrixGraph.html), by using a sentinel value (such as 0 for numeric
+/// types) instead of the default `Option<E>`.
+///
+/// Note that if you're already using the standard non-zero types (such as `NonZeroU32`), you don't
+/// have to use this wrapper and can leave the default Null type argument.
+pub struct NotZero<N>(N);
+
+impl<N: Zero> Default for NotZero<N> {
+    fn default() -> Self {
+        NotZero(N::zero())
+    }
+}
+
+impl<N: Zero> Nullable for NotZero<N> {
+    type Wrapped = N;
+
+    fn new(value: T) -> Self {
+        assert!(!value.is_zero());
+        NotZero(value)
+    }
+
+    // implemented here for optimization purposes
+    fn is_null(&self) -> bool {
+        self.0.is_zero()
+    }
+
+    fn as_ref(&self) -> Option<&Self::Wrapped> {
+        if !self.is_null() { Some(&self.0) }
+        else { None }
+    }
+
+    fn as_mut(&mut self) -> Option<&mut Self::Wrapped> {
+        if !self.is_null() { Some(&mut self.0) }
+        else { None }
+    }
+}
+
+impl<N: Zero> Into<Option<N>> for NotZero<N> {
+    fn into(self) -> Option<N> {
+        if !self.is_null() { Some(self.0) }
+        else { None }
+    }
+}
+
+/// Base trait for types that can be wrapped in a [`NotZero`](trait.NotZero.html).
+///
+/// Implementors must provide a singleton object that will be used to mark empty edges in a
+/// [`MatrixGraph`](struct.MatrixGraph.html).
+///
+/// Note that this trait is already implemented for the base numeric types.
+pub trait Zero {
+    /// Return the singleton object which can be used as a sentinel value.
+    fn zero() -> Self;
+
+    /// Return true if `self` is equal to the sentinel value.
+    fn is_zero(&self) -> bool;
+}
+
+macro_rules! not_zero_impl {
+    ($t:ty,$z:expr) => {
+        impl Zero for $t {
+            fn zero() -> Self {
+                $z as $t
+            }
+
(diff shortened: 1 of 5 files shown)


jmcomets:
FYI I'm working on this a bit in [this branch](https://github.com/jmcomets/petgraph/tree/adjacency-matrix-v2). Don't merge this just yet.

I'm trying to use a lower triangular matrix for undirected graphs instead of having a heap for edges and duplicating the address in both `(a, b)` and `(b, a)` entries in the matrix (if I'm making any sense).

jmcomets:
Done & ready for review/test. I've got quite a bit of unit-tests in place, although there are still a bunch of edge cases that could use checking.

bluss:
Looks cool. I'm sorry if it will take a long time for me to review this fully.

Since I haven't reviewed it very well yet, these thoughts are rather loose and I may have missed something:

- AdjacencyMatrixGraph or MatrixGraph work for me too
- Using a HashSet for removed nodes, why is this needed? :-)   The "pop" it is doing is also not constant time -- OrderMap::pop() which is constant time comes to mind :-P
  - That said, simpler data structure like a sorted vector?
- We will look for a compact matrix graph where edges that don't exist are stored as zeros and so on, so without the Option overhead. Is it infeasible to make this that data structure? It seems like it could support both zero vs nonzero edge weights as None vs Some edge weights.

src/matrix.rs
```diff
@@ -0,0 +1,1145 @@
...
+
+    #[inline]
+    fn to_edge_position(&self, a: NodeIndex, b: NodeIndex) -> usize {
+        to_linearized_matrix_position::<Ty>(a.index(), b.index(), self.node_capacity)
+    }
+
+    /// Remove all nodes and edges
+    #[inline]
+    pub fn clear(&mut self) {
+        *self = Matrix::default();
```
bluss: Clear does not typically for a collection deallocate its capacity, it maintains it.

src/matrix.rs
```diff
@@ -0,0 +1,1145 @@
...
+            () => {
+                self.node_identifiers()
+                    .filter(|n| n.index() < self.node_capacity)
+            }
+        };
+
+        let out_edges_to_remove: Vec<_> = node_ids!()
+            .map(|n| self.to_edge_position(a, n))
+            .filter(|&p| self.node_adjacencies[p].is_some())
+            .collect();
```
bluss: this is a temporary vector we need to remove. Same below.

bluss: Oh, this is remove_node. It's not so important, then.

jmcomets:
>  AdjacencyMatrixGraph or MatrixGraph work for me too

Will amend this commit with a renamed version then, MatrixGraph is shorter to type and just as clear so I think I'll go with that.

> Using a HashSet for removed nodes, why is this needed?

The `IdStorage` is a bit like a heap with fixed block size:

- to allocate, you move data to a block and receive an address for constant-time access
- to deallocate, you add the address to a free list, optionally clearing the block (this can be done when allocating as well)
- to iterate over the elements in the heap, for each adress in the adress space you yield the allocated blocks (eg. not in the free list)

Now, the free list could be a queue to speed up deallocation, but that comes at the cost of iteration, which needs to have a fast lookup in the free list.

In summary, the free list needs to have the following as fast as possible: insertion, removal, lookup. As far as the `MatrixGraph` goes, lookup needs to be the fastest as insertion/removal are already costly operations for the adjacency matrix, whereas lookup is used massively when looking at a node's neighborhood.

I'll take a look at `OrderMap` and run some benchmarks to see if it fits. ;)

> We will look for a compact matrix graph where edges that don't exist are stored as zeros and so on, so without the Option overhead. Is it infeasible to make this that data structure? It seems like it could support both zero vs nonzero edge weights as None vs Some edge weights.

After thinking about this for quite a bit, storing the edge weights as `Option<E>` inside the array that represents the adjacency matrix seems like the best way to go.

For zero-sized edge weights the `Option` degrades to a single byte, for nullable types the size is unaffected, and there's no extra indirection so accessing the edge weights can be very cache-compliant. The only big issue I see with this method is that it works well as long as edge weights are ~small. If edge weights are larger, using a `Box<E>` seems appropriate.

jmcomets:
I've switched to `OrderMap` which has an explicit `pop()` operation, but I need to run a few more benchmarks that heavily abuse `remove_node` and `neighbors` in order to have a definitive opinion on this part.

src/matrix_graph.rs
```diff
@@ -0,0 +1,1147 @@
...
+
+    #[inline]
+    fn to_edge_position(&self, a: NodeIndex, b: NodeIndex) -> usize {
+        to_linearized_matrix_position::<Ty>(a.index(), b.index(), self.node_capacity)
+    }
+
+    /// Remove all nodes and edges
+    #[inline]
+    pub fn clear(&mut self) {
+        *self = MatrixGraph::default();
```
bluss: In a typical datastructure, clear should remove all contents but not drop the capacity. That said, we can wait with fixing this until more whole picture issues are resolved.

jmcomets: I've just pushed a `clear()` which should keep the capacity as is. Note that it's `O(V^2)` because the entire adjacency matrix needs to have each of its entries set to `None`, although it does allow much faster node / edge addition afterwards.

bluss: Right, that's certainly a drawback, but it's expected.

bluss:
This is a really solid data structure, thanks for your work, and sorry that I don't have so much time to look closely at it.

What I want us to reach is an adajcency matrix that can be used with no overhead for edges, like the very common one where 0 means no edges and nonzero means an edge with that weight. I think we could have both? We don't need to make this the default behavior for floats or integers, but we could make it opt in -- so a matrix graph of `MatrixGraph<Something, NonZero<f64>>` would be one where the edges are nonzero and non-edges are represented by 0f64.  Does that idea make sense? `NonZero<T>` and `Option<T>` could both implement the same trait.

jmcomets:
I understand the idea, but doesn't `Option<T>` have no overhead with non-zero types?

[This example](https://play.rust-lang.org/?gist=df7d3351122dadbed6203aeff31a6cca&version=stable) passes no problem:

```rust
use std::mem;

fn main() {
    macro_rules! check_option_overhead {
        ($t:ty) => {
            assert_eq!(mem::size_of::<Option<$t>>(), mem::size_of::<$t>());
        }
    }
    
    check_option_overhead!(&str);
    check_option_overhead!(Vec<bool>);
}
```

jmcomets:
I think I understand what you mean. I don't see how we could do this without specialization though. 😞

bluss:
I don't think we should use std NonZero. Just a confusing name I picked. `struct NotZero<T>(T)`.

jmcomets:
FYI I'm working on this [here](https://github.com/jmcomets/petgraph/commit/1e50481b77737ae8dabacdc404cf1aa0647689ae). I'm not quite satisfied with the trait's API yet, but is that what you were looking for?

bluss:
Yes! I think the trait can be simplified. I'll have time to look at this next week.

jmcomets:
@WaDelma This PR was pretty much ready to merge. I've just rebased it and the tests should pass easily. The `Build` trait isn't implemented but since it's a bit of a hassle to implement, it can be done in a further PR.

waywardmonkeys:
@jmcomets I suspect you meant your last comment for me. I will review this this week.

waywardmonkeys:
@jmcomets Regarding the `NonZero` / `Option` discussion ... I think `rustc` itself has improved some in this area, no?

jmcomets:
Sorry about that, autocomplete failed me 😄 

Regarding the state of rustc for the `NonZero$type`, things seem to have moved forward since Rust 1.28 (August 2018). I've done some work to support this myself on [this branch](https://github.com/jmcomets/petgraph/tree/adjacency-matrix-v2-not-zero-opt-in), plugging the standard types shouldn't prove too difficult. Maybe you were thinking about removing the extra trait and only supporting the standard types?

bluss:
That looks awesome @jmcomets - using a trait or just supporting the built NonZero types, either way seems fine, I guess the goal is to support both for example Option<T> and NonZeroU32 — though in the standard types I'd be missing the NonZero float types?

jmcomets:
Yeah I think I'm going to specifically go for the trait option, otherwise we'll be missing the float types. I'll include the patch for this in this PR.

jmcomets:
I've added extra docs and reworked the `Zero`/`Nullable` traits. Things are looking pretty good now, I'm pretty confident about merging this.

[Current docs](http://docs.jmcomets.com/petgraph/matrix_graph/index.html)

src/matrix_graph.rs
```diff
@@ -169,12 +170,27 @@ not_zero_impls!(u8, u16, u32, u64, usize);
...
+/// `MatrixGraph<N, E, Ty, Null>` is a graph datastructure using an adjacency matrix
+/// representation.
+///
+/// `MatrixGraph` is parameterized over:
+///
+/// - Associated data `N` for nodes and `E` for edges, called *weights*.
+///   The associated data can be of arbitrary type.
+/// - Edge type `Ty` that determines whether the graph edges are directed or undirected.
+/// - Nullable type `Null`, which denotes the edges' presence (defaults to `Option<E>`). You may
+///   specify `NotZero<E>` if you want to use a sentinel value (such as 0) to mark the absence of
```
waywardmonkeys: Could you add a link to `NotZero` docs here?

src/matrix_graph.rs
```diff
@@ -393,26 +412,45 @@ impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> MatrixGraph<N, E, Ty, Null>
         self.node_adjacencies[p].as_mut().unwrap()
     }
 
+    /// Return an iterator of all nodes with an edge starting from `a`.
+    ///
+    /// - `Directed`: Outgoing edges from `a`.
+    /// - `Undirected`: All edges from or to `a`.
+    ///
+    /// Produces an empty iterator if the node doesn't exist.<br>
+    /// Iterator element type is `NodeIndex<Ix>`.
```
waywardmonkeys: And a link to the docs for `NodeIndex` here?

src/matrix_graph.rs
```diff
@@ -450,6 +488,16 @@ impl<N, E, Ty: EdgeType, Null: Nullable<Wrapped=E>> MatrixGraph<N, E, Ty, Null>
...
+    /// Return an iterator of all neighbors that have an edge between them and
+    /// `a`, in the specified direction.
+    /// If the graph's edges are undirected, this is equivalent to *.neighbors(a)*.
+    ///
+    /// - `Directed`, `Outgoing`: All edges from `a`.
+    /// - `Directed`, `Incoming`: All edges to `a`.
+    /// - `Undirected`: All edges from or to `a`.
+    ///
+    /// Produces an empty iterator if the node doesn't exist.<br>
+    /// Iterator element type is `NodeIndex<Ix>`.
```
waywardmonkeys: And here...

src/matrix_graph.rs
```diff
@@ -458,6 +506,14 @@ impl<N, E, Null: Nullable<Wrapped=E>> MatrixGraph<N, E, Directed, Null> {
...
     }
 
+    /// Return an iterator of all edges of `a`, in the specified direction.
+    ///
+    /// - `Directed`, `Outgoing`: All edges from `a`.
+    /// - `Directed`, `Incoming`: All edges to `a`.
+    /// - `Undirected`: All edges connected to `a`.
+    ///
+    /// Produces an empty iterator if the node `a` doesn't exist.<br>
+    /// Iterator element type is `EdgeReference<E, Ix>`.
```
waywardmonkeys: And link to `EdgeReference` here?

waywardmonkeys:
I agree that this is looking great. :)

src/matrix_graph.rs
waywardmonkeys: You did link this in the next commit, sorry. But the other changes are still valid.

jmcomets:
I added a bunch of links to the referenced structs / traits & updated a wrongly written doc for `remove_edge()`.

jmcomets:
After some dogfooding:
- I added the impl for `Index`/`IndexMut` using an edge.
- I renamed `with_nodes` to `with_capacity`, which is clearer as all it does is pre-allocate.
- I got around to injecting the `Ix` type argument instead of hardcoding it to `u16` (though it's still the default).

bluss:
:rocket:  I guess we need to round up any straggling breaking changes we want to make and head for a 0.5 release

PR #178, opened 2017-10-15
Add initial dominance_frontiers implementation

m4b:
**WIP**

@bluss need your opinion/recommendation on:

1. The return type of dominance frontiers
2. The signature, e.g., whether:

     a. `dominance_frontiers` should be `dominance_frontier(&self, graph: G, node: G::NodeId) -> DominanceFrontier` (i.e., returns frontier for a given node (ideally the frontier is an iterator)) _OR_,

     b. the current impl `dominance_frontiers(&self, graph: G) -> DominanceFrontiers` which computes the frontier for every node in the graph, given the dominator computation in `&self`

For 1, I don't see an immediately obvious way to return an iterator instead of a vector/set for the frontier; would appreciate any insight you have here.

**EDIT**

For 1b, I tried this initially but I couldn't see how to correctly implement the algorithm as defined in that paper, since frontiers are inserted essentially out of order w.r.t. the current node in the graph node iteration, iiuc.

The algo is here:

```
for all nodes, b
  if the number of predecessors of b ≥ 2
     for all predecessors, p, of b
       runner ← p
       while runner != doms[b]
         add b to runner’s dominance frontier set
         runner = doms[runner]
```

### DIFF
--- a/src/algo/dominators.rs
+++ b/src/algo/dominators.rs
@@ -12,10 +12,13 @@
 //! strictly dominates **B** and there does not exist any node **C** where **A**
 //! dominates **C** and **C** dominates **B**.
 
+
+use std::iter::FromIterator;
 use std::collections::{HashMap, HashSet};
 use std::hash::Hash;
 
-use visit::{DfsPostOrder, GraphBase, IntoNeighbors, Visitable, Walker};
+use visit::{DfsPostOrder, GraphBase, IntoNeighbors, IntoNeighborsDirected, Visitable, IntoNodeIdentifiers, VisitMap, Walker};
+use Direction;
 
 /// The dominance relation for some graph and root.
 #[derive(Debug, Clone)]
@@ -75,6 +78,52 @@ impl<N> Dominators<N>
         } else {
             None
         }
+    }
+
+    /// Returns the dominance frontier of every node in `G`.
+    ///
+    /// From Cytron et. al, the dominance frontier is defined as:
+    /// > the set of all cfg nodes, y, such that b dominates a predecessor of y but does not strictly
+    /// > dominate y
+    ///
+    /// We use the algorithm detailed here: https://www.cs.rice.edu/~keith/EMBED/dom.pdf
+    pub fn dominance_frontiers<G>(&self, graph: G) -> HashMap<G::NodeId, Vec<G::NodeId>>
+        where
+        <G as Visitable>::Map: VisitMap<N>,
+        G: IntoNeighborsDirected + IntoNodeIdentifiers + IntoNeighbors + Visitable + GraphBase<NodeId = N>,
+        <G as IntoNeighborsDirected>::NeighborsDirected: Clone,
+        <G as GraphBase>::NodeId: Eq + Hash + Ord {
+        let mut frontiers = HashMap::<G::NodeId, Vec<G::NodeId>>::from_iter(graph.node_identifiers().map(|v| (v, vec![])));
+
+        for node in graph.node_identifiers() {
+            let (predecessors, predecessors_len) = {
+                let ret = graph.neighbors_directed(node, Direction::Incoming);
+                let count = ret.clone().count();
+                (ret, count)
+            };
+
+            if predecessors_len >= 2 {
+                for p in predecessors {
+                    let mut runner = p;
+
+                    match self.immediate_dominator(node) {
+                        Some(dominator) => {
+                            while runner != dominator {
+                                frontiers.entry(runner).or_insert(vec![]).push(node);
+                                runner = self.immediate_dominator(runner).unwrap();
+                            }
+
+                        },
+                        None => ()
+                    }
+                }
+                for (_, frontier) in frontiers.iter_mut() {
+                    frontier.sort();
+                    frontier.dedup();
+                }
+            }
+        }
+        frontiers
     }
 }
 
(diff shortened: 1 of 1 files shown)


bluss:
ah I forgot, I want to ping them here. @fitzgen: What do you think of this?

m4b:
I think I actually prefer the `(graph, node) -> Frontier` version, instead of computing frontier for every node in the graph (e.g., it's more lazy).

This would also allow us ideally not to allocate the vecs and the hashmap.

But the algorithm at least in that paper seems to preclude this possibility, since:

1. nodes for each frontier are constructed out of order
2. nodes can be duplicated (and hence must be deduped)

Consequently, it's not immediately clear how to do this without a HashMap from node refs to vectors of node  refs

fitzgen:
I haven't dealt with dominance frontiers before. My experience with dominators has been with heap graph algorithms and memory tooling rather than compilers and SSA construction, which is my understanding of what frontiers are usually useful for.

That's a long way of saying that I'm not sure I have anything particularly insightful to add here :)

In general, if we can do smaller amounts of work at a time, that seems like a better API so that people who don't need everything don't have to compute everything. So, assuming there is no fundamental reason we can't compute frontiers on a per-node basis, then we should do frontiers on a per-node basis.

> the algorithm as defined in that paper,

Which paper?

m4b:
Ah sorry, should have posted here, same as in the doc for dominators, paper by Cooper et. al:

https://www.cs.rice.edu/~keith/EMBED/dom.pdf

m4b:
I should also add, in case it’s being considered for non inclusion, that dominance frontiers are mentioned at the top of the wiki article, right after definition of immediate dominator, so I think it’s a pretty natural thing to expect from the dominator module:

https://en.m.wikipedia.org/wiki/Dominator_(graph_theory)

fitzgen:
I agree that it is a natural fit, I just don't have too much to contribute.

Thinking about it some more, I don't think there is an easy way to incrementalize the dominance frontiers. The Cytron et al algorithm seems a little more amenable, but not by much: we'd still need to consider the whole subtree of the given node in the dominator tree.

Since the simple-fast dominance frontiers algorithm wants predecessor sets, it makes sense to reuse the onces already created inside the `simple_fast` function. But that is a bit annoying if one doesn't intend to use domincance frontiers at all, an only cares about dominators themselves. Throw generics at the problem? `Dominators<WithFrontiers>` and `Dominators<WithoutFrontiers>`? I don't find that particularly satisfying :-/

The Cytron algorithm, on the other hand, only needs the graph and the dominators.

bluss:
Hi -- I haven't had the time to come back to this yet. Feel free to ping me later.

m4b:
Doesn't seem like this is getting reviewed/progress getting made, closing

bluss:
Yes, it's not in my bandwidth — we need other contributors for reviews.

src/algo/dominators.rs
```diff
@@ -76,6 +79,52 @@ impl<N> Dominators<N>
...
         }
     }
+
+    /// Returns the dominance frontier of every node in `G`.
+    ///
+    /// From Cytron et. al, the dominance frontier is defined as:
+    /// > the set of all cfg nodes, y, such that b dominates a predecessor of y but does not strictly
+    /// > dominate y
+    ///
+    /// We use the algorithm detailed here: https://www.cs.rice.edu/~keith/EMBED/dom.pdf
```
fitzgen: This doc comment should mention big-O runtime complexity.

src/algo/dominators.rs
```diff
@@ -76,6 +79,52 @@ impl<N> Dominators<N>
...
+    pub fn dominance_frontiers<G>(&self, graph: G) -> HashMap<G::NodeId, Vec<G::NodeId>>
+        where
+        <G as Visitable>::Map: VisitMap<N>,
+        G: IntoNeighborsDirected + IntoNodeIdentifiers + IntoNeighbors + Visitable + GraphBase<NodeId = N>,
+        <G as IntoNeighborsDirected>::NeighborsDirected: Clone,
+        <G as GraphBase>::NodeId: Eq + Hash + Ord {
```
fitzgen: Nitpick: the `{` should go on a new line.

src/algo/dominators.rs
fitzgen: It should also mention that `graph` must be the same, un-mutated graph that this `Dominators` instance was constructed from.

src/algo/dominators.rs
```diff
@@ -76,6 +79,52 @@ impl<N> Dominators<N>
...
     }
+
+    /// Returns the dominance frontier of every node in `G`.
+    ///
+    /// From Cytron et. al, the dominance frontier is defined as:
+    /// > the set of all cfg nodes, y, such that b dominates a predecessor of y but does not strictly
+    /// > dominate y
+    ///
+    /// We use the algorithm detailed here: https://www.cs.rice.edu/~keith/EMBED/dom.pdf
+    pub fn dominance_frontiers<G>(&self, graph: G) -> HashMap<G::NodeId, Vec<G::NodeId>>
```
fitzgen: I think we should return a newtype over the hash map, `DominanceFrontiers`, and expose semantically meaningful operations on that newtype.

src/algo/dominators.rs
fitzgen: Some more context on *why* one might be interested in dominance frontiers would also be nice, ie because they represent join points in a CFG.

PR #187, opened 2017-12-06
Fix markdown warning with current Rust.

waywardmonkeys:


waywardmonkeys:
The current version of Rust has warnings when the rendering of the docs differs when using the upcoming Pulldown-based markdown rendering. This fixes that and should render the same with both Markdown implementations.

bluss:
Thanks!

PR #191, opened 2017-12-19
Use size hints when building graphs at once

sanmai-NL:
When deserializing or calling `.from_elements()`, we can do better than to assume an empty graph as result. This change results in fewer memory allocations.


sanmai-NL:
```rust
pub trait FromElements : Create {
    fn from_elements<I>(iterable: I) -> Self
```
What is the value of this trait? It’ll potentially allocate a lot of memory to keep duplicate information, i.e. the output the graph as well the input vector of elements. What graphs are not indexable? What am I missing?

bluss:
I don't really understand the drawback you see. It is true that the trait is mostly unused in petgraph.

As for the PR, can't we just use the lower bound (like Vec's FromIterator does).

bluss:
Or in fact, since we don't have any good info about the size hint of the graph I'd just let it be.

sanmai-NL:
It’s a very reliable heuristic to assume it won’t turn out be an empty graph. 🙂  I’m fine with the lower bound too -- but I can’t tell of the top off my head how that’s calculated. 

We know the total number of vertices plus edges, so the best I think we can do to avoid extra allocations is to assume the total numbers of nodes as well as edges are equal to the number of elements, and then finally the excess memory (just [`gr.shrink_to_fit()`](https://docs.rs/petgraph/0.4.5/petgraph/graph/struct.Graph.html#method.shrink_to_fit)).

bluss:
Good point. Lower_bound / 2 for nodes and edges sounds fine. My idea is usually if we don't know, we don't know; the default push/extend growth of a Vec must sometimes be good enough. In going with the lower bound, we assume the lower not the upper, so lower/ 2 seems ok with that.

If you end up with significant logic to compute the cap, I'd break it out into a function.

src/serde_utils.rs
```diff
@@ -50,7 +50,7 @@ impl<'de, F, T, R> Visitor<'de> for MappedSequenceVisitor<T, R, F>
     fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
         where A: SeqAccess<'de>,
     {
-        let mut v = Vec::new();
+        let mut v = Vec::with_capacity(seq.size_hint().unwrap_or_else(|| 0));
```
bluss: This one actually requires the most research. Is the sequence size hint untrusted user input? How do deserializers usually handle this. We only want to oom if the serialization input has size to match.

bluss: Ok, serde itself uses a cautious approach when deserializing to `Vec`. It seems like the right thing to do here too. In serde we can find they use this:

```rust
    #[inline]
    pub fn cautious(hint: Option<usize>) -> usize {
        cmp::min(hint.unwrap_or(0), 4096)
    }
```

sanmai-NL:
@bluss/@fitzgen: what is the way forward for this stale PR? Any advice to conclude it?

PR #200, opened 2018-01-31
Add a CONTRIBUTING file

fitzgen:
This makes explicit the standards that pull requests should live up to, and what should be included in them. Since this stuff wasn't already documented, I just took a guess. Let me know if you want anything changed!

Also, I've never really used rst before, but chose it because that's what the README was written in. If I've mis-formatted anything, let me know how to fix it.


bluss:
Hah, that's great

PR #206, opened 2018-03-17
Minor maintenance fixes

bluss:


bluss:
Again, thanks @cuviper for the workaround we use in travis to test using quickcheck on Rust 1.12 :-)

cuviper:
I hope that hack won't be my legacy... 😉

PR #209, opened 2018-03-19
Add support for pruning nodes from a DFS.

khuey:
Since you closed my other PR rather than allowing me to push changes to it ...

### DIFF
--- a/src/visit/dfsvisit.rs
+++ b/src/visit/dfsvisit.rs
@@ -96,7 +96,7 @@ impl<B> ControlFlow for Control<B> {
 impl<C: ControlFlow, E> ControlFlow for Result<C, E> {
     fn continuing() -> Self { Ok(C::continuing()) }
     fn should_break(&self) -> bool {
-        self.is_err()
+        if let Ok(ref c) = *self { c.should_break() } else { true }
     }
     fn should_prune(&self) -> bool {
         if let Ok(ref c) = *self { c.should_prune() } else { false }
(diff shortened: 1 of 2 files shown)


src/visit/dfsvisit.rs
```diff
@@ -60,26 +69,38 @@ impl<B> Control<B> {
...
+        }
+    }
 }
 
-impl<E> ControlFlow for Result<(), E> {
-    fn continuing() -> Self { Ok(()) }
+impl<C: ControlFlow, E> ControlFlow for Result<C, E> {
+    fn continuing() -> Self { Ok(C::continuing()) }
     fn should_break(&self) -> bool {
         if let Err(_) = *self { true } else { false }
```
bluss: We can pass through the "should_break" here like in should_prune. I think that's a good way to combine the power of Result with the control flow thing.

bluss:
This looks like a good addition, (What do others think?). Ready to go once the last 0.4.x release is out of the way.

jmcomets:
Looks great!

bluss:
It seems like we really only have micro changes unreleased in master, so I guess we can just roll in the changes for 0.5 in this PR and others

Dylan-DPC:
@bluss i think it is better to do it in a seperate release-tracking PR. I can do the release if you want.

bluss:
Ah yeah I don't know why I wrote it like that, I don't mean that we bring other changes into this PR. The 0.4.12 release is done already.. Tracking everything for 0.5 sounds good

bluss:
Thanks for this, and I'm sorry for the delay

khuey:
Thanks for merging this!

PR #212, opened 2018-04-13
Add comments to astar and bellman_ford

ksadorf:
Hi,

It adds ASCII representation for the graph of the astar example and also create an example for the bellman_ford algo.

However for the bellman_ford algo, I was not able to recreate the case where it should return Err(NegativeCycle).
I think there is an error in the implementation because the following example does not work

```rust
use petgraph::Graph;
use petgraph::algo::bellman_ford;
use petgraph::algo::NegativeCycle;

let graph_with_neg_cycle = Graph::<(), f32>::from_edges(&[
     (0, 1, -2.0),
     (0, 3, -4.0),
     (1, 2, -1.0),
     (1, 5, -25.0),
     (2, 4, -5.0),
     (4, 5, -25.0),
     (3, 4, -1.0),
]);

assert_eq!(bellman_ford(&graph_with_neg_cycle, a),Err(NegativeCycle));
```

This code do not even compile because of something related to NegativeCycle but I was not able to make compile.

Should I create tests for the bellman_ford in order to catch this?

Noé

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -122,7 +122,7 @@ pub fn toposort<G>(g: G, space: Option<&mut DfsSpace<G::NodeId, G::Map>>)
                         }
                         if !dfs.discovered.is_visited(&succ) {
                             dfs.stack.push(succ);
-                        } 
+                        }
                     }
                 } else {
                     dfs.stack.pop();
@@ -321,7 +321,7 @@ pub fn tarjan_scc<G>(g: G) -> Vec<Vec<G::NodeId>>
 
     #[derive(Debug)]
     struct Data<'a, G>
-        where G: NodeIndexable, 
+        where G: NodeIndexable,
           G::NodeId: 'a
     {
         index: usize,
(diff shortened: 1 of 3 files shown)


carrutstick:
Graphs default to using directed edges, in which case your example has no negative cycle. Try running the following instead:
```rust
    use petgraph::Graph;
    use petgraph::algo::bellman_ford;

    let graph_with_neg_cycle = Graph::<(), f32, Undirected>::from_edges(&[
         (0, 1, -2.0),
         (0, 3, -4.0),
         (1, 2, -1.0),
         (1, 5, -25.0),
         (2, 4, -5.0),
         (4, 5, -25.0),
         (3, 4, -1.0),
    ]);

    let a = NodeIndex::new(0);
    assert!(bellman_ford(&graph_with_neg_cycle, a).is_err());
```

ksadorf:
Thanks for pointing this!
I'm still not used to Directed as default.

I added your example for the error.

bluss:
thanks!

PR #213, opened 2018-04-16
Optimize is isomorphic

### LINKED ISSUE #210: is_isomorphic/is_isomorphic_matching stack overflows on large graphs.
I couldn't figure out how to replicate other than to use my crate [Chelone](https://github.com/Aaronepower/Chelone). I have
a binary in the crate that checks the equality of two Turtle documents.
[This implementation](https://github.com/Aaronepower/Chelone/blob/9db365581baa4e5d29a3ee8a82d83df02f625b2e/src/triple.rs#L21-L35) hashes the RDF graph converts into a `Graph` and
calls `algo::is_isomorphic`. Using the `test/data/manifest.ttl` file which
is a large file containing 4344 nodes and 2172 edges.

### Steps To Reproduce

```
cargo run --bin is_equal tests/data/manifest.ttl tests/data/manifest.ttl
```

### Actual Results

```
thread 'main' has overflowed its stack
fatal runtime error: stack overflow
[1]    5969 abort      cargo run --bin is_equal tests/data/manifest.ttl tests/data/manifest.ttl
```
### Expected Results

To not stack overflow.

jrraymond:
Fix #210 by replacing recursive algorithm with iterative implementation. Added a test case for large graphs on which the previous recursive implementation overflowed the stack (I failed actually duplicate the bug on the original input (compilation errors, didn't look into them too much); instead I wrote a test case on which the recursive implementation overflowed the stack).


src/isomorphism.rs
```diff
@@ -173,7 +173,7 @@ pub fn is_isomorphic<N, E, Ty, Ix>(g0: &Graph<N, E, Ty, Ix>,
     }
 
     let mut st = [Vf2State::new(g0), Vf2State::new(g1)];
-    try_match(&mut st, g0, g1, &mut NoSemanticMatch, &mut NoSemanticMatch).unwrap_or(false)
+    return try_match(&mut st, g0, g1, &mut NoSemanticMatch, &mut NoSemanticMatch).unwrap_or(false);
```
carrutstick: I think idiomatic rust does not use a `return` unless strictly necessary.

jrraymond:
hmm. The iterative version is more memory efficient but also slower. I should fix the performance regression before this is merged.

jrraymond:
I think the performance difference may be due to allocating the vector from the heap. The implementations are the same except for the iterative version explicitly maintains the stack with a `Vec` on the heap as opposed to using the call stack. If I pre-allocate the vec the performance of the iterative version is slightly better for the praust benchmark.
```
test iso_100n_100e_iter            ... bench:      21,250 ns/iter (+/- 2,159)
test iso_100n_100e_rec             ... bench:      20,312 ns/iter (+/- 335)
test praust_undir_no_iso_bench     ... bench:   1,048,081 ns/iter (+/- 11,552)
test praust_undir_no_iso_rec_bench ... bench:   1,103,181 ns/iter (+/- 5,885)
```
Where as if I don't reserve then the performance of the iterative version is worse.
```
test iso_100n_100e_iter            ... bench:      21,799 ns/iter (+/- 3,434)
test iso_100n_100e_rec             ... bench:      20,169 ns/iter (+/- 384)
test praust_undir_no_iso_bench     ... bench:   1,078,678 ns/iter (+/- 157,112)
test praust_undir_no_iso_rec_bench ... bench:   1,074,249 ns/iter (+/- 17,892)
```
The iteractive version is also much noisier (perhaps also related to allocating on the heap?).

bluss:
@jrraymond Prealloc sounds more than good enough. Do you have the code for doing that, is there a heuristic for depth?

jrraymond:
I benchmarked a preallocating version and still didn't see an improvement. I'll have to dig into the performance issue further when I get the chance.

bluss:
Let's merge this improvement first. It's really great that you've worked on this

Issue #214, opened 2018-04-23
The Dot generator does not escape backslashes, causes rendering issues


sbstp:
The issue can be seen inside the Regex object in these graphs (nodes and edges).

Here's the unedited version:

```dot
digraph {
    0 [label="PatternInfo { pattern: DontMatch, group: None }"]
    1 [label="PatternInfo { pattern: Regex(\w+), group: Some(\"series\") }"]
    2 [label="PatternInfo { pattern: Regex(^s\d\d?e\d\d?$), group: Some(\"season_episode\") }"]
    3 [label="PatternInfo { pattern: Regex(s\d\d?), group: Some(\"season\") }"]
    4 [label="PatternInfo { pattern: Regex(e\d\d?), group: Some(\"episode\") }"]
    5 [label="PatternInfo { pattern: Regex(\d\d?), group: Some(\"season\") }"]
    6 [label="PatternInfo { pattern: Regex(\d\d?), group: Some(\"episode\") }"]
    0 -> 1 [label="PatternInfo { pattern: Regex(\w+), group: Some(\"series\") }"]
    1 -> 1 [label="PatternInfo { pattern: Regex(\w+), group: Some(\"series\") }"]
    1 -> 2 [label="PatternInfo { pattern: Regex(^s\d\d?e\d\d?$), group: Some(\"season_episode\") }"]
    1 -> 3 [label="PatternInfo { pattern: Regex(s\d\d?), group: Some(\"season\") }"]
    3 -> 4 [label="PatternInfo { pattern: Regex(e\d\d?), group: Some(\"episode\") }"]
    1 -> 5 [label="PatternInfo { pattern: Regex(\d\d?), group: Some(\"season\") }"]
    5 -> 6 [label="PatternInfo { pattern: Regex(\d\d?), group: Some(\"episode\") }"]
}
```
And the image generated with this: https://i.imgur.com/hX8W6JK.png

Here's a version that I have manually escaped
```dot
digraph {
    0 [label="PatternInfo { pattern: DontMatch, group: None }"]
    1 [label="PatternInfo { pattern: Regex(\\w+), group: Some(\"series\") }"]
    2 [label="PatternInfo { pattern: Regex(^s\\d\\d?e\\d\\d?$), group: Some(\"season_episode\") }"]
    3 [label="PatternInfo { pattern: Regex(s\\d\\d?), group: Some(\"season\") }"]
[...]

PR #215, opened 2018-04-24
Support more escape codes inside labels 

sbstp:
Fix for #214 

Now escapes `|`, `&`, `<`, `>` and `\`. Turned out to be a bit more tricky than I anticipated, but it's all good now.

### DIFF
--- a/src/dot.rs
+++ b/src/dot.rs
@@ -29,7 +29,7 @@ use visit::{GraphRef};
 /// println!("{:?}", Dot::with_config(&graph, &[Config::EdgeNoLabel]));
 ///
 /// // In this case the output looks like this:
-/// // 
+/// //
 /// // digraph {
 /// //     0 [label="\"A\""]
 /// //     1 [label="\"B\""]
@@ -171,10 +171,10 @@ impl<W> fmt::Write for Escaper<W>
 
     fn write_char(&mut self, c: char) -> fmt::Result {
         match c {
-            '"' => try!(self.0.write_char('\\')),
+            '"' | '\\' => try!(self.0.write_char('\\')),
             // \l is for left justified linebreak
-            '\n' => return self.0.write_str(r#"\l"#),
-            _   => { }
+            '\n' => return self.0.write_str("\\l"),
+            _ => {}
         }
         self.0.write_char(c)
     }
@@ -188,7 +188,7 @@ impl<T> fmt::Display for Escaped<T>
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if f.alternate() {
-            write!(&mut Escaper(f), "{:#}\\l", &self.0)
+            write!(&mut Escaper(f), "{:#}\n", &self.0)
         } else {
             write!(&mut Escaper(f), "{}", &self.0)
         }
@@ -204,4 +204,14 @@ impl<T> fmt::Display for DebugFmt<T>
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         self.0.fmt(f)
     }
+}
+
+#[test]
+fn test_escape() {
+    let mut buff = String::new();
+    {
+        let mut e = Escaper(&mut buff);
+        let _ = e.write_str("\" \\ \n");
+    }
+    assert_eq!(buff, "\\\" \\\\ \\l");
 }
--- a/tests/graph.rs
+++ b/tests/graph.rs
@@ -1420,12 +1420,13 @@ fn dot() {
         b: &'static str,
     };
     let mut gr = Graph::new();
-    let a = gr.add_node(Record { a: 1, b: "abc" });
+    let a = gr.add_node(Record { a: 1, b: r"abc\" });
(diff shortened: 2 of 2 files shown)


src/dot.rs
```diff
@@ -171,10 +171,13 @@ impl<W> fmt::Write for Escaper<W>
 
     fn write_char(&mut self, c: char) -> fmt::Result {
         match c {
-            '"' => try!(self.0.write_char('\\')),
+            '"' | '\\' | '|' | '<' | '>' | '&' => self.0.write_char('\\')?,
             // \l is for left justified linebreak
-            '\n' => return self.0.write_str(r#"\l"#),
-            _   => { }
+            '\n' => {
+                self.0.write_char('\\')?;
```
Dylan-DPC: error: the `?` operator is not stable (see issue #31436)

src/dot.rs
```diff
@@ -171,10 +171,13 @@ impl<W> fmt::Write for Escaper<W>
 
     fn write_char(&mut self, c: char) -> fmt::Result {
         match c {
-            '"' => try!(self.0.write_char('\\')),
+            '"' | '\\' | '|' | '<' | '>' | '&' => self.0.write_char('\\')?,
```
Dylan-DPC: error: the `?` operator is not stable (see issue #31436)

Dylan-DPC:
Hey. Thanks for the PR

We need to support the stable rust version and thus have to depend on stable features as much as possible.

sbstp:
`?` is stable since [1.13 (2016-11-10)](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1130-2016-11-10),  but it wasn't in 1.12, so I've updated the PR match that. Everything should be fine now.

From reading the spec again, I don't think escaping `<`, `>` and `&` is required because HTML is not parsed inside strings delimited by quotes.

Dylan-DPC:
Yeah that's what i meant. Maybe it is time to bump up the minimum rustc version soon.

src/dot.rs
```diff
@@ -171,10 +171,13 @@ impl<W> fmt::Write for Escaper<W>
...
     fn write_char(&mut self, c: char) -> fmt::Result {
         match c {
-            '"' => try!(self.0.write_char('\\')),
+            '"' | '\\' | '|' | '<' | '>' | '&' => try!(self.0.write_char('\\')),
             // \l is for left justified linebreak
-            '\n' => return self.0.write_str(r#"\l"#),
-            _   => { }
+            '\n' => {
+                try!(self.0.write_char('\\'));
+                return self.0.write_char('l');
```
bluss: Why did you change from write_str to 2 write_chars here? I guess the difference is trivial, but normally write_str should be better w.r.t code quality.

src/dot.rs
```diff
@@ -188,7 +191,7 @@ impl<T> fmt::Display for Escaped<T>
 {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if f.alternate() {
-            write!(&mut Escaper(f), "{:#}\\l", &self.0)
+            write!(&mut Escaper(f), "{:#}\n", &self.0)
```
bluss: Is there a reason for this change?

bluss:
Bumping the minimum rust version sounds good for the next version (and thus current master). Not something super new, but for example Rust 1.18 maybe. That can be a different PR.

Here, it sounds like we should adjust the escaping to match the spec and don't escape those extra chars then?

src/dot.rs
sbstp: Yes, since backslashes are escaped now, `\\l` turns to `\\\\l` which is not a the intent. Since `\\l` turns to `\n`, I put the converted form there directly to avoid escaping `\\l`.

src/dot.rs
sbstp: I had double escaping issues at some point because I was using `self.write_str`, but I think it might be fine if I use `self.0.write_str`.

src/dot.rs
bluss: ah, of course. nice

sbstp:
The spec is not very detailed, they don't talk about escaping `|` for instance, but it renders differently when escaped. Escaping the html `<`, `>` and `&` seems not to be necessary but it does not affect rendering since graphviz ignores lone backslashes. So I think I'll remove the html escape for keep the backslashes and the pipe symbol.

src/dot.rs
sbstp: I got a little confused there. `\\l` would turn to `\\\\l`, and in order to get `\\l` in the output, `\n` is used.

sbstp:
The pipe is relevant when a [node is in record mode](https://www.graphviz.org/doc/info/shapes.html#record). Since the Dot generator does not use record mode, I believe it's safe not to escape it.

sbstp:
Mixing Debug with escaping can be surprisingly confusing. I left a comment for the next person maintaining this.

sbstp:
The nightly build seems to have an issue which is not related to PR so I think this is finally ready.

bluss:
Thanks for this!

PR #217, opened 2018-05-12
Implement GraphML output for graphs

jonasbb:
This PR implements GraphML output for graphs. It is a XML based format, thus it add the new dependency on xml-rs, although I put everything behind feature flags.

This PR is not in a final state, as nothing is documented it. I will add some documentation, but I want to get some feedback on the API design first.

You can see some examples for usage in the included tests. The definition of the function which converts the weights (for nodes and edges) into attributes is a bit complex, but I don't really see a better way.

The functions have to be (`PrintWeight` type definition):

```rust
for<'a> Fn(&'a W) -> Vec<(Cow<'static, str>, Cow<'a, str>)>
```

The reasons for this are:
1. Attributes are arbitrary key-value pairs in GraphML, therefore the function has to return a vector.
2. It should be possible to generate the attribute lists, thus owned values have to be returned.
3. I wanted to avoid dynamic allocations if possible, otherwise a simpler `Vec<(String, String)>` would suffice as return type.

Right now there is a `to_string() -> String` and a `to_writer(writer: Writer) -> Result<_>` method to produce the output.

I'd love to get some feedback on this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ quickcheck = { optional = true, version = "0.4", default-features = false }
 ordermap = { version = "0.3.0", optional = true }
 serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }
+xml-rs = { version = "0.7.0", optional = true }
 
 [dev-dependencies]
 rand = "0.3"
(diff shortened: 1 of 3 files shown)


bluss:
Hi, that looks useful. I would prefer if this was published in a separate crate. As in: building on top of petgraph, not being a part of it.

jonasbb:
Oh sure, thanks for the feedback. I will put it in its own crate then. I thought it might be a fit, because petgraph already contains dot output functionality.

PR #219, opened 2018-05-17
Implement IntoNeighborsDirected for EdgeFiltered(#202)

### LINKED ISSUE #202: implement IntoNeighborsDirected for EdgeFiltered
### Summary

Is it possible to implement IntoNeighborsDirected for EdgeFiltered? Right now it does not seem to be implemented.

### Motivation

I wanted to use a `DfsPostOrder(Reversed(EdgeFiltered))` stack. One workaround, which I think does the same and compiles, is to use `DfsPostOrder(EdgeFiltered(Reversed))`. This is because Reversed requires IntoNeighborsDirected for EdgeFiltered, which is not provided. Is there a reason the first variant cannot work?

**EDIT:**
Actually the 2nd variant can't be used either, but for a different reason - IntoNeighbors is not implemented for Reversed.

### Details

    I would guess, in `petgraph::visit::filter`:
    ```
    impl<'a, G, F> IntoNeighborsDirected for &'a EdgeFiltered<G, F> where ... { ... }
    ```

    Nothing new as far as I'm aware

  implement it?

   I could try implementing, but I'm kind of lost navigating all the instances, probably not ready to contribute

kettlebell:
Hi everyone, I've implemented the trait to address #202 and added a test as well.

Fixes #202

### DIFF
--- a/src/visit/filter.rs
+++ b/src/visit/filter.rs
@@ -348,7 +348,7 @@ where
         EdgeFilteredNeighborsDirected {
             iter: self.0.edges_directed(n, dir),
             f: &self.1,
-            dir,
+            dir: dir,
         }
     }
 }
--- a/src/visit/filter.rs
+++ b/src/visit/filter.rs
@@ -349,6 +349,7 @@ where
             iter: self.0.edges_directed(n, dir),
             f: &self.1,
             dir: dir,
+            from: n,
         }
     }
 }
@@ -433,6 +434,7 @@ where
     iter: G::EdgesDirected,
     f: &'a F,
     dir: Direction,
+    from: G::NodeId,
 }
 
 impl<'a, G, F> Iterator for EdgeFilteredNeighborsDirected<'a, G, F>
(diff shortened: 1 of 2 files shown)


jmcomets:
Thanks for this! I have a few small style issues with your impl but otherwise it looks good to me.

kettlebell:
Thank you for your feedback. My implementation was adapted from the existing `EdgeFilteredNeighbors` struct, which also calls `filter_map` within each `.next`.

I've tried to manually create an instance of `FilterMap` up front and get `EdgeFilteredNeighborsDirected` to take ownership of it, but I get problems with the closure type parameter of `FilterMap` being unsized.

src/visit/filter.rs
```diff
@@ -409,6 +425,40 @@ impl<'a, G, I, F> Iterator for EdgeFilteredEdges<'a, G, I, F>
...
+    type Item = G::NodeId;
+    fn next(&mut self) -> Option<Self::Item> {
+        let f = self.f;
+        let dir = self.dir;
+        (&mut self.iter)
+            .filter_map(move |edge| {
+                if f.include_edge(edge) {
+                    match dir {
+                        Direction::Incoming => Some(edge.source()),
+                        Direction::Outgoing => Some(edge.target()),
```
bluss: Is this producing the right neighbors for undirected graphs? I'm not sure it does. We need to follow this:

To follow the doc: https://docs.rs/petgraph/0.4.13/petgraph/visit/trait.IntoNeighborsDirected.html

The underlying IntoEdgesDirected, for undirected graphs, should already produce edges where the source is the starting node (`n` in the `neighbors_directed` new method in this PR) and the target is the other edge node.

kettlebell: Thanks. I've got it working for undirected graphs without having to work with the IntoNeighborsDirected trait. What I do is persist the starting node `n`, and then look at both the source and target of each edge; the node that `!= n` is the neighbor, except when it's an edge to itself.

bluss:
Thanks for this! I'll merge and push a micro cleanup on top of it

jgrund:
Will this be included in a published release soon?

bluss:
We are merging breaking changes and can release 0.5 when it is done. It is basically done

PR #224, opened 2018-08-26
Fix conflict with now-stable method Iterator::find_map

### LINKED ISSUE #223: Fails to compile on nightly-2018-08-26
```console
$ rustc --version
rustc 1.30.0-nightly (39e6ba821 2018-08-25)

$ cargo build
error[E0034]: multiple applicable items in scope
    --> src/graph_impl/stable_graph/mod.rs:1171:19
     |
1171 |         self.iter.find_map(|(i, node)| {
     |                   ^^^^^^^^ multiple `find_map` found
     |
note: candidate #1 is defined in an impl of the trait `iter_utils::IterUtilsExt` for the type `_`
    --> src/iter_utils.rs:5:5
     |
5    | /     fn find_map<F, R>(&mut self, mut f: F) -> Option<R>
6    | |         where F: FnMut(Self::Item) -> Option<R>
7    | |     {
8    | |         for elt in self {
...    |
13   | |         None
14   | |     }
     | |_____^
     = note: candidate #2 is defined in an impl of the trait `std::iter::Iterator` for the type `std::iter::Enumerate<_>`
```

Looks like there has been a warning about this method since 1.27.0.

```
warning: a method with this name may be added to the standard library in the future
    --> src/graph_impl/stable_graph/mod.rs:1171:19
     |
1171 |         self.iter.find_map(|(i, node)| {
     |                   ^^^^^^^^
     |
     = note: #[warn(unstable_name_collisions)] on by default
     = warning: once this method is added to the standard library, the ambiguity may cause an error or change in behavior!
     = note: for more information, see issue #48919 <https://github.com/rust-lang/rust/issues/48919>
     = help: call with fully qualified syntax `iter_utils::IterUtilsExt::find_map(...)` to keep using the current method
     = note: add #![feature(iterator_find_map)] to the crate attributes to enable `std::iter::Iterator::find_map`
```

bluss:
To quick fix this without changing rust version, rename our local
extension methods to use an ex_ prefix.

Fixes #223

### DIFF
--- a/src/graph_impl/stable_graph/mod.rs
+++ b/src/graph_impl/stable_graph/mod.rs
@@ -1168,7 +1168,7 @@ impl<'a, N, Ix> Iterator for NodeReferences<'a, N, Ix>
     type Item = (NodeIndex<Ix>, &'a N);
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.iter.find_map(|(i, node)| {
+        self.iter.ex_find_map(|(i, node)| {
             node.weight.as_ref().map(move |w| (node_index(i), w))
         })
     }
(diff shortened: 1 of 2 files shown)


bluss:
travis: test compile is broken on Rust 1.12 but not the plain build, shipping.

PR #227, opened 2018-10-30
upgrade bincode, indexmap, quickcheck and rand

### LINKED ISSUE #225: Update dependencies
Hey!

we're considering packaging petgraph in debian, but it seems both the ordermap and quickcheck dependencies are out of date. It seems the quickcheck dependency can be bumped without any source changes, the ordermap update requires some changes but I'm it seems I'm not familiar enough with it to provide a patch.

Thanks!

zimond:
This PR upgrades several critical dependencies for petgraph. Hopes to resolve #225 

Note: Currently quickcheck 0.7 depends on `rand-core` instead of `rand`, this makes `rand` a hard dependency for petgraph. So I changed `quickcheck` feature to include an optional `rand`. This could be a break change.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -198,7 +198,7 @@ Recent Changes
   - ``GraphMap`` can now have directed edges. ``GraphMap::new`` is now generic
     in the edge type. ``DiGraphMap`` and ``UnGraphMap`` are new type aliases.
   - Add type aliases ``DiGraph, UnGraph, StableDiGraph, StableUnGraph``
-  - ``GraphMap`` is based on the ordermap crate. Deterministic iteration
+  - ``GraphMap`` is based on the indexmap crate. Deterministic iteration
     order, faster iteration, no side tables needed to convert to ``Graph``.
   - Improved docs for a lot of types and functions.
   - Add graph visitor ``DfsPostOrder``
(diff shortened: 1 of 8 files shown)


daiheitan:
oh seems we need to modify travis config to pass this PR

bluss:
Thanks a lot, this is certainly overdue

README.rst
```diff
@@ -308,7 +308,7 @@ Recent Changes
   - Add Graph::capacity(), GraphMap::capacity()
   - Fix bug in Graph::reverse()
   - Graph and GraphMap have `quickcheck::Arbitrary` implementations,
-    if optional feature `quickcheck` is enabled.
+    if optional feature `check` is enabled.
```
bluss: I'll have to think about this, this seems wrong. We'd like to keep the name of this feature flag.

bluss: We want to fix our quickcheck code (inside src, not the tests) so that we use the randomness api that quickcheck is reexporting.

zimond: I'm not familiar with quickcheck. So I think I' better leave this to you or anyone who's interested.

bluss:
Note, quickcheck is a public dependency, so bumping it is what makes this a breaking change, primarily.

zimond:
@bluss  currently `quickcheck` only depends on `RngCore`, which do not have a `gen_range()` api that petgraph relies on. It seems that quickcheck encourages users to implement their own arbitrary type for ranges randoms, but that needs a `rand` dependency. Could you please share your thoughts on how to change the quickcheck support code in petgraph? I really want this version bump to be settled.

bluss:
I've looked at it, I think we can use my fix. Ill push it when I'm home

bluss:
Thanks! Merging, but will apply fixes to quickcheck/rand immediately, rand is not becoming a public dependency.

PR #228, opened 2018-10-30
Pub edges_undirected

zimond:
This PR makes `Graph::edges_undirected` and `StableGraph::edges_undirected` public. I didn't find any reason that they are private. Also modified misleading doc on `edges` to match its actual behavior.

### DIFF
--- a/src/graph_impl/stable_graph/mod.rs
+++ b/src/graph_impl/stable_graph/mod.rs
@@ -635,10 +635,7 @@ impl<N, E, Ty, Ix> StableGraph<N, E, Ty, Ix>
         }
     }
 
-    /// Return an iterator of all edges of `a`.
-    ///
-    /// - `Directed`: Outgoing edges from `a`.
-    /// - `Undirected`: All edges connected to `a`.
+    /// Return an iterator of all outgoing edges of `a`.
     ///
     /// Produces an empty iterator if the node doesn't exist.<br>
     /// Iterator element type is `EdgeReference<E, Ix>`.
@@ -648,9 +645,8 @@ impl<N, E, Ty, Ix> StableGraph<N, E, Ty, Ix>
 
     /// Return an iterator of all edges of `a`, in the specified direction.
     ///
-    /// - `Directed`, `Outgoing`: All edges from `a`.
-    /// - `Directed`, `Incoming`: All edges to `a`.
-    /// - `Undirected`: All edges connected to `a`.
+    /// - `Outgoing`: All edges from `a`.
+    /// - `Incoming`: All edges to `a`.
     ///
     /// Produces an empty iterator if the node `a` doesn't exist.<br>
     /// Iterator element type is `EdgeReference<E, Ix>`.
@@ -672,7 +668,7 @@ impl<N, E, Ty, Ix> StableGraph<N, E, Ty, Ix>
     ///
     /// Produces an empty iterator if the node `a` doesn't exist.<br>
     /// Iterator element type is `EdgeReference<E, Ix>`.
-    fn edges_undirected(&self, a: NodeIndex<Ix>) -> Edges<E, Ty, Ix> {
+    pub fn edges_undirected(&self, a: NodeIndex<Ix>) -> Edges<E, Ty, Ix> {
         Edges {
             skip_start: a,
             edges: &self.g.edges,
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -800,10 +800,7 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
         }
     }
 
-    /// Return an iterator of all edges of `a`.
-    ///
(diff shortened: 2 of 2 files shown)


bluss:
I don't agree with pub-ing the methods, these methods are not good for the public API, they are quirky and hard to use. We don't want to support them.

src/graph_impl/stable_graph/mod.rs
```diff
@@ -635,10 +635,7 @@ impl<N, E, Ty, Ix> StableGraph<N, E, Ty, Ix>
         }
     }
 
-    /// Return an iterator of all edges of `a`.
-    ///
-    /// - `Directed`: Outgoing edges from `a`.
-    /// - `Undirected`: All edges connected to `a`.
+    /// Return an iterator of all outgoing edges of `a`.
```
bluss: I'm curious, in what way was the old doc misleading? It's important to note how the directedness (or not) of the graph's edges influences the behaviour here.

jmcomets:
The user can definitely get the same behavior by chaining two calls to `edges_directed` like so:

```rust
let edges = graph.edges_directed(node, Direction::Outgoing)
                 .chain(graph.edges_directed(node, Direction::Incoming));   
```

I approved because it just seems like less code to write from the user's perspective. :man_shrugging:

jmcomets:
Ah indeed! Though the semantic of `edges_directed` for an undirected graph are pretty unintuitive to me.

No issues with your comments BTW, brutal honesty is better than fluff. :)

src/graph_impl/stable_graph/mod.rs
daiheitan: I get it now. So this doc means that for different kinds of graphs, `edges` has different behavior. I misunderstood that

daiheitan:
I was never aware of chaining `edges_directed` to get the same result as `edges_undirected`. Indeed these methods are confusing. Here's my proposal:

1. remove `edges_directed`
2. change `edges(NodeIndex)` to `edges(a: NodeIndex, direction: Direction)`, with documentation stating that :
    - for undirectional graph, `direction` has no effect
    - for directed graph, this method behaves as expected. you could chain the result iterator to get all edges

jmcomets:
I'm not in favor of this API, `edges_directed` allows specifying what kind of edges you want to view, whereas `edges` gives the default (outgoing edges when directed, all edges when undirected).

While `edges_undirected` is clunky because you get edges where the base node is either the source or the target, it does fill a missing use-case: iterating over all edges of a node (including weights). I see several ways of addressing this:

1. Having a public `edges_undirected` method which yields pairs of `(neighbor, edge weight)`, which mirrors `neighbors_undirected`.
2. Adding documentation under `edges`/`edges_directed` suggesting the use of `.chain(...)` on directed graphs and the general clunkiness of such a solution.
3. As originally suggested, pub-ing `edges_undirected`, adding a more detailed note on how edges are yielded.
4. Specializing `edges_directed` to only directed graphs, and pub-ing `edges_undirected` for undirected graphs. This has severe implications on some algorithms such as `DFSPostOrder` or anything involving the `Reversed` adaptor. (Try making [this branch][1] compile just for fun.)

To summarize:

1. new method
2. just documentation
3. use existing method, documentation
4. rely on type system, large API change

[1]: https://github.com/jmcomets/petgraph/tree/no_directed_methods_for_undirected_graphs

daiheitan:
Having an API named as `edges_directed` on an undirected graph is a little unnatural. The first time when I was using petgraph and found the APIs, my thought was (I was building a directed graph):

1. `edges_directed`: This API gives directional edges connected to a node. And the API does what it says. Nice.
2. `edges`: With the context that I already found an API named as `edges_directed`, I thought this API gives all edges, which means edges in two directions.
3. Found that `edges` in fact gives only outgoing edges.
4. Check the doc.

This catches me from time to time, it's very subtle and I wrote quite a few bugs because of `::edges()`. I don't know what's the best solution here. At least let's add some documentation to tell users that there is a way to get all edges (chaining), and emphasize that `edges` gives only outgoing edges.

bluss:
> Having an API named as edges_directed on an undirected graph is a little unnatural.

Great, then we can agree to not expose edges_undirected and make it worse :smile: 

> emphasize that edges gives only outgoing edges.

I think that this is expressed in the doc being removed in this PR. We have: `Directed`: Outgoing edges from `a`, `Undirected`: All edges connected to `a`.  Non-directed graphs don't have "outgoing" edges, so we just say "edges connected to".

The way we call it is also `graph.edges(node)`. The implicit part here is that it's "edges from `node`" (for example it's not "edges related to `node`". For directed and undirected graphs, the interpretation should be unambiguous as long as that is clear.

zimond:
@bluss Fine, I think the confusion comes when `Graph` could be either directed or undirected. In other popular graph libs like python's networkx, they are separated. Anyway, that's beyond this PR. I think all we could do know is letting users know that you could chain on `edges_directed`. I' closing this PR.

BTW could you please check PR #227 ?

PR #229, opened 2018-11-01
Add an immutable Graph::node_weights() iterator

dylanmckay:
This adds a new helper function to `Graph` that allows users to
immutably iterate through all node weights.

This is similar to the existing `Graph::node_weights_mut()` function.

This fills a tiny gap in the API - in most cases, every mutable accessor is
accompanied by an immutable accessor of the same name. This is so callers from
immutable contexts can use the same API as the corresponding logic but in
an mutable context..

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -939,6 +939,15 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
         NodeIndices { r: 0..self.node_count(), ty: PhantomData }
     }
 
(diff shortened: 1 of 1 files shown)


bluss:
This is covered generically by https://docs.rs/petgraph/0.4.13/petgraph/visit/trait.IntoNodeReferences.html#tymethod.node_references which gives you a combination of node id and weight for each node.

bluss:
We don't want a new trait just for the convenience (we have some organizational "overhead" due to each trait we define in petgraph's visit module, just look at how they are defined and implemented) but other ways to solve it could be approached

PR #232, opened 2018-12-04
Implement contains_edge for StableGraph

abreis:
This PR implements `fn contains_edge() -> bool` for `StableGraph`.

While the implementation is trivial, `contains_edge()` exists in `Graph`, so this helps bring `StableGraph` up to par.


bluss:
Thanks!

PR #233, opened 2018-12-04
Add missing impl to EdgeIndex

abreis:
`NodeIndex` in `Graph` has the following trait impl:
```rust
impl<Ix: IndexType> From<Ix> for NodeIndex<Ix> {
    fn from(ix: Ix) -> Self { NodeIndex(ix) }
}
```

This PR adds the same `impl From<Ix>` to `EdgeIndex`, which it did not have.


bluss:
Thanks!

PR #234, opened 2018-12-07
Don't suggest use of usize instead of u32

hellow554:
usize machine dependent and on x86 machines it is the same as `u32`. Instead use a machine independent size e.g. `u64` or `u128`.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -303,7 +303,7 @@ impl<E, Ix: IndexType> Edge<E, Ix>
 /// take its place. Similarly, removing an edge shifts the index of the last edge.
 ///
 /// The `Ix` parameter is `u32` by default. The goal is that you can ignore this parameter
-/// completely unless you need a very big graph -- then you can use `usize`.
+/// completely unless you need a very big graph -- then you can use `u64` or `u128`.
 ///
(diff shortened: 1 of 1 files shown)


hellow554:
The failure is not my fault 👼

bluss:
This change should be reverted, since it doesn't make any sense.

`usize` implements `IndexType` - so users can use usize. The two mentioned types - u64 and u128, don't implement IndexType unless I'm missing something, so it's going to be frustrating advice if the docs tell the user to use that.

hellow554:
But then I ask myself why does `u32` implement `IndexType`? Wouldn't it be easier to implement it for `u8`, `u16`, ... `u128` instead of reverting this pull?

bluss:
Petgraph is not perfect, but I guess there is some kind of knowledge baked into the choices that we have made and documented, and it's good to both take a step back and ask, why was it designed like this, and respect that, while also thinking of how to improve it or expand it.

The docs say - we prefer u32 to keep the size of the graph, in memory, down. The data structure contains a lot of indices internally. Smaller data in memory means faster running algorithms.

One reason to not implement for u128 is that no Rust platform supports vectors (std::vec::Vec) with that many elements, so it's pointless. We are limited by the biggest usize.

PR #236, opened 2019-01-06
Add `equiv()` function for more easily readable equivalence testing

nicholas-miklaucic:
A [different union-find implementation](https://docs.rs/disjoint-sets/0.4.2/disjoint_sets/struct.UnionFind.html#method.equiv) has a method called `equiv()` for determining whether two elements are equivalent (that is, they belong to the same set.) 

I think this is a useful addition to the library:

 - It is more easily readable than its implementation, and it's more immediately obvious what it does
 - It usefully abstracts exactly how it is implemented
 - Unlike `union()`, which already implements this, it has no side-effects and is named more consistently with that functionality

There's a test that simply duplicates the test above it, just using `equiv()` instead of `assert_eq!`

### DIFF
--- a/src/unionfind.rs
+++ b/src/unionfind.rs
@@ -92,6 +92,12 @@ impl<K> UnionFind<K>
         }
     }
 
+    /// Returns `true` if the given elements belong to the same set, and returns
+    /// `false` otherwise.
+    pub fn equiv(&self, x: K, y: K) -> bool {
+        self.find(x) == self.find(y)        
+    }
+
 
     /// Unify the two sets containing `x` and `y`.
     ///
--- a/tests/unionfind.rs
+++ b/tests/unionfind.rs
@@ -34,6 +34,34 @@ fn uf_test() {
 }
 
 #[test]
+fn uf_test_with_equiv() {
+    let n = 8;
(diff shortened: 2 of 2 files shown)


nicholas-miklaucic:
That makes sense. I'll note that [the disjoint set implementation I referenced earlier](https://github.com/tov/disjoint-sets-rs) has the advantage of having basically the same API as petgraph: I was using that crate, but because I need petgraph in the future for other things I combined the dependencies. The `equiv()` calls were the only thing I had to change besides imports.

I wonder how common my use case actually is, though: using disjoint sets for something other than graph algorithms.

src/unionfind.rs
```diff
@@ -92,6 +92,12 @@ impl<K> UnionFind<K>
         }
     }
 
+    /// Returns `true` if the given elements belong to the same set, and returns
```
bluss: For future prs, just a small nit, prefer active language in the doc (So "Return...")

nicholas-miklaucic:
Thanks for the info,  didn't know that!

bluss:
Oh, derp, this PR isn't even merged yet. i had assumed it was. I'm on it..

PR #243, opened 2019-03-04
Add topology sorting

thinkrapido:
I suggest Kahn's algorithm for topology sorting. (https://en.wikipedia.org/wiki/Topological_sorting)

Please don't hesitate to give honest feedback. (I'm still a novice rustician)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ debug = true
 
 [dependencies]
 fixedbitset = { version = "0.1.4" }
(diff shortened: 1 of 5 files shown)


bluss:
be advised that it already exists as `petgraph::algo::toposort` - are there important differences?

thinkrapido:
My bad. Havn't seen the function.
Reject this PR.
I'm going to read and use your solution.
If you're willing to give me feedback on my solution. You're welcome.

PR #248, opened 2019-04-28
chore: Update quickcheck to 0.8

ignatenkobrain:


waywardmonkeys:
I ran the same tests locally as Travis and it seems like this should work.

waywardmonkeys:
And .. it didn't work.

PR #249, opened 2019-05-02
add `node_weights_mut` to `StableGraph`

endeav0r:


bluss:
The interface of this method is not complete (it needs to be an iterator of &mut N, not &mut Option<N>, so more work and a separate iterator method is needed.

XVilka:
@endeav0r do you still wish to continue this PR? Rebase and address the feedback?

ABorgna:
This method was added in #332.

PR #254, opened 2019-05-31
Rust 2018 transition

XVilka:
Updating to the [Rust 2018 edition](https://doc.rust-lang.org/edition-guide/rust-2018/index.html).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ repository = "https://github.com/bluss/petgraph"
 keywords = ["data-structure", "graph", "unionfind", "graph-algorithms"]
 categories = ["data-structures"]
 
+edition = "2018"
+
 [lib]
 
 name = "petgraph"
--- a/src/algo/dominators.rs
+++ b/src/algo/dominators.rs
@@ -15,7 +15,7 @@
 use std::collections::{HashMap, HashSet};
 use std::hash::Hash;
 
-use visit::{DfsPostOrder, GraphBase, IntoNeighbors, Visitable, Walker};
+use crate::visit::{DfsPostOrder, GraphBase, IntoNeighbors, Visitable, Walker};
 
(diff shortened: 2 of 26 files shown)


XVilka:
Ping?

waywardmonkeys:
The minimum Rust version has now been bumped to 1.37 for the 0.5 release. I'd like to visit things related to Rust 2018 in a bit-by-bit fashion, with separate commits for each type of change.

XVilka:
Will split into separate changes once PR is green.

XVilka:
@waywardmonkeys ok, it is done - green, split and ready for review!

jmcomets:
Is there a way we could add lints for these? I'm specifically thinking of things like the `use crate::` syntax that tend to creep back.

XVilka:
@jmcomets there are two things that can be done - [setting up `cargo fmt`](https://github.com/bluss/petgraph/issues/269) and running `cargo clippy` (there are many warnings btw).

Cargo.toml
```diff
@@ -29,7 +29,7 @@ debug = true
 
 [dependencies]
 fixedbitset = { version = "0.1.4" }
-quickcheck = { optional = true, version = "0.8", default-features = false }
+quickcheck = { optional = true, version = "0.8.5", default-features = false }
```
waywardmonkeys: Does this need to be 0.8.5? or will any 0.8 work?

XVilka: Yes, 0.8 will work too. I just saw that there was a quickcheck version bump, but not to the latest one. I don't see why not to bump to the latest quickcheck for the new release though. You can drop this commit if you don't want to.

waywardmonkeys: Okay. If you can drop that commit from this PR, we can get this landed.

waywardmonkeys:
Thanks!

PR #259, opened 2019-08-07
Fix bug where Csr::add_node adds existing edges to new node

noctune:
`add_node` used 0 as column offset, causing all existing edges to be added to the new node.


bluss:
Thanks for finding this

PR #261, opened 2019-08-11
Remove Travis support for 1.15, bump minimum to 1.25.

waywardmonkeys:
Dependencies haven't built on Travis in some time for Rust 1.15.

Adjust the claimed minimum version to 1.25 which is what Travis
has been testing successfully.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: rust
 sudo: false
 matrix:
   include:
-    - rust: 1.15.0
-      env:
-      - ALL=' '
     - rust: 1.25.0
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
I think that this should actually bump the minimum version to 1.28 or even something more recent.

This would require that the next version published be in the 0.5 series.

Comments and thoughts welcome! (Especially from @bluss.)

bluss:
Thanks for looking after this — for the new release (0.5) I think you can pick minimum version freely, I'd go with something relatively recent.

PR #265, opened 2019-08-16
Bump minimum Rust version tested to 1.37.

waywardmonkeys:
Input on this from @bluss, @mitchmindtree, or others would be welcome.


waywardmonkeys:
Per comment by @bluss in #261, going to bump to the current stable.

PR #266, opened 2019-08-18
example showing how to find index of a node

ultrasaurus:
I'm submitting this as a step toward addressing https://github.com/bluss/petgraph/issues/258

I welcome feedback about whether this is the right approach 
or any suggested improvements.

### DIFF
--- a/examples/find_node.rs
+++ b/examples/find_node.rs
@@ -1,0 +1,19 @@
+use petgraph::Graph;
+
+// cargo run --example find_node
+
+fn main() {
+    println!("hello graph!");
+    let mut g = Graph::<&str, i32>::new();
+
+    // add nodes to graph
+    let items = ["apple", "book", "cat"];
+    for s in &items {
+      let i = g.add_node(s);
+      println!("added {:?} with index {:?}", s, i);
+    }
+
+    // maybe later you want to find the index of a node
+    let index = g.node_indices().find(|i| g.node_weight(*i) == Some(&"book")).unwrap();
+    println!("book index: {:?}", index);
+}
--- a/examples/find_node.rs
+++ b/examples/find_node.rs
@@ -1,3 +1,5 @@
+extern crate petgraph;
+
 use petgraph::Graph;
 
 // cargo run --example find_node


XVilka:
https://github.com/waywardmonkeys good addition for 0.5 release as well, in my opinion.

waywardmonkeys:
Instead of example files, I'd like to see this worked into the docs. It'll get tested then and it'll be present when people are reading the library docs.

examples/find_node.rs
```diff
@@ -0,0 +1,21 @@
...
+
+    // add nodes to graph
+    let items = ["apple", "book", "cat"];
+    for s in &items {
+      let i = g.add_node(s);
+      println!("added {:?} with index {:?}", s, i);
+    }
+
+    // maybe later you want to find the index of a node
+    let index = g.node_indices().find(|i| g.node_weight(*i) == Some(&"book")).unwrap();
```
bluss: In instructive examples I'd prefer if indexing notation was used instead of using `node_weight` - just to show a style preference and how to write simpler and easier to read code. For example here the closure can be `|i| g[*i] == "book"`.

But let's look at this from a higher level - searching all of node indices is frustrating and should only be done rarely. The example should make users aware of that we often save and hold on to node indices, and don't need to search for them. And when we do search, we often have more graph-specific searches and can use the graph traversals to find what we are looking for - not just linear search.

bluss: Maybe both approaches to finding an index can be shown

ultrasaurus: 👍 on the index notation `|i| g[*i] == "book".` -- for some reason I couldn't figure that out from the docs.... (I'm an experienced programmer, but new to Rust)

Agree that this shouldn't be the only example of finding nodes!  The example could put that more clearly in context.  As written, it certainly wouldn't be a good approach.

ultrasaurus:
Ok then, @waywardmonkeys comment with ❤️ by @bluss inspired me to learn how to build the documentation.  I haven't yet re-worked the example, but I thought I would start with a small change, and also check my understanding.... see https://github.com/bluss/petgraph/pull/273 for a small doc change that paves the way for a place to put this (I think).

bluss:
Closed in favour of #275 thanks

PR #273, opened 2019-08-21
attempt to clarify index documentation

ultrasaurus:
It seemed to be that "Send and Sync" safety isn't directly particularly related to the use of indices, so I moved that up.  It's true that "The Graph is a regular rust collection" is saying that it has indices, but this seemed like a reasonable overview for the introductory section.

On first reading, I wondered if "Pros and Cons of Indices" was about this particular Graph (as opposed to other kinds of graphs in the crate).  After I re-read a few times, I realized this was an attempt to draw attention to when you would or would not want to use indices.  I've adjusted the section headings and some formatting, which I think makes it more clear.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -262,6 +262,9 @@ impl<E, Ix: IndexType> Edge<E, Ix>
 /// - Edge type `Ty` that determines whether the graph edges are directed or undirected.
 /// - Index type `Ix`, which determines the maximum size of the graph.
 ///
+/// The `Graph` is a regular Rust collection and is `Send` and `Sync` (as long
+/// as associated data `N` and `E` are).
+///
 /// The graph uses **O(|V| + |E|)** space, and allows fast node and edge insert,
 /// efficient graph search and graph algorithms.
 /// It implements **O(e')** edge lookup and edge and node removals, where **e'**
@@ -296,16 +299,15 @@ impl<E, Ix: IndexType> Edge<E, Ix>
 /// example for *n* nodes indices are 0 to *n* - 1 inclusive.
 ///
 /// `NodeIndex` and `EdgeIndex` are types that act as references to nodes and edges,
-/// but these are only stable across certain operations.
-/// **Adding nodes or edges keeps indices stable.
-/// Removing nodes or edges may shift other indices.**
-/// Removing a node will force the last node to shift its index to
-/// take its place. Similarly, removing an edge shifts the index of the last edge.
+/// but these are only stable across certain operations:
+///
+/// * **Removing nodes or edges may shift other indices.** Removing a node will
+/// force the last node to shift its index to take its place. Similarly,
+/// removing an edge shifts the index of the last edge.
+/// * Adding nodes or edges keeps indices stable.
 ///
 /// The `Ix` parameter is `u32` by default. The goal is that you can ignore this parameter
 /// completely unless you need a very big graph -- then you can use `usize`.
-///
-/// ### Pros and Cons of Indices
 ///
 /// * The fact that the node and edge indices in the graph each are numbered in compact
 /// intervals (from 0 to *n* - 1 for *n* nodes) simplifies some graph algorithms.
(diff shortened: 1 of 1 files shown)


jmcomets:
I think the new version is much clearer 👍 Don't see any issue merging this.

src/graph_impl/mod.rs
```diff
@@ -262,6 +262,9 @@ impl<E, Ix: IndexType> Edge<E, Ix>
 /// - Edge type `Ty` that determines whether the graph edges are directed or undirected.
 /// - Index type `Ix`, which determines the maximum size of the graph.
 ///
+/// The `Graph` is a regular rust collection and is `Send` and `Sync` (as long
```
waywardmonkeys: Please capitalize "Rust" here.

ultrasaurus: Ah, I should have caught that when I copy/pasted the code.  Thanks will fix

ultrasaurus:
addressed comment -- Rust capitalized.  @waywardmonkeys pls take another look

waywardmonkeys:
Could you rebase this into a single commit and then I'll land it...

bluss:
Send and sync are a benefit of using indices, but whatever makes the docs best to read is fine

ultrasaurus:
@bluss it's not at all clear how I would use internal pointers or Arc shared pointers with Graph.  Are you referring to why it was good that Graph is implemented using indices internally?  or comparing it with another kind of Graph?  or am I totally missing something here?

bluss:
@ultrasaurus Yep, I'm comparing to other well-known ways of implementing graphs in general in Rust, for example using Arc or Rc

bluss:
Given the dicussion, maybe When to use / and when not to use is a strange formulation for the Graph doc? Since when you're using exactly the Graph/StableGraph types, you don't have a choice - you're using a graph datastructure based on this indexing implementation, and the interface uses indices. In that sense the benefits and drawbacks story might be easier to understand.

bluss:
By that reminds me, that if we update Graph we should probably update StableGraph as well :)

ultrasaurus:
My thinking on "when to use / when not to use" was really to call out the danger of removal.  Also partially to spark this conversation, testing my (lack of) understanding of what the items in the bulleted list are really about.  I would guess some notes are about why **the implementation** uses indices, not about how you would use the graph -- I think those kind of notes are very interesting, but should be in their own section, separated from **how to use** the graph API (which is typically what is covered in this kind of documentation).

Can someone shed a bit of light on the items in the bulleted list?  which ones are just FYI about the implementation (or what they are about if not that and not how to use this graph API)

bluss:
I think all the points you have retained are items that can be informative when using the datastructure. The Send/Sync thing is indeed mostly a result of the implementation, so it seems good to separate.

I think all of the points are kind of "things to be aware of", but in particular the point about removal can be formulated in those terms. I think it's good to be aware of how removal works - not that it shouldn't be used - then it wouldn't exist. 

Maybe we can do this?

```rust
/// ### Things to be aware of
/// * Indices don't allow as much compile time checking as references.
///
/// * Some graph indices shift during node or edge removal, which is a drawback
///   of `Graph`. For working with algorithms that perform removals, or if you need
///   graph indices to be stable under removals, consider using `StableGraph` instead.
///
```

bluss:
Yeah, the gist of "Indices don't allow as much compile time checking as references" is that the coder has to be careful when they use indices due to this lack of checking. They could use it as an argument against petgraph and picking something different, but the information is mostly something that's good-to-know and you can be careful about it?

bluss:
I think it's great to get fresh eyes on the documentation, but for me we still haven't gotten the context right here in this text — maybe I could rewrite the documentation instead.

If you read the StableGraph docs, it also has a list of points much in the same spirit, but with different details, is that much easier to read?

ultrasaurus:
StableGraph seems to have arbitrarily different text without solving the key issue I was attempting to resolve.  I've updated the PR to remove the headers that I added which didn't seem to be helpful, and mirror the lack of header above the bullets as seen in StableGraph.

So, this is updated to a smaller change which may be a small formatting improvement that may be easy to merge.  Feel free to apply or close, since I found a different place to put the [sample code](https://github.com/petgraph/petgraph/commit/5034ff57b566a8b57deb6a0661551f9e356313dd) that I was originally making a place for here.

If helpful to further improvements, the key questions I had when reading "Pros and Cons of Indices" was that for several bullet points it wasn't clear whether you were giving me guidance about:
* how to use indices when writing my code (e.g. be care with remove)
* whether to use this particular graph (in choosing between different petgraph options)
* the internal implementation and its impact on how my code will behave

In particular the following two were not particularly clear (at first read), yet also did not get in the way of my using the library. Leaving a note here in case anyone is motivated to provide further clarifications: 

* Indices don't allow as much compile time checking as references. 
   * are you suggesting the use of GraphMap which uses weight values as references if I want this? 
   * or is this just a warning that I need to be careful because an index reference could be to a node or an edge and this doesn't allow for the kind of compile-time checking as is typical in Rust programs?
* "Using indices allows mutation..." 
   * this caused a bit of a double-take since I was also warned about removal, but looking further into `Dfs` I saw that this is probably about changing edges

bluss:
Thanks!

PR #274, opened 2019-08-21
Remove UnionFind in favor of disjoint-sets crate

jmcomets:
Initially suggested in #236 

This replaces the current implementation of `UnionFind` with the use of the [disjoint-sets][] crate.

Note that ~~there will be~~ I expected *(see [latest benchmarks](https://github.com/bluss/petgraph/pull/274#issuecomment-523895858))* a performance hit since the current implementation [uses unchecked access][] while the new one [uses checked access][].

[disjoint-sets]: https://github.com/tov/disjoint-sets-rs
[uses unchecked access]: https://github.com/bluss/petgraph/blob/1b5c96d8f0eb4bae1a9e638a0dd358a8987c9cb4/src/unionfind.rs#L84
[uses checked access]: https://github.com/tov/disjoint-sets-rs/blob/88ab08df21f04fcf7c157b6e042efd561ee873ba/src/array.rs#L146


bluss:
I'm against this change since we have a good well tested and battle tested implementation, and as you say, this is a performance regression. It's easier to sell changes if it's just about positive changes.

XVilka:
I suggest to put these changes on hold and open an issue in the https://github.com/tov/disjoint-sets-rs

jmcomets:
I was able to run some benchmarks on my laptop which is running `rustc 1.36.0-nightly (73a3a90d2 2019-05-17)`.

I benchmarked the following tests, which use `UnionFind`:

```rust
#[bench]
fn connected_components_praust_undir_bench(bench: &mut test::Bencher) {
    let a: Graph<(), (), $dir> = parse_graph($graph_a);
    let b: Graph<(), (), $dir> = parse_graph($graph_b);

    bench.iter(|| {
        (connected_components(&a), connected_components(&b))
    });
}

#[bench]
fn is_cyclic_undirected_praust_undir_bench(bench: &mut test::Bencher) {
    let a: Graph<(), (), $dir> = parse_graph($graph_a);
    let b: Graph<(), (), $dir> = parse_graph($graph_b);

    bench.iter(|| {
        (is_cyclic_undirected(&a), is_cyclic_undirected(&b))
    });
}

#[bench]
fn min_spanning_tree_praust_undir_bench(bench: &mut test::Bencher) {
    let a: Graph<(), (), $dir> = parse_graph($graph_a);
    let b: Graph<(), (), $dir> = parse_graph($graph_b);

    bench.iter(|| {
        (min_spanning_tree(&a), min_spanning_tree(&b))
    });
}
```
- `$graph` being one of `PRAUST`, `FULL`, `PETERSEN`, taken from the [isomorphism benchmarks][]
- `$direction` being one of `Directed`, `Undirected`

The benches are submitted [here](https://github.com/jmcomets/petgraph/blob/607b65536c8befc0a4115cad4a0049b10af49a77/benches/unionfind.rs), if you want to run them yourself.

Here are the results:

|  test                                       |  before  |  after  |  diff  |
|  ----                                       |  ------  |  -----  |  ----  |
|  connected_components_full_dir_bench        |  1890    |  924    |  -966  |
|  connected_components_full_undir_bench      |  994     |  581    |  -413  |
|  connected_components_petersen_dir_bench    |  646     |  400    |  -246  |
|  connected_components_petersen_undir_bench  |  398     |  300    |  -98   |
|  connected_components_praust_dir_bench      |  1951    |  1018   |  -933  |
|  connected_components_praust_undir_bench    |  1064    |  610    |  -454  |
|  is_cyclic_undirected_full_dir_bench        |  86      |  89     |  3     |
|  is_cyclic_undirected_full_undir_bench      |  86      |  89     |  3     |
|  is_cyclic_undirected_petersen_dir_bench    |  175     |  134    |  -41   |
|  is_cyclic_undirected_petersen_undir_bench  |  210     |  149    |  -61   |
|  is_cyclic_undirected_praust_dir_bench      |  161     |  131    |  -30   |
|  is_cyclic_undirected_praust_undir_bench    |  166     |  132    |  -34   |
|  min_spanning_tree_full_dir_bench           |  441     |  437    |  -4    |
|  min_spanning_tree_full_undir_bench         |  293     |  299    |  6     |
|  min_spanning_tree_petersen_dir_bench       |  219     |  225    |  6     |
|  min_spanning_tree_petersen_undir_bench     |  169     |  175    |  6     |
|  min_spanning_tree_praust_dir_bench         |  438     |  446    |  8     |
|  min_spanning_tree_praust_undir_bench       |  284     |  290    |  6     |

- "before" column is taken from a [recent head of master][] (minus the semicolon fix)
- "after" column is taken from the [current head of remove-unionfind][]

For completeness, my laptop's CPU is a `Intel Core i5-7200U @ 2.50GHz`, and has 8 GB of DDR4 RAM.

I ran the benchmarks a few times and the results seem to be consistent, showing a major gain for `connected_components`, and a barely noticeable impact on `is_cyclic_undirected` / `min_spanning_tree`.

I haven't looked too much into the graphs themselves too understand if these benchmarks are really representative. Another thing that needs to be looked into is the impact for other types of graphs, though I don't believe we'll see anything surprising there.

[isomorphism benchmarks]: https://github.com/bluss/petgraph/blob/ec0ceea202c77c669b1a362d12b0c10e4bbf5859/benches/iso.rs
[recent head of master]: https://github.com/bluss/petgraph/tree/1b5c96d8f0eb4bae1a9e638a0dd358a8987c9cb4
[current head of remove-unionfind]: https://github.com/jmcomets/petgraph/tree/1b5c96d8f0eb4bae1a9e638a0dd358a8987c9cb4

XVilka:
Thanks for the benchmarks! I checked also the repository - seems like the project is abandoned - no updates for a year, which is a bad sign.

bluss:
Nice benchmarks. I had to look at it and see why the difference was so big. I'd like to just fix it on our side: #278  That would mean we are learning from their implementation which seems to be of high quality.

The IndexType dependency we have is one thing I'd use as another argument against removing our UnionFind. With that said, right now we don't actually use IndexType in an interesting way -- all the algorithms use `usize`, not the respective graphs' index types.

bluss:
I've merged the "status quo"ish prs that keep unionfind in-crate. I think we don't really gain so much by picking up a dependency here, but we could revisit the decision.

You're clearly right that an outside crate could have good implementation quality. The performance improvement in #278 shows that. On the whole, our implementation is well tested though — we have used it in petgraph for years.

jmcomets:
I'm ok with closing this since the crate I suggested doesn't seem actively maintained. I'll open a PR to add the benchmarks.

PR #277, opened 2019-09-02
Add all_simple_paths algorithm

fan-tom:
Add `all_simple_paths` algorithm, adopted from https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html

### DIFF
--- a/src/simple_paths.rs
+++ b/src/simple_paths.rs
@@ -8,7 +8,6 @@ use indexmap::IndexSet;
 use crate::{
     Direction::Outgoing,
     visit::{
-        GraphBase,
         IntoNeighborsDirected,
         NodeCount,
     },
@@ -18,14 +17,13 @@ use crate::{
 /// and at most `max_intermidiate_nodes`, if given, limited by graph's order otherwise
 /// Simple path is path without repetitions
 /// Algorithm is adopted from https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html
-pub fn all_simple_paths<'g, TargetColl: 'g, G>(graph: &'g G,
-                                               from: G::NodeId,
-                                               to: G::NodeId,
-                                               min_intermidiate_nodes: usize,
-                                               max_intermidiate_nodes: Option<usize>) -> impl Iterator<Item=TargetColl> + 'g
+pub fn all_simple_paths<TargetColl, G>(graph: G,
+                                       from: G::NodeId,
+                                       to: G::NodeId,
+                                       min_intermidiate_nodes: usize,
+                                       max_intermidiate_nodes: Option<usize>) -> impl Iterator<Item=TargetColl>
     where G: NodeCount,
-          G: GraphBase<NodeId=<&'g G as GraphBase>::NodeId>,
-          &'g G: IntoNeighborsDirected,
+          G: IntoNeighborsDirected,
           G::NodeId: Eq + Hash,
           TargetColl: FromIterator<G::NodeId>
 {
--- a/src/simple_paths.rs
+++ b/src/simple_paths.rs
@@ -15,10 +15,10 @@ use crate::{
 /// Simple path is path without repetitions
 /// Algorithm is adopted from https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html
 pub fn all_simple_paths<'g, TargetIter: 'g, N, E, Ty, Ix>(graph: &'g Graph<N, E, Ty, Ix>,
-                                                      from: NodeIndex<Ix>,
-                                                      to: NodeIndex<Ix>,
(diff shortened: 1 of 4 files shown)


src/simple_paths.rs
```diff
@@ -0,0 +1,155 @@
...
+            vec![0, 3, 4, 5],
+        ];
+
+        println!("{}", Dot::new(&graph));
+        let actual_simple_paths_0_to_5: Vec<Vec<_>> = all_simple_paths(&graph, 0u32.into(), 5u32.into(), 0, None)
+            .map(|v: Vec<_>| v.into_iter().map(|i| i.index()).collect())
+            .collect();
+        println!("{:#?}", actual_simple_paths_0_to_5);
+        assert_eq!(actual_simple_paths_0_to_5.len(), 8);
+        equal(expexted_simple_paths_0_to_5, &actual_simple_paths_0_to_5);
```
bluss: this should be assert_equal, not the equal function I think

src/simple_paths.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Returns iterator that produces all simple paths from `from` node to `to`, which contains at least `min_intermidiate_nodes` nodes
+/// and at most `max_intermidiate_nodes`, if given, limited by graph's order otherwise
+/// Simple path is path without repetitions
+/// Algorithm is adopted from https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html
+pub fn all_simple_paths<'g, TargetIter: 'g, N, E, Ty, Ix>(graph: &'g Graph<N, E, Ty, Ix>,
+                                                      from: NodeIndex<Ix>,
+                                                      to: NodeIndex<Ix>,
+                                                      min_intermidiate_nodes: usize,
+                                                      max_intermidiate_nodes: Option<usize>) -> impl Iterator<Item=TargetIter> + 'g
+    where Ix: IndexType, Ty: EdgeType, TargetIter: FromIterator<NodeIndex<Ix>>
```
bluss: The TargetIter type parameter should have a different name - it's not an iterator, so Iter should not be in the name.

src/simple_paths.rs
```diff
@@ -0,0 +1,155 @@
...
+    Graph,
+    graph::IndexType,
+    prelude::NodeIndex,
+};
+
+/// Returns iterator that produces all simple paths from `from` node to `to`, which contains at least `min_intermidiate_nodes` nodes
+/// and at most `max_intermidiate_nodes`, if given, limited by graph's order otherwise
+/// Simple path is path without repetitions
+/// Algorithm is adopted from https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html
+pub fn all_simple_paths<'g, TargetIter: 'g, N, E, Ty, Ix>(graph: &'g Graph<N, E, Ty, Ix>,
```
bluss: It looks like this algorithm can be implemented in terms of the generic visitor algorithms. I know working with Graph is much easier, but using the graph traits is more versatile and how we need to prefer to do it. You only need IntoNeighborsDirected, NodeCount and it should mostly be fine with that — I think, at least.

fan-tom:
I see that CI fails because `all_simple_paths` uses `indexmap` crate, which is under feature-gate, while `all_simple_paths` is not. Should I hide that function under the same feature-gate or make `indexmap` dependency not optional?

src/simple_paths.rs
```diff
@@ -0,0 +1,165 @@
...
+        IntoNeighborsDirected,
+        NodeCount,
+    },
+};
+
+/// Returns iterator that produces all simple paths from `from` node to `to`, which contains at least `min_intermidiate_nodes` nodes
+/// and at most `max_intermidiate_nodes`, if given, limited by graph's order otherwise
+/// Simple path is path without repetitions
+/// Algorithm is adopted from https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.simple_paths.all_simple_paths.html
+pub fn all_simple_paths<'g, TargetColl: 'g, G>(graph: &'g G,
```
bluss: Please substitute &'g G with plain G. Everywhere -- it's an amazing trick and means that `&g` and `Reversed(&g)` will both work, for example.

See other algo for examples.

bluss:
Nice. Using impl Iterator is a delight in petgraph, it's about time we adopt that. I'm not sure about if indexmap should be a default dependency, it sounds reasonable, its .pop() is that useful...

fan-tom:
Thanks, so then I'll make indexmap required dependency

bluss:
I'm undecided on indexmap, but the issue is a bit moot since graphmap is a default feature anyway, so we're not adding any compile time to the default configuration.

bluss:
Rebased

bluss:
Thanks!

PR #280, opened 2019-09-10
Use factories to run benches on sample graphs

jmcomets:
I refactored the benches to use a few factory methods to create basic graphs to run algorithms on. This removes the duplication of `parse_graph()`, `parse_stable_graph()` and the various `PETERSEN_A` (etc) constants. This makes writing benchmarks for algorithms pretty straightforward, but doesn't allow generating benchmarks for a set of sample graphs (although it's a step in that direction IMO).

While I was implementing this on a pretty old machine with barely 1GB RAM, lI also noticed that the benches in [`benches/ograph.rs`](https://github.com/petgraph/petgraph/blob/9a153c2a096c704d8439dd48f655f9e169956b5e/benches/ograph.rs) ran unbounded `add_node)()`/`add_edge()` loops which were crashing due to out-of-memory errors. I'll tackle that in a separate PR.

### DIFF
--- a/benches/factories.rs
+++ b/benches/factories.rs
@@ -1,0 +1,251 @@
+use std::marker::PhantomData;
+
+use petgraph::prelude::*;
+use petgraph::visit::NodeIndexable;
+use petgraph::data::Build;
+
+use petgraph::EdgeType;
+
+/// Petersen A and B are isomorphic
+///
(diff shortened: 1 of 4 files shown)


bluss:
I think you can move the file to `benches/common/factories.rs` if you need to avoid that the helper file itself is treated as a benchmark test. That strategy seems pretty robust where I've used it.

PR #282, opened 2019-09-29
GraphMap: include self loops in incoming edges

### LINKED ISSUE #281: Listing incoming nodes for digraphmap
### Test Case
```rust
use ::petgraph::{ graphmap::DiGraphMap, Direction};

fn main() {
    let mut graph: DiGraphMap<u32, u32> = DiGraphMap::new();

    let n0 = graph.add_node(0);
    let n1 = graph.add_node(1);
    let n2 = graph.add_node(2);

    /*
    Graph:
    0 -> 1
    1 -> 1
    1 -> 2
     */

    for (x, y) in [
        (n0, n1),
        (n1, n1),
        (n1, n2),
    ].iter() {
        graph.add_edge(*x, *y, 0);
    }

    println!("Graph edges:");
    for (src, dst, _) in graph.all_edges() {
        println!("{} -> {}", src, dst);
    }

    println!("Outgoing nodes:");
    for node in graph.nodes() {
        println!("Node: {:?}", node);
        for predecessor in graph.neighbors_directed(node, Direction::Outgoing) {
            println!("\toutgoing: {:?}", predecessor);
        }
    }

    println!("Incoming nodes:");
    for node in graph.nodes() {
        println!("Node: {:?}", node);
        for predecessor in graph.neighbors_directed(node, Direction::Incoming) {
            println!("\tincoming: {:?}", predecessor);
        }
    }
}

```

### Steps To Reproduce
Compile and run the example :smiley:

### Actual Results
```
Graph edges:
0 -> 1
1 -> 1
1 -> 2
Outgoing nodes:
Node: 0
        outgoing: 1
Node: 1
        outgoing: 1
        outgoing: 2
Node: 2
Incoming nodes:
Node: 0
Node: 1
        incoming: 0
Node: 2
        incoming: 1
```

### Expected Results
```
Graph edges:
0 -> 1
1 -> 1
1 -> 2
Outgoing nodes:
Node: 0
        outgoing: 1
Node: 1
        outgoing: 1
        outgoing: 2
Node: 2
Incoming nodes:
Node: 0
Node: 1
        incoming: 0
        incoming: 1
Node: 2
        incoming: 1
```

### Summary

For me there is a bug here: The "incoming" result should be symetric to the "outgoing" result. 

### Motivation

So if a node self loops, as the "outgoing" result displays the node self looping, the api should definitlely return the self looping node in the "incoming" result as well.

### Details

In the present case the calling parent is:
```rust
    pub fn neighbors_directed(&self, a: N, dir: Direction)
        -> NeighborsDirected<N, Ty>
    {
        NeighborsDirected {
            iter: match self.nodes.get(&a) {
                Some(neigh) => {
                    neigh.iter()
                },
                None => [].iter(),
            },
            dir: dir,
            ty: self.ty,
        }
    }
```
This code seems ok as if we look at the `neigh.iter()` for node 1, we have 3 neighbors (0, 1, 2).

If I understand correctly the code, in "src/graphmap.rs", the bug seems to be in:
```rust
            (&mut self.iter)
                .filter_map(move |&(n, dir)| if dir == self_dir {
                    Some(n)
                } else { None })
                .next()
```

The direction information for node 1 seems to be stored as an "outgoing" link between 1 and 1.
So when the iterators test the direction, it skips node 1 so we have the current result.

My first idea was to fix NeighborsDirected to add a test of the self node, but the iterator doesn't seem to have access to the original node :cry:

jmcomets:
After some fiddling with the tests, I believe the existing test was there to
prevent self loops being added twice for undirected graphs.

Fixes #281

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -294,7 +294,7 @@ impl<N, E, Ty> GraphMap<N, E, Ty>
     /// ```
     pub fn remove_edge(&mut self, a: N, b: N) -> Option<E> {
         let exist1 = self.remove_single_edge(&a, &b, Outgoing);
-        let exist2 = if a != b {
+        let exist2 = if a != b || Ty::is_directed() {
             self.remove_single_edge(&b, &a, Incoming)
         } else { exist1 };
         let weight = self.edges.remove(&Self::edge_key(a, b));
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -244,12 +244,13 @@ impl<N, E, Ty> GraphMap<N, E, Ty>
             self.nodes.entry(a)
                       .or_insert_with(|| Vec::with_capacity(1))
                       .push((b, CompactDirection::Outgoing));
-            if a != b {
-                // self loops don't have the Incoming entry
+
+            if a != b || Ty::is_directed() {
                 self.nodes.entry(b)
-                          .or_insert_with(|| Vec::with_capacity(1))
-                          .push((a, CompactDirection::Incoming));
+                    .or_insert_with(|| Vec::with_capacity(1))
+                    .push((a, CompactDirection::Incoming));
             }
+
             None
         }
     }
--- a/tests/graphmap.rs
+++ b/tests/graphmap.rs
@@ -323,3 +323,16 @@ fn undirected_neighbors_includes_self_loops() {
     assert_eq!(neighbors.next(), Some(node));
     assert_eq!(neighbors.next(), None);
 }
+
+#[test]
+fn self_loops_can_be_removed() {
+    let mut graph = DiGraphMap::new();
+
+    let node = graph.add_node(());
+    graph.add_edge(node, node, ());
+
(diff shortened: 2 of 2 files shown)


bluss:
I had thought that we need to fix this in the edges iterator and not change the data model. I haven't looked closely though. (It looks like edge removal needs update if this change is made.)

bluss:
I'm not sure why you think it's a bug that self loops don't have the incoming entry, just wanted clarification on that? I had to look closer since this changes the data model, but this way of solving it should work just as well as filtering it out in NeighborsDirected. We do the latter in Graph, so I'd be inclined for that solution.

bluss:
The alt-fix just for comparison is over in https://github.com/petgraph/petgraph/pull/new/graphmap-neighbors-directed

jmcomets:
Honestly, if we can fix this issue by changing the iterator rather than the data-structure, I'm all for it. :+1:

PR #294, opened 2019-11-08
docs (depth_first_search): note that it also works for Result<(),_>.

ratmice:
Just add a tiny note about Result here to the docs that wasn't obvious to me would work until I tried it.

### DIFF
--- a/src/visit/dfsvisit.rs
+++ b/src/visit/dfsvisit.rs
@@ -141,18 +141,23 @@ impl<B> Default for Control<B> {
 /// and edge classification of each reachable edge. `visitor` is called for each
 /// event, see [`DfsEvent`][de] for possible values.
 ///
-/// If the return value of the visitor is simply `()`, the visit runs until it
-/// is finished. If the return value is a `Control<B>`, it can be used to
-/// change the control flow of the search. `Control::Break` will stop
-/// the visit early, returning the contained value from the function.
+/// The return value should implement the trait `ControlFlow`, and can be used to change
+/// the control flow of the search.
+///
+/// `Control` Implements `ControlFlow` such that `Control::Continue` resumes the search.
+/// `Control::Break` will stop the visit early, returning the contained value.
 /// `Control::Prune` will stop traversing any additional edges from the current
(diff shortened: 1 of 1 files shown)


src/visit/dfsvisit.rs
```diff
@@ -117,10 +117,10 @@ impl<B> Default for Control<B> {
...
 /// event, see [`DfsEvent`][de] for possible values.
 ///
-/// If the return value of the visitor is simply `()`, the visit runs until it
-/// is finished. If the return value is a `Control<B>`, it can be used to
-/// change the control flow of the search. `Control::Break` will stop
-/// the visit early, returning the contained value from the function.
+/// If the return value of the visitor is simply `()` or `Result<(), _>`, the
+/// visit runs until it is finished. If the return value is a `Control<B>`, it
+/// can be used to change the control flow of the search. `Control::Break` will
+/// stop the visit early, returning the contained value from the function.
```
jmcomets: I wrote a test [here](https://github.com/jmcomets/petgraph/commit/5b24f31072a305023a7b17497250180d4fac0d87) to check this statement and I think you meant to write the opposite (eg. `Result<(), B>` works similarly than `Control<B>`).

ratmice: Indeed you are right, I tried split it up into separate paragraphs about Control, and the implementations of ControlFlow.

I had a bit of trouble understanding the new changes to Control/ControlFlow with should_prune/should_break, perhaps I should just have said something value like:
"The implementation for Result provides more control over exiting the search early."
rather than stating anything specific?

jmcomets: IIRC the reasoning behind that impl is for cases where there are fallible traversal functions, to wrap the "pure" control flow within a `Result<C, E>` where `C: ControlFlow`. The reason why it works for `Result<(), E>` is because `(): ControlFlow`. Something along those lines (maybe with an example?) should clear things up.

ratmice: Alright I added another example there, and tried to clear it up the result case.

XVilka:
Thank you!

Issue #107, opened 2016-10-21
Implement the Lengauer-Tarjan dominance algorithm


fitzgen:
...and use `quickcheck` to get differential testing between LT and `algo::dominators::simple_fast`.

PR #297, opened 2019-11-20
Implement the Lengauer-Tarjan dominance algorithm

crgl:
See #107 .

This passes quickcheck tests against the current implementation of `simple_fast`, but might need some renaming/refactoring. The implementation uses the two recommendations of ["Finding Dominators in Practice"](http://jgaa.info/accepted/2006/GeorgiadisTarjanWerneck2006.10.1.pdf): replacing a `Vec<Vec<usize>>` for the buckets of each semi-dominator with a single `Vec<usize>` of size |V|, and skipping calls to `eval` for nodes semi-dominated by the root.

Somewhat foolishly, I had not actually checked the open pull requests and I now realize there is already an open pull request (#182). I'm happy to merge my work there, bring the documentation more in line, or anything that's appropriate. Just checking in to see if anyone has time to review, because I do think it would be nice to have the option for larger graphs!

r? @bluss


XVilka:
@fitzgen since you wrote the previous implementation, could you please take a brief look at this one, and if everything is fine, probably close your https://github.com/petgraph/petgraph/pull/182?

PR #301, opened 2019-12-12
Implement IntoEdges and IntoEdgesDirected for Reversed

### LINKED ISSUE #292: IntoEdges/IntoEdgesDirected for Reversed?
Hey --

I'm trying to write some generic edge-based graph algorithms, and it looks like I've run into a snag with the `Reversed` adapter for graphs. The adapter doesn't implement `IntoEdges` and `IntoEdgesDirected`, which is somewhat inconvenient :)

I tried giving an implementation a shot:

```
impl<G> IntoEdges for Reversed<G>
    where G: IntoEdgesDirected
{
    type Edges = ReversedEdges<G::EdgesDirected>;
    fn edges(self, a: Self::NodeId) -> Self::Edges {
        ReversedEdges {
            iter: self.0.edges_directed(a, Incoming),
        }
    }
}

impl<G> IntoEdgesDirected for Reversed<G>
    where G: IntoEdgesDirected
{
    type EdgesDirected = ReversedEdges<G::EdgesDirected>;
    fn edges_directed(self, a: Self::NodeId, dir: Direction) -> Self::Edges {
        ReversedEdges {
            iter: self.0.edges_directed(a, dir.opposite()),
        }
    }
}

/// A reversed edges iterator.
pub struct ReversedEdges<I> {
    iter: I,
}

impl<I> Iterator for ReversedEdges<I>
    where I: Iterator, I::Item: EdgeRef
{
    type Item = ReversedEdgeReference<I::Item>;
        self.iter.next().map(ReversedEdgeReference)
    }
}
```

This appears to work for directed graphs but not for undirected ones. Here's why:

For a directed graph `g`:

```
0 --> 1 --> 2
```

consider `g.edges(1)`. `self.0.edges_directed(1, Incoming)` would return an edge with `source = 0, target = 1`. Reversing it with `ReversedEdgeReference` is the right thing to do.

For an *undirected* graph `g`:

```
0 -- 1 -- 2
```

consider `g.edges(1)`. `self.0.edges_directed(1, Incoming)` would return edges with `source = 1, target = 0`, as well as `source = 1, target = 2`. This means that reversing the edges with `ReversedEdgeReference` is the *wrong* thing to do.

Basically, this seems to be reflecting the idea that for an undirected graph, `Reversed` shouldn't do any transformations. But I don't see any way to do this in a generic fashion. Any ideas? :)

sunshowers:
This is broken up into two commits -- each one has a detailed message and should compile separately.

Closes #292.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -816,33 +816,19 @@ impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
     ///
     /// - `Directed`, `Outgoing`: All edges from `a`.
     /// - `Directed`, `Incoming`: All edges to `a`.
-    /// - `Undirected`: All edges connected to `a`.
+    /// - `Undirected`, `Outgoing`: All edges connected to `a`, with `a` being the source of each
+    ///   edge.
+    /// - `Undirected`, `Incoming`: All edges connected to `a`, with `a` being the target of each
+    ///   edge.
     ///
     /// Produces an empty iterator if the node `a` doesn't exist.<br>
     /// Iterator element type is `EdgeReference<E, Ix>`.
     pub fn edges_directed(&self, a: NodeIndex<Ix>, dir: Direction) -> Edges<E, Ty, Ix>
     {
-        let mut iter = self.edges_undirected(a);
-        if self.is_directed() {
-            iter.direction = Some(dir);
-        }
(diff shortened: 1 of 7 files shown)


sunshowers:
cc @bluss and @XVilka for whenever you get the chance to look at it. Thank you :D

XVilka:
Thank you!

PR #303, opened 2019-12-19
Fix elements emitted by min_spanning_tree on StableGraph

### LINKED ISSUE #288: StableGraph::from_elements(min_spanning_tree(&graph)) asserts that node does not exist in graph
### Test Case

Here's the JSON-serialized representation of the `Undirected` `StableGraph` I'm using: https://gist.github.com/bonsairobo/1897e608f031cf076ca09df12f5665c3

### Steps To Reproduce

Load the graph and call `StableGraph::from_elements(min_spanning_tree(&graph))`.

### Actual Results

Here's the stack trace:

```
thread '<unnamed>' panicked at 'StableGraph::add_edge: node index 978 is not a node in the graph', .cargo/registry/src/github.com-1ecc6299db9ec823/petgraph-0.4.13/src/graph_impl/stable_graph/mod.rs:400:17
stack backtrace:
   0: std::panicking::default_hook::{{closure}}
   1: std::panicking::default_hook
   2: <std::panicking::begin_panic::PanicPayload<A> as core::panic::BoxMeUp>::get
   3: std::panicking::continue_panic_fmt
   4: std::thread::local::fast::Key<T>::try_initialize
   5: petgraph::graph_impl::stable_graph::StableGraph<N,E,Ty,Ix>::add_edge
   6: <petgraph::graph_impl::stable_graph::StableGraph<N,E,Ty,Ix> as petgraph::data::Build>::add_edge
             at .cargo/registry/src/github.com-1ecc6299db9ec823/petgraph-0.4.13/src/data.rs:166
   7: petgraph::data::from_elements_indexable
             at .cargo/registry/src/github.com-1ecc6299db9ec823/petgraph-0.4.13/src/data.rs:295
   8: <petgraph::graph_impl::stable_graph::StableGraph<N,E,Ty,Ix> as petgraph::data::FromElements>::from_elements
             at .cargo/registry/src/github.com-1ecc6299db9ec823/petgraph-0.4.13/src/data.rs:323
```

### Expected Results

A valid graph is created from the MST.

carrutstick:
Fixes #288 

Currently, when graphs are constructed from `Element` `Iterator`s, it is assumed that the `source` and `target` members of the `Edge` `Element`s refer to the order in which the `Node` `Element`s were emitted by the iterator. That is, if we see an edge with `source=0`, then we add an edge to the graph originating at the first `Node` we got from the current iterator.

This causes a problem when generating `Iterator`s from `StableGraph`s which have had nodes deleted. We have to remember that the index of a particular node as represented by the graph does not necessarily correspond to the order in which the node will be emitted by the iterator.

For instance, if we start with a `StableGraph` described by:
```
Nodes: [0, 1, 2]
Edges: [(0, 1), (1, 2)]
```
and then delete node 0, we end up with:
```
Nodes: [1, 2]
Edges: [(1, 2)]
```
Before this pull request, calling `min_spanning_tree` on this graph would result in the following stream of `Element`s:
```
[Node, Node, Edge { 1, 2 }]
```
Attempting to reconstruct a graph from this iterator would throw an error when trying to add the edge, as it looks for the third (index 2) node added to the graph and is unable to find it.

The solution in this patch is to maintain a mapping of `NodeId` to emitted node index in the `MinSpanningTree` iterator, and to emit `Edge` `Element`s which correctly reference the order in which the node was emitted by the iterator.

Note that it will still be very easy for contributors to make the same mistake as was made in `min_spanning_tree`, by creating `Element` iterators that are unaware of the possibility of missing node indexes from `StableGraph`s.

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -6,7 +6,8 @@
 
 pub mod dominators;
 
-use std::collections::BinaryHeap;
+use core::hash::Hash;
+use std::collections::{BinaryHeap, HashMap};
 use std::cmp::min;
 
 use crate::prelude::*;
@@ -558,7 +559,8 @@ pub fn condensation<N, E, Ty, Ix>(g: Graph<N, E, Ty, Ix>, make_acyclic: bool) ->
 ///
 /// Use `from_elements` to create a graph from the resulting iterator.
 pub fn min_spanning_tree<G>(g: G) -> MinSpanningTree<G>
-    where G::NodeWeight: Clone,
+    where G::NodeId: Eq + Hash,
+          G::NodeWeight: Clone,
           G::EdgeWeight: Clone + PartialOrd,
           G: IntoNodeReferences + IntoEdgeReferences + NodeIndexable,
 {
@@ -578,6 +580,8 @@ pub fn min_spanning_tree<G>(g: G) -> MinSpanningTree<G>
         node_ids: Some(g.node_references()),
         subgraphs: subgraphs,
         sort_edges: sort_edges,
+        node_map: HashMap::new(),
+        node_count: 0,
     }
 
 }
@@ -590,11 +594,14 @@ pub struct MinSpanningTree<G>
     node_ids: Option<G::NodeReferences>,
     subgraphs: UnionFind<usize>,
     sort_edges: BinaryHeap<MinScored<G::EdgeWeight, (G::NodeId, G::NodeId)>>,
+    node_map: HashMap<G::NodeId, usize>,
+    node_count: usize,
 }
 
 
 impl<G> Iterator for MinSpanningTree<G>
     where G: IntoNodeReferences + NodeIndexable,
+          G::NodeId: Eq + Hash,
           G::NodeWeight: Clone,
           G::EdgeWeight: PartialOrd,
 {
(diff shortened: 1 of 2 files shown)


bluss:
Thanks that's great. Do you know if it can be made non breaking by using node indexable - node IDs as integers.

carrutstick:
Ah, good point; wasn't really familiar with that trait.

src/algo/mod.rs
```diff
@@ -618,12 +625,17 @@ impl<G> Iterator for MinSpanningTree<G>
...
         while let Some(MinScored(score, (a, b))) = self.sort_edges.pop() {
-            let g = self.graph;
             // check if the edge would connect two disjoint parts
-            if self.subgraphs.union(g.to_index(a), g.to_index(b)) {
+            let (a_index, b_index) = (g.to_index(a), g.to_index(b));
+            if self.subgraphs.union(a_index, b_index) {
+                let (&a_order, &b_order) = match (self.node_map.get(&a_index),
+                                                  self.node_map.get(&b_index)) {
+                    (Some(a_id), Some(b_id)) => (a_id, b_id),
+                    _ => panic!("Edge references unknown node"),
```
XVilka: I am not sure if `panic!` here is a good choice to be honest.

carrutstick: What do you think would be a better option? That branch should only run if the graph or the iterator itself is in an inconsistent state, so it doesn't seem to be worth changing the return type to `Result`. I guess the other option would be to just `unwrap` the node indexes?

carrutstick: Certainly open to handling it differently

XVilka:
Nevermind, it's OK.

PR #304, opened 2019-12-19
Fix bug in order of iterative DFS

crgl:
Currently, the preorder traversal of depth-first search is incorrect. Nodes are marked visited when they are pushed onto the stack instead of when they are popped off. This guarantees O(n) memory usage where n is number of nodes, but isn't true DFS and thus prevents its use in certain algorithms including the Lengauer-Tarjan algorithm for finding immediate dominators. With this fix, memory usage goes up to O(m) where m is the number of edges, but the resulting preorder is correct.

### DIFF
--- a/src/visit/traversal.rs
+++ b/src/visit/traversal.rs
@@ -86,7 +86,6 @@ impl<N, VM> Dfs<N, VM>
     /// the dfs from a particular node.
     pub fn move_to(&mut self, start: N)
     {
-        self.discovered.visit(start);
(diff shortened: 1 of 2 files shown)


XVilka:
Can you please add the test for this?

crgl:
Strictly speaking this test will not catch all possible order bugs, but since most implementations should either go through children in order or reverse order this should be a reasonable test. At least, it identifies the problem with the original implementation.

PR #307, opened 2019-12-23
Check also the formatting in CI

XVilka:


XVilka:
Seems there is a difference between stable Rust fmt and the nightly one.

PR #311, opened 2019-12-28
Add configurable functions for adding attributes to dotfile features

timthelion:
This allows you to easily style dotfile output like so:

```
Dot::with_attr_getters(
       self.state_network,
        &[],
        &|_, _| "arrowhead = vee".to_string(),
        &|_, nr| format!("shape = box color = {}", if nr.id() == self.state { "red" } else { "black" }).to_string(),
);
```

index

### DIFF
--- a/src/dot.rs
+++ b/src/dot.rs
@@ -2,7 +2,7 @@
 
 use std::fmt::{self, Display, Write};
 
-use crate::visit::GraphRef;
+use crate::visit::{GraphBase, GraphRef, Data, GraphProp, NodeRef, EdgeRef, IntoEdgeReferences, IntoNodeReferences, NodeIndexable};
 
 /// `Dot` implements output to graphviz .dot format for a graph.
 ///
@@ -45,9 +45,14 @@ use crate::visit::GraphRef;
 ///
 /// // If you need multiple config options, just list them all in the slice.
 /// ```
-pub struct Dot<'a, G> {
+pub struct Dot<'a, G>
+where
+    G: IntoEdgeReferences + IntoNodeReferences,
+{
     graph: G,
     config: &'a [Config],
+    get_edge_attributes: &'a dyn Fn(G, G::EdgeRef) -> String,
+    get_node_attributes: &'a dyn Fn(G, G::NodeRef) -> String,
 }
 
 static TYPE: [&str; 2] = ["graph", "digraph"];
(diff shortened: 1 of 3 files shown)


sunshowers:
Thanks! I've really wanted to do something like this.

What do you think of returning the attribute name and value as separate return values? So:

```
Dot::with_attr_getters(
       self.state_network,
        &[],
        &|_, _| ("arrowhead", "vee".to_string()),
        &|_, _| ("shape", "box".to_string()),
        &|_, nr| ("color", if nr.id() == self.state { "red" } else { "black" }).to_string()),
);
```

timthelion:
attribute name + value pairs was the first thing I considered. But when I looked at the [spec](https://graphviz.gitlab.io/_pages/doc/info/attrs.html) I learned that graphviz has an extremely rich type system, and we would have to support all of those types. This way, all types are supported, and third party graphviz serialization code can be bolted on without trouble. This way we aren't getting things wrong, just incomplete.

You are right, however, that passing the function a [Write](https://doc.rust-lang.org/1.5.0/std/io/trait.Write.html) would be slightly more efficient due to reduced heap allocations (not that this code is performance critical). I can change the code to to that if desired.

sunshowers:
> But when I looked at the spec I learned that graphviz has an extremely rich type system, and we would have to support all of those types

Ahh, right, that makes sense!

Another thing to consider is if the label should be configurable as well instead of just being the `Debug` or `Display` impl. Feels like it should be but I could be wrong.

timthelion:
I agree that it would be better if the label could be programmatically changed, this PR can accommodate that by adding a new Config option which leaves out the label attr. Then label would just be another configurable attribute.

XVilka:
@timthelion could you please update your PR according to this label suggestion?

timthelion:
@XVilka I just added a new option, `NodeNoLabel` https://github.com/petgraph/petgraph/pull/311/files#diff-4a4e14fd8e77f7876abced2cdff02387R98 which is analogous to the existing `EdgeNoLabel` and allows for programmatic reconfiguration of node labels. I also added a number of tests.

XVilka:
Looks good, thank you! I will merge it later if there will be no objections.

timthelion:
@sunshowers is this good now?

XVilka:
Thank you! No objections and it looks good to me.

PR #313, opened 2020-01-01
perf: Allow the state in tarjan_scc to be reused

Marwes:
This lets the vectors allocated for `tarjan_scc` be reused as well as lets the caller opt into a closure + iterator based output, avoiding the need to allocate and return `Vec`s.


XVilka:
Thank you! It looks good and indeed helps to write a better code.

PR #316, opened 2020-01-02
Fix typo in example of shorthand types

DustinByfuglien:


XVilka:
Thank you! Good catch.

PR #317, opened 2020-01-03
A couple minor post-0.5.0 updates

sunshowers:
* Added a breaking changes section.
* Exported a type that needs to be.

The latter might justify a 0.5.1 release -- what do you think?


sunshowers:
cc @XVilka, thanks again! :D

XVilka:
Thank you!

PR #318, opened 2020-01-13
Add UndirectedAdaptor

GeorgeScrivener:
Creates an Undirected adaptor that erases the direction from a graph.

I believe I've implemented the required traits, though I'm not sure whether I should have implemented `IntoNeighborsDirected` and `IntoEdgesDirected`.  Let me know if I've missed any

### DIFF
--- a/src/visit/mod.rs
+++ b/src/visit/mod.rs
@@ -753,3 +753,4 @@ where
 
 mod filter;
 mod reversed;
+mod undirected_adaptor;
--- a/src/visit/undirected_adaptor.rs
+++ b/src/visit/undirected_adaptor.rs
@@ -1,0 +1,111 @@
+use crate::visit::{
+    Data, GraphBase, GraphProp, GraphRef, IntoEdgeReferences, IntoEdges, IntoEdgesDirected,
+    IntoNeighbors, IntoNeighborsDirected, IntoNodeIdentifiers, IntoNodeReferences,
+    NodeCompactIndexable, NodeCount, NodeIndexable, Visitable,
+};
+use crate::Direction;
+
+/// An edge direction removing graph adaptor.
+#[derive(Copy, Clone, Debug)]
+pub struct UndirectedAdaptor<G>(G);
+
+impl<G: GraphRef> GraphRef for UndirectedAdaptor<G> {}
+
+impl<G> IntoNeighbors for UndirectedAdaptor<G>
+where
+    G: IntoNeighborsDirected,
+{
+    type Neighbors = std::iter::Chain<G::NeighborsDirected, G::NeighborsDirected>;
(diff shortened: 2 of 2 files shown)


sunshowers:
Yes, I think including those two traits is reasonable since they have defined semantics for undirected graphs.

GeorgeScrivener:
@sunshowers thanks for the review, I'll update the PR

sunshowers:
Thank you!

> Good point about the self loops. I'll filter `self` from the `Outgoing` iterator unless there's a better way you can think of?

That seems totally reasonable. If you want to be consistent with the `Edges` implementation you can filter `self` from the `Incoming` iterator instead: https://docs.rs/petgraph/0.5.0/src/petgraph/graph_impl/mod.rs.html#1647

> As in the order should be Outgoing and then Incoming?

Hmm, not quite. If you look at https://docs.rs/petgraph/0.5.0/petgraph/visit/trait.IntoEdgesDirected.html you'll see defined semantics for `Outgoing` and `Incoming` with undirected graphs (they aim to achieve commonality with directed ones FWIW). So what you want to do is to return every outgoing edge as is, and return every incoming edge with the source and target swapped.

XVilka:
@GeorgeScrivener could you please update this PR so we can get it merged?

TheLostLambda:
Hello all!

I'm looking to use `petgraph` soon and will almost certainly need this functionality when fragmenting a graph in my application! I'm essentially looking to take a directed graph where all components are at least weakly connected, then "cut" one of the edges and compute which nodes are present in the two resultant components (though this cutting could also leave a single connected component too depending on the graph).

My approach currently is to pick an edge, delete it, then start a search from both of the nodes it previously connected to build up a list of nodes in either component. To do this, I might need to walk against the direction specified in the directed graph. If there is a better way to just as quickly split a graph into two subgraphs, let me know, but otherwise I'm willing to try to wrap up this PR and get it merged!

starovoid:
@TheLostLambda Hey, I rebased these changes in #695 and I think it will be merged soon.

XVilka:
@GeorgeScrivener thank you! These changes were merged in https://github.com/petgraph/petgraph/commit/4913deba9bbed4116584d8b4acee584234fea8fe

GilShoshan94:
Hi @XVilka, the changes are merged but not exported into the public visibility.
There is a 
```rust
mod undirected_adaptor;
```
That is private, and it's missing the pub reexport like it's done for `filter` and `reversed`.
On [line 67](https://docs.rs/petgraph/latest/src/petgraph/visit/mod.rs.html#67), it should be added the following:
```rust
pub use self::undirected_adaptor::*;
```

starovoid:
@XVilka, @GilShoshan94 
Fix it in #717

PR #323, opened 2020-01-22
Fixing spelling.

insideoutclub:


jmcomets:
Nice catch! I don't see any issue merging this as-is.

PR #326, opened 2020-02-04
Dijkstra optimization

bwalter:
Adding a node to the `visit_next` heap should not be needed when the calculated distance is not shorter than the existing one.

This change improves the performance of the Dijkstra algorithm by avoiding a heap insertion in such cases. According to the (new) benchmark:
- from 3,338,159 ns/iter (+/- 210,817)
- to 2,312,579 ns/iter (+/- 93,191)

(please check the above assumption as I'm not a graph specialist ;)).

### DIFF
--- a/benches/dijkstra.rs
+++ b/benches/dijkstra.rs
@@ -1,0 +1,33 @@
+#![feature(test)]
+
+extern crate petgraph;
+extern crate test;
+
+use petgraph::prelude::*;
+use std::cmp::{ max, min };
+use test::Bencher;
+
+use petgraph::algo::dijkstra;
+
+#[bench]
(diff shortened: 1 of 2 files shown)


jmcomets:
This looks good to me.

Since the shortest path to the goal necessarily contains the shortest path to each successive node in the final path, it makes perfect sense to only consider *shorter* paths to each node.

Moreover, even if you're adding less nodes to the heap all you're doing is skipping future `pop()`...`continue` sequences because the nodes are already being visited once at most (see the `is_visited(&node)` check.

I don't see an issue merging this as-is.

XVilka:
Thank you!

PR #328, opened 2020-02-07
Add k-shortest-path algorithm

teuron:
Add the k-shortest-path algorithm as a generalization of Dijkstra according to the [Wikipedia Pseudo-Code](https://en.wikipedia.org/wiki/K_shortest_path_routing#Algorithm).

I also added a bench, which tests the same graph as #326.


XVilka:
Thank you!

PR #330, opened 2020-02-14
Allow the GraphMap's hasher to be overriden

### LINKED ISSUE #133: Allow to swap out the Hasher GraphMap uses
The default one is very slow. API should expose a way to swap it out.

jmcomets:
This PR adds a new factory method `with_capacity_and_hasher()`, as well as a
new type parameter. A current limitation is that the hasher should be the same
type for both nodes and edges.

I believe this is a breaking change, since it can break type inference for
`GraphMap::new()` & similar factory methods.

Side note: since this is a big change (though simple), maybe I could use the
opportunity to move the separate trait impls for GraphMap back to the module?

Fixes #133

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -3,6 +3,8 @@
 use crate::graph::IndexType;
 #[cfg(feature = "graphmap")]
 use crate::graphmap::{GraphMap, NodeTrait};
+#[cfg(feature = "graphmap")]
+use std::hash::BuildHasher;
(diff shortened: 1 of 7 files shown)


XVilka:
> since this is a big change (though simple), maybe I could use the
opportunity to move the separate trait impls for GraphMap back to the module?

Why not?

starovoid:
Closed in favor of #622.

PR #341, opened 2020-03-28
Add "serde-1" support to GraphMap

schulzch:
This PR propagates `serde-1` to `indexmap` and implements a custom serializer for `(N, N)` tuples, i.e., edges (`[Source, Target, Weight]` array format). The custom serializer is required for `serde_json` to work (tuples are not allowed as object keys). Also tested using `bincode`.

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -57,10 +57,78 @@ pub type DiGraphMap<N, E> = GraphMap<N, E, Directed>;
 ///
 /// Depends on crate feature `graphmap` (default).
 #[derive(Clone)]
-pub struct GraphMap<N, E, Ty> {
+#[cfg_attr(
+    feature = "serde-1",
+    derive(Serialize, Deserialize),
+    serde(bound(
+        serialize = "N: serde::Serialize, E: serde::Serialize",
+        deserialize = "N: serde::Deserialize<'de>, E: serde::Deserialize<'de>"
+    ))
+)]
+pub struct GraphMap<N, E, Ty> where N: Eq + Hash {
     nodes: IndexMap<N, Vec<(N, CompactDirection)>>,
+    #[cfg_attr(feature = "serde-1", serde(with = "pair_indexmap_serialize"))]
     edges: IndexMap<(N, N), E>,
+    #[cfg_attr(feature = "serde-1", serde(skip))]
     ty: PhantomData<Ty>,
+}
+#[cfg(feature = "serde-1")]
+mod pair_indexmap_serialize {
+    use indexmap::IndexMap;
+    use serde::de::{Deserialize, Deserializer, SeqAccess, Visitor};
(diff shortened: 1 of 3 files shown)


bluss:
Don't you think we need to look at the whole picture and make this serialize in a format identical to or as similar as possible to, what Graph and StableGraph use?

src/graphmap.rs
```diff
@@ -75,6 +143,7 @@ impl<N> NodeTrait for N where N: Copy + Ord + Hash {}
 
 // non-repr(usize) version of Direction
 #[derive(Copy, Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde-1", derive(Serialize, Deserialize))]
 enum CompactDirection {
```
bluss, line 147: This enum is internal so it would be good to not expose it in serialization

src/graphmap.rs
```diff
@@ -57,11 +57,79 @@ pub type DiGraphMap<N, E> = GraphMap<N, E, Directed>;
...
-pub struct GraphMap<N, E, Ty> {
+#[cfg_attr(
+    feature = "serde-1",
+    derive(Serialize, Deserialize),
+    serde(bound(
+        serialize = "N: serde::Serialize, E: serde::Serialize",
+        deserialize = "N: serde::Deserialize<'de>, E: serde::Deserialize<'de>"
+    ))
+)]
+pub struct GraphMap<N, E, Ty> where N: Eq + Hash {
```
bluss, line 68: The implementation details - which fields are used, of graph map are internal, so it would be good to not expose them. Otherwise make clear in documentation, that serialization of GraphMap in particular, is not stable across semver incompatible versions of the library. Thanks. See other comments, a more deliberate serialization format - not derived - could be appropriate.

schulzch:
> Don't you think we need to look at the whole picture and make this serialize in a format identical to or as similar as possible to, what Graph and StableGraph use?

You are right - I didn't care much about that up until now. Is there some documation or issue that discusses the format of those? Seems like yet another reason to have a graph trait across all types of containers.

OT: I do not like `Graph`/`StableGraph` as standard at all. Its a fake-good and misleading default choice due to non O(1) complexities---took me a week to replace all instances with GraphMap in my code. A `LinearScan-`, `List-`, or `Slow-` in its name would help ;).

XVilka:
@schulzch do you have any updates regarding the comments?

schulzch:
The project I'm using petgraph for is currently on hold (but not dead), and I haven't had time to work on a generalizable solution, yet.

symphorien:
Maybe serialize should be implemented on `Visitable+Data` or something like that.

XVilka:
Closed in favor of another implementation of the same: https://github.com/petgraph/petgraph/commit/ffb085f865b390d12eef7d6ab72a5a7d84a644c5

PR #343, opened 2020-04-23
 [visit] access original edge reference from a reversed edge 

sunshowers:
This can be useful when one is trying to write code that is always
interested in the original edge reference, whether the graph under
consideration is reversed or not.

### DIFF
--- a/src/visit/reversed.rs
+++ b/src/visit/reversed.rs
@@ -95,6 +95,16 @@ where
 #[derive(Copy, Clone, Debug)]
 pub struct ReversedEdgeReference<R>(R);
 
+impl<R> ReversedEdgeReference<R> {
+    /// Return the original, unreversed edge reference.
+    pub fn as_unreversed(&self) -> &R { &self.0 }
+
+    /// Consume `self` and return the original, unreversed edge reference.
+    pub fn into_unreversed(self) -> R {
+        self.0
+    }
+}
+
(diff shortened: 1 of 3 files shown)


bluss:
Makes sense. It's a close call but I'd prefer a method instead, what do you think? Like `fn inner(self) -> R` or `fn unreversed(self) -> R` or similar. Main motivation is code readbility. Encapsulation is secondary.

sunshowers:
Sure, that was what I was originally going to do but I thought symmetry with `Reversed` might be more valuable.

sunshowers:
(just updated, thanks!)

src/visit/reversed.rs
```diff
@@ -95,6 +95,16 @@ where
 #[derive(Copy, Clone, Debug)]
 pub struct ReversedEdgeReference<R>(R);
 
+impl<R> ReversedEdgeReference<R> {
+    /// Return the original, unreversed edge reference.
+    pub fn as_unreversed(&self) -> &R { &self.0 }
+
+    /// Consume `self` and return the original, unreversed edge reference.
+    pub fn into_unreversed(self) -> R {
```
bluss, line 103: Nice in this case we know R: Copy because it's an edge ref, but no change is needed

XVilka:
Looks good, thank you!

Issue #306, opened 2019-12-23
Fix all clippy warnings


XVilka:
.

XVilka:
Fixed in https://github.com/petgraph/petgraph/commit/d625ea67985e2c87321d67b1de199fb9bd8a430f

PR #346, opened 2020-05-05
Fix some clippy warnings

nw0:
Fix a number of clippy warnings. See #306.

I have not fixed `clippy::type_complexity` or `clippy::float_cmp`.

Note: the last commit applies `cargo fmt`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: rust
 sudo: false
 matrix:
   include:
-    - rust: 1.37.0
+    - rust: 1.40.0
       env:
       - ALL=' '
     - rust: stable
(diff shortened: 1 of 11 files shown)


nw0:
Hmm, it looks like build fails because I use `std::mem::take`. `std::mem::take` was stabilised in 1.40.0, but the build uses 1.37.0.

In this case it looks like the build configuration should be updated.

XVilka:
I can merge it without the 1.40.0-only change. Could you please undo it and send in a separate PR? Not everyone uses `rustup`.

XVilka:
I checked https://repology.org/project/rust/versions and it's seems to safe to move to >=1.40 versions in 0.6.0 release: https://github.com/petgraph/petgraph/milestone/4

PR #348, opened 2020-05-15
Fix some clippy warnings

nw0:
See discussion on #346

### DIFF
--- a/benches/dijkstra.rs
+++ b/benches/dijkstra.rs
@@ -4,7 +4,7 @@ extern crate petgraph;
 extern crate test;
 
 use petgraph::prelude::*;
-use std::cmp::{ max, min };
+use std::cmp::{max, min};
 use test::Bencher;
 
 use petgraph::algo::dijkstra;
@@ -20,7 +20,7 @@ fn dijkstra_bench(bench: &mut Bencher) {
         let j_from = max(0, i as i32 - neighbour_count as i32 / 2) as usize;
         let j_to = min(NODE_COUNT, j_from + neighbour_count);
         for j in j_from..j_to {
-            let n2  = nodes[j];
+            let n2 = nodes[j];
             let distance = (i + 3) % 10;
             g.add_edge(n1, n2, distance);
         }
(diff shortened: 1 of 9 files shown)


XVilka:
```
     Running `rustc --edition=2018 --crate-name petgraph src/lib.rs --color always --crate-type lib --emit=dep-info,metadata,link -C debuginfo=2 --cfg 'feature="default"' --cfg 'feature="graphmap"' --cfg 'feature="matrix_graph"' --cfg 'feature="stable_graph"' -C metadata=01be66fcbbfcb2ed -C extra-filename=-01be66fcbbfcb2ed --out-dir /home/travis/build/petgraph/petgraph/target/debug/deps -C incremental=/home/travis/build/petgraph/petgraph/target/debug/incremental -L dependency=/home/travis/build/petgraph/petgraph/target/debug/deps --extern fixedbitset=/home/travis/build/petgraph/petgraph/target/debug/deps/libfixedbitset-6450057900dc7259.rlib --extern indexmap=/home/travis/build/petgraph/petgraph/target/debug/deps/libindexmap-3cfa4c4ec29adaa9.rlib`
error[E0658]: use of unstable library feature 'mem_take'
   --> src/matrix_graph.rs:364:26
    |
364 |         let old_weight = mem::take(&mut self.node_adjacencies[p]).into().unwrap();
    |                          ^^^^^^^^^
    |
    = note: for more information, see https://github.com/rust-lang/rust/issues/61129
error: aborting due to previous error
For more information about this error, try `rustc --explain E0658`.
error: Could not compile `petgraph`.
```

nw0:
> Hmm, it looks like build fails because I use std::mem::take. std::mem::take was stabilised in 1.40.0, but the build uses 1.37.0.

I can easily change it back: is this what you want?

I do note that supporting 1.37 does conflict with "Fix all clippy warnings" (#306).

XVilka:
@nw0 what I meant is that fix all warnings except those that require 1.40, so the changes would be fine for 1.37 too.

nw0:
Updated! Hope this works

XVilka:
Perfect, thank you!

PR #350, opened 2020-05-17
Fix Travis CI build

XVilka:


XVilka:
Hmm, the failure seems very strange.

Issue #356, opened 2020-07-11
Documentation about space use of petgraph::graph::Graph is unclear


glandium:
The documentation says "The graph uses O(|V| + |E|) space", but I can't find any description of what `V` is.

A quick grep for `O(|V|` shows there are other similar uses of `V`, with no definition that I can find.

Is it `N`?

MiltFra:
`V` is the set of vertices and thus `|V|` is the number of vertices in the graph.

glandium:
What's the difference between vertices and nodes? Everything is talking about nodes.

MiltFra:
The two words are synonyms. I think "vertex" is more common in the purely mathematical areas but that's just a feeling I have. Also, the set of nodes/vertices is very commonly called `V` which is why I believe the library should stick to that convention.

[Wikipedia](https://en.wikipedia.org/wiki/Graph_(discrete_mathematics)) says that "point" is also a possible term.

I do agree, however, that the documentation should be consistent and easy to understand for people who don't know a lot about the subject. Can you try and fix what you think should be changed and submit a PR?

timthelion:
Just FYI, I have a detailed discussion of petgraph's internal data structures and their pros and cons here: https://hobbs.cz/rust-play/petgraph-internals.html

PR #357, opened 2020-07-11
Change occurrences of `V` with `N` in the documentation.

glandium:
The petgraph structs are generic over `N` (node) rather than `V`, so the
use of `V` (which stands for vertices, but that's rarelt spelled out) in
the documentation is confusing.

Also change occurrences of vertex/vertices with node/nodes in rustdoc
comments, for more consistency.

Fixes: #356


starovoid:
An alternative proposal was accepted (#358 and #814).

PR #358, opened 2020-07-12
Clarification of references to nodes and V as described in #356

mifrandir:
This is an alternative suggestion to #357 and trying to fix the issue #356.

# What changes have been made?

1. All occurrences, except for those that are explicitly alternatives, of "vertex"/"vertices", have been replaced with "node"/"nodes".
2. Most occurrences of `|V|` and `|E|` now include "where V is the set of nodes and E is the set of edges".

# Why do I think this is better than #357?

The letter V is very widely used to refer to the set of nodes/vertices, regardless of which word you prefer. I believe it is more important to stick to this convention to make the documentation more understandable to experienced readers as N is more commonly used to refer to a natural number.

This mismatch between symbols and words happens all the time and is nothing to worry about. (see electric current (I) in physics, for example)

I understand however that readers might be confused as to the meaning of V in particular. For this reason, I have included sufficient explanations within the documentation.

Fixes: #356

### DIFF
--- a/src/algo/dominators.rs
+++ b/src/algo/dominators.rs
@@ -115,7 +115,7 @@ const UNDEFINED: usize = ::std::usize::MAX;
 /// This algorithm is **O(|V|²)**, and therefore has slower theoretical running time
 /// than the Lengauer-Tarjan algorithm (which is **O(|E| log |V|)**. However,
 /// Cooper et al found it to be faster in practice on control flow graphs of up
-/// to ~30,000 vertices.
+/// to ~30,000 nodes.
 ///
 /// [0]: http://www.cs.rice.edu/~keith/EMBED/dom.pdf
 pub fn simple_fast<G>(graph: G, root: G::NodeId) -> Dominators<G::NodeId>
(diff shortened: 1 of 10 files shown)


glandium:
But then, all the types are e.g. `Graph<N, E, ...>`... not `Graph<V, E, ...>`.

MiltFra:
But the `N` in `Graph<N, E, ...>` does not refer to the set of nodes but to the node type, right? So, since they are two different things, they don't need to have the same symbol.

XVilka:
I think this approach is indeed better. Also explicit clarification of what `V` and `N` mean is helpful.

XVilka (approved):

RaoulLuque:
@mifrandir If you would be fine with that, I would take over this PR and rebase it to the current branch and check for new spots where a change might be necessary :) @starovoid I am also tagging you to make sure, you are not already taking care of this?

starovoid:
>  @starovoid I am also tagging you to make sure, you are not already taking care of this?

I plan to dive into this topic soon. We have two competing PRs (this and #357), and I plan to prefer the notation that @indietyp  used in his branches `next` and `bm/...`.

starovoid:
The changes were adopted in #814. Thanks!

Issue #338, opened 2020-03-09
A* with cache (multiple start w/ same finish)


IcanDivideBy0:
### Summary

Consider adding an A* implementation that can be reused multiple times for the same finish vertice.

### Motivation

I'm currently building a simple navigation system for a game. Let's say there's multiple monsters targeting the same hero, they all have to find the shortest path to their target. Having a way to pre-populate the A* algorithm with a previous iteration would drastically reduce the cost of this computation (depending of the number of monsters of course).

### Details

* What code needs to change? New traits? Which modules?
`petgraph::algo::astar`, not sure of the direction that it might take, but a I'm thinking of adding someting like this:
```rust
struct AstarInstance<'a> {
    graph, // &'a
    is_goal,
    edge_cost,
    estimate_cost,

    // cached structs from actual astar fn
    visited,
    scores,
}

impl AstarInstance {
  fn path_from(a: NodeId) -> Option<...> { ... }
}
```

* Is this a new graph algorithm? Include links to papers, Wikipedia, or other
No

* Are you willing to implement this yourself? Mentor someone else and help them
Yes!

MiltFra:
This sounds like a pretty cool idea. I've discovered this project today and I'm willing to work on it. 
Have you started implementing this yet? Do you have another project that already implements this?

IcanDivideBy0:
@MiltFra I didn't had time to start on this, and I don't think i'll start until any maintainer gives an opinion about it. If you want to give it a shot, I'd be happy to see the outcome :)

MiltFra:
@IcanDivideBy0 Let me know what you think about my implementation.

I suggested some more things in the PR that could be done. If that sounds useful to you, let me know.

PayasR:
This issue makes an excellent case for implementing algorithms as structs, not functions. Caching greatly improves performance, and there's enough precedence for it- [Boost graph](https://github.com/boostorg/graph/blob/develop/include/boost/graph/breadth_first_search.hpp) does it well, and Deitmar Kuhl's [paper](http://www.dietmar-kuehl.de/generic-graph-algorithms.pdf) is an excellent read on the subject. Maintainers, would you agree?

PR #359, opened 2020-07-12
A* instance implementation as requested in #338

MiltFra:
This is a simple extension of the already existing A*-implementation.

I've tried to keep it simple, more optimisation and caching could be done for certain scenarios but that would come with tradeoffs (e.g. caching results for calls to `run`, using actual shortest path lengths from previous calls as "estimates", ...); I'd be happy to implement them though, just let me know.

The time complexity does not change, at all (unless `estimate_cost` has a bigger complexity than `astar`). Optimisation is achieved by reducing calls to `estimate_cost`. If `estimate_cost` is very inexpensive (e.g. `|_| 0`) this might even result in a significant performance penalty.

The space complexity has technically not changed either.

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -25,6 +25,7 @@ use crate::visit::Walker;
 use crate::visit::{Data, IntoNodeReferences, NodeRef};
 
 pub use super::astar::astar;
+pub use super::astar::AstarInstance;
 pub use super::dijkstra::dijkstra;
 pub use super::isomorphism::{is_isomorphic, is_isomorphic_matching};
 pub use super::simple_paths::all_simple_paths;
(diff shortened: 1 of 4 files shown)


starovoid:
Hi, I like the proposed feature. What is holding this PR back from merging?

mifrandir:
Nothing to my knowledge. Although I'm not currently in a position to make any further changes.

XVilka:
> Hi, I like the proposed feature. What is holding this PR back from merging?

Conflicts should be fixed.

mifrandir:
> > Hi, I like the proposed feature. What is holding this PR back from merging?
> 
> Conflicts should be fixed.

Should be fine now.

Issue #361, opened 2020-07-24
is_cyclic_directed does not work for undirected graphs


rbeard0330:
### Test Case

```
#[cfg(test)]
mod tests {
    use petgraph::algo::{is_cyclic_directed, is_cyclic_undirected};
    use petgraph::prelude::*;
    use petgraph::stable_graph::StableGraph;
    use petgraph::visit::{depth_first_search, Control, DfsEvent, IntoNodeIdentifiers};

    fn get_tree_graph() -> StableGraph<f32, f32, Undirected, u32> {
        StableGraph::from_edges(vec![(0, 1), (0, 2), (1, 3), (1, 4)])
    }

    #[test]
    fn test_is_cyclic_directed() {
        let graph = get_tree_graph();
        assert_eq!(is_cyclic_directed(&graph), is_cyclic_undirected(&graph));
    }

    #[test]
    fn log_dfs() {
        let graph = get_tree_graph();
        depth_first_search(&graph, graph.node_identifiers(), |event| {
            dbg!(event);
            Control::<usize>::Continue
        });
    }
}
```

### Steps To Reproduce

See above:

* First, create an undirected and acyclic graph
* Then, apply is_cyclic_directed()

### Actual Results

Always returns true.  is_cyclic_directed is essentially a wrapper around depth_first_search, which reports a back edge each time it encounters a new node in an undirected graph, because it does not distinguish the edge that was traversed to reach the new node from undiscovered edges.  This behaviour is probably incorrect in general, but certainly incorrect for cycle detection.

### Expected Results

is_cyclic_directed should just not be callable on undirected graphs.  PR incoming.

PR #362, opened 2020-07-24
Pull request to fix Issue #361

rbeard0330:
More detail here:

https://github.com/petgraph/petgraph/issues/361

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -180,7 +180,7 @@ where
 /// needed.
 pub fn is_cyclic_directed<G>(g: G) -> bool
 where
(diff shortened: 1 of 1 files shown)


XVilka:
Could you please add this test as well?

rbeard0330:
> Could you please add this test as well?

I don't think there are any tests that make sense to add.  The test I put in the issue thread to show the problem on undirected graphs won't compile under this PR.  There is an existing unit test for is_cyclic_directed() on directed graphs.

Let me know if I'm missing something.  I'm happy to write whatever tests would be helpful.

PR #366, opened 2020-08-12
Implement Clone for iterators, wherever possible

mullr:


mullr:
Is there anything I can do to help usher this through? I'd very much like to be using the official version of PetGraph, rather than a fork.

PR #238, opened 2019-02-01
No_std feature switch 

jonimake:
Using this feature switch will also turn on alloc feature switch which
is used to determine if the collections used in the library should be
imported from the alloc library or not.

Most notable change in no_std is that instead of a HashMap and Set
a BTreeMap and Set are used.

**Benchmarks**

***Standard***
```
running 6 tests
test bench_praust_mst          ... bench:         446 ns/iter (+/- 19)
test full_iso_bench            ... bench:       1,758 ns/iter (+/- 75)
test petersen_iso_bench        ... bench:       1,332 ns/iter (+/- 103)
test petersen_undir_iso_bench  ... bench:         977 ns/iter (+/- 42)
test praust_dir_no_iso_bench   ... bench:   1,115,236 ns/iter (+/- 89,176)
test praust_undir_no_iso_bench ... bench:   1,116,010 ns/iter (+/- 90,303)

test result: ok. 0 passed; 0 failed; 0 ignored; 6 measured; 0 filtered out

     Running target/release/deps/ograph-0a0ed0f5ec42a012

running 3 tests
test bench_add_edge ... bench:         489 ns/iter (+/- 13)
test bench_inser    ... bench:           4 ns/iter (+/- 1)
test bench_remove   ... bench:         221 ns/iter (+/- 13)

test result: ok. 0 passed; 0 failed; 0 ignored; 3 measured; 0 filtered out

     Running target/release/deps/stable_graph-5231646bce1ffe18

running 12 tests
test full_edges_default        ... bench:          10 ns/iter (+/- 0)
test full_edges_in             ... bench:          11 ns/iter (+/- 0)
test full_edges_out            ... bench:          10 ns/iter (+/- 0)
test graph_map                 ... bench:         272 ns/iter (+/- 13)
test neighbors_default         ... bench:           6 ns/iter (+/- 0)
test neighbors_in              ... bench:          10 ns/iter (+/- 0)
test neighbors_out             ... bench:           8 ns/iter (+/- 1)
test sccs_graph                ... bench:       3,119 ns/iter (+/- 246)
test sccs_stable_graph         ... bench:       3,221 ns/iter (+/- 230)
test stable_graph_map          ... bench:         389 ns/iter (+/- 36)
test stable_graph_retain_edges ... bench:         415 ns/iter (+/- 17)
test stable_graph_retain_nodes ... bench:          63 ns/iter (+/- 3)

test result: ok. 0 passed; 0 failed; 0 ignored; 12 measured; 0 filtered out
```

[...]

PR #370, opened 2020-08-26
NO_STD

zendurix:
I have succesfully implemented no_std with hasbrown crate (to enable no_std use with default-features = false
It passed no_std check (https://github.com/mystor/cargo-no-std-check)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,7 +103,14 @@
 //! * **matrix_graph** -
 //!   Defaults on. Enables [`MatrixGraph`](./matrix_graph/struct.MatrixGraph.html).
 //!
+#![cfg_attr(not(feature = "std"), no_std)]
 #![doc(html_root_url = "https://docs.rs/petgraph/0.4/")]
+
+#[cfg(not(feature = "std"))]
+extern crate core as std;
+
+#[cfg(not(feature = "std"))]
+extern crate alloc;
 
 extern crate fixedbitset;
 #[cfg(feature = "graphmap")]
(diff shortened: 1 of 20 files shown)


src/lib.rs
```diff
@@ -173,6 +180,60 @@ pub mod graph {
...
+    #[cfg(feature = "std")]
+    pub use std::collections::{hash_map::Entry::{Occupied, Vacant}, HashSet, hash_map::Iter};
+    #[cfg(not(feature = "std"))]
+    pub use hashbrown::{ HashSet, hash_map::{ HashMap, Entry::{Occupied, Vacant}, Iter}};
+
+    #[cfg(feature = "std")]
+    pub use std::collections::{BinaryHeap, HashMap,VecDeque};
+    #[cfg(not(feature = "std"))]
+    pub use alloc::collections::{BinaryHeap, VecDeque};
+
```
XVilka, line 228: Please remove unnecessary empty lines.

XVilka (changes requested):
I am not sure `use crate::lib::*;` is a good practice. The idea to wrap everything is good, but probably better to import only used modules. Please also add the corresponding CI job from https://github.com/petgraph/petgraph/pull/238

XVilka:
Also CI is failing:
```
error[E0432]: unresolved imports `rand::thread_rng`, `rand::ChaChaRng`
 --> tests/unionfind.rs:5:12
  |
5 | use rand::{thread_rng, ChaChaRng, Rng, SeedableRng};
  |            ^^^^^^^^^^  ^^^^^^^^^ no `ChaChaRng` in the root
  |            |
  |            no `thread_rng` in the root
warning: unused import: `SeedableRng`
 --> tests/unionfind.rs:5:40
  |
5 | use rand::{thread_rng, ChaChaRng, Rng, SeedableRng};
  |                                        ^^^^^^^^^^^
  |
  = note: #[warn(unused_imports)] on by default
warning: unused import: `Rng`
 --> tests/unionfind.rs:5:35
  |
5 | use rand::{thread_rng, ChaChaRng, Rng, SeedableRng};
  |                                   ^^^
error: aborting due to previous error
For more information about this error, try `rustc --explain E0432`.
error: Could not compile `petgraph`.
```

zendurix:
Should all test be included in no_std? this will require some additional work

XVilka:
Yes.

zendurix:
Maybe leave thoose test that uses RNG outside of `no_std`?

XVilka:
Yes, that makes sense.

zendurix:
I have implemented requested changes, all tests are passing in both std and no_std

XVilka:
Looks better, thanks!
Could you please also add a `no_std` CI job?

zendurix:
@XVilka 
I am not sure how to do that, should I just add this line at the end of travis.yml ?
`&&
      cargo test --all --no-default-features --features="graphmap","stable_graph","matrix_graph"`

XVilka:
@zendurix you can just take `.travis.yml` from the old PR: https://github.com/petgraph/petgraph/pull/238/files#diff-354f30a63fb0907d4ad57269548329e3

XVilka:
Also I am not sure, instead of adding `#[cfg(feature="std")]` before every `println!` maybe it's better to define own macro or function and do the switch once?

zendurix:
> Also I am not sure, instead of adding `#[cfg(feature="std")]` before every `println!` maybe it's better to define own macro or function and do the switch once?
> 
> By the way, you can also use https://github.com/hobofan/cargo-nono for checks.

I can think of smth like
``` 
pub fn print(args) {
   #[cfg(feature = "std)]
   println!(args)
}
```

But I am not very good with macros, and I am not sure how to do this

XVilka:
Maybe something like
```rust
macro_rules! no_std_println {
  ($(t:tt)*) => {
    #[cfg(feature = "std")]
    {
      println!($($t)*);
    }
  };
}
```

zendurix:
I see problems with CI (cause old PR have feature no_std) I will impl this macro and fix CI later today

XVilka:
@zendurix have you had any luck with this update? Could you please rebase because we migrated from Travis CI to GitHub Actions?

XVilka:
@zendurix could you please rebase your PR?

src/lib.rs
```diff
@@ -173,6 +180,52 @@ pub mod graph {
...
+    #[cfg(feature = "std")]
+    pub use std::collections::hash_map::RandomState;
+    #[cfg(not(feature = "std"))]
+    pub use hashbrown::hash_map::DefaultHashBuilder as RandomState;
+
+    #[cfg(feature = "std")]
+    pub use std::collections::{hash_map::Entry::{Occupied, Vacant}, HashSet, hash_map::Iter};
+    #[cfg(not(feature = "std"))]
+    pub use hashbrown::{ HashSet, hash_map::{ HashMap, Entry::{Occupied, Vacant}, Iter}};
```
bluss, line 220: HashMap is used in the public API - this PR would mean that public API changes breakingly based on a cargo feature, and that's not an allowed usage of cargo features in general.

bluss (changes requested):

chaosprint:
Hi, this is such a wonderful job. Is there any plan that the rebase can be done in the near future?

indietyp:
0.7.0 will have no-std support, until this version has landed I will keep this PR open.

starovoid:
Hi all, this work can definitely be useful in further refactoring, but since `no_std` support was added in #747, I decided to convert this PR into a draft.

Issue #378, opened 2020-10-13
A* fails to find optimal solution with admissible but inconsistent heuristic


lillanes:
A* using an [admissible but inconsistent](https://en.wikipedia.org/wiki/Consistent_heuristic) heuristic should find optimal paths. The current implementation seems to assume that heuristics will always be consistent, and fails to find optimal paths in some cases.

### Test Case

```rust
use petgraph::prelude::*;
use petgraph::algo::astar;

#[test]
fn test_astar_admissible_inconsistent() {
    let mut g = Graph::new();
    let a = g.add_node("A");
    let b = g.add_node("B");
    let c = g.add_node("C");
    let d = g.add_node("D");
    g.add_edge(a, b, 3);
    g.add_edge(b, c, 3);
    g.add_edge(c, d, 3);
    g.add_edge(a, c, 8);
    g.add_edge(a, d, 10);

    let admissible_inconsistent = |n: NodeIndex| match g[n] {
        "A" => 9,
        "B" => 6,
        "C" => 0,
        &_ => 0,
    };

    let optimal = astar(&g, a, |n| n == d, |e| *e.weight(), admissible_inconsistent);
    assert_eq!(optimal, Some((9, vec![a, b, c, d])));
}
```

### Steps To Reproduce

See above.

### Actual Results

`astar` finds the path `A->D` with cost 10.

### Expected Results

The algorithm should find the optimal path `A->B->C->D` with cost 9.

bluss:
Fixed by #379

PR #379, opened 2020-10-13
Re-visit nodes in A* (fix for #378)

lillanes:
You can see more details in #378 

This PR forces A* to re-visit nodes if they are reached through a shorter path than before. This ensures that the algorithm can find optimal paths when using an inconsistent (but admissible) heuristic.

### DIFF
--- a/src/astar.rs
+++ b/src/astar.rs
@@ -3,7 +3,7 @@ use std::collections::{BinaryHeap, HashMap};
 
 use std::hash::Hash;
 
-use super::visit::{EdgeRef, GraphBase, IntoEdges, VisitMap, Visitable};
+use super::visit::{EdgeRef, GraphBase, IntoEdges};
 use crate::scored::MinScored;
 
 use crate::algo::Measure;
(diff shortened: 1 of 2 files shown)


src/astar.rs
```diff
@@ -71,63 +71,66 @@ pub fn astar<G, F, H, K, IsGoal>(
     mut estimate_cost: H,
 ) -> Option<(K, Vec<G::NodeId>)>
 where
-    G: IntoEdges + Visitable,
+    G: IntoEdges,
```
bluss, line 74:
```suggestion
    G: IntoEdges + Visitable,
```

I would just keep this in the interface for flexibility, even if it's no longer used.

src/astar.rs
```diff
@@ -3,7 +3,7 @@ use std::collections::{BinaryHeap, HashMap};
 
 use std::hash::Hash;
 
-use super::visit::{EdgeRef, GraphBase, IntoEdges, VisitMap, Visitable};
+use super::visit::{EdgeRef, GraphBase, IntoEdges};
```
bluss, line 6:
```suggestion
use super::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
```

bluss:
Thanks a lot for the fix and the testcases and the improvement

bluss:
Ah yeah, use keywords to close issues smoothly https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #380, opened 2020-11-18
Add a few more`fn size_hint` method overrides

regexident:
The default implementation of `fn size_hint` returns `(0, None)` which is correct for any iterator, but might prohibit certain optimizations (such as for pre-allocating a collecting buffer).

While some of the Iterator implementations already provided an implementation of `fn size_hint`, others don't yet, but could.

There are basically two possible straight-forward implementations for wrapping iterators to pick from:

Non-filtering iterators:
```rust
fn size_hint(&self) -> (usize, Option<usize>) {
    self.iter.size_hint()
}
```

Filtering iterators:
```rust
fn size_hint(&self) -> (usize, Option<usize>) {
    let (_, upper) = self.iter.size_hint();
    (0, upper)
}
```

### DIFF
--- a/src/algo/dominators.rs
+++ b/src/algo/dominators.rs
@@ -139,6 +139,10 @@ where
         }
         None
     }
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (_, upper) = self.iter.size_hint();
+        (0, upper)
+    }
 }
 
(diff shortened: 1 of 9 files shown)


ABorgna (approved):
Thanks for this!
Could you please rebase over master to run the new CI checks?

PR #383, opened 2020-12-17
Added an example on how to use the all_simple_paths function

alexanderkjall:
I had a hard time understanding what `TargetColl: FromIterator<G::NodeId>` meant in the function definition, so I thought it would help other beginners if the documentation had an example on how to use this function.


reinvantveer:
👍 Yes it was definitely helpful

XVilka (approved):

PR #386, opened 2020-12-31
Algorithm: Greedy feedback arc set

tobz1000:
Obtains a feedback arc set of a graph which is both decently small and has good time complexity.

This procedure involves the creation of a 64-bit-indexed graph mirroring the input graph, which may somewhat negate the memory advantages of 32-bit for huge graphs (I couldn't find a better way to support but 32-bit- and 64-bit-indexed graphs).

Credit for finding the algorithm used here goes to @steffahn in [this discussion](https://users.rust-lang.org/t/algorithm-technique-for-minimum-required-indirection-in-a-type-structure/53413/6).


ABorgna (commented):
Hi, thanks for this. Can you rebase, so that it runs the tests with the new CI?

The paper uses a vector of doubly linked lists to find node candidates, so it avoids the linear search in your implementation.
It would be nice to implement that, but if you prefer we can merge this PR as-is and optimize it afterwards.

On that vein, we could use a small benchmark to test future changes.

tobz1000:
> Hi, thanks for this. Can you rebase, so that it runs the tests with the new CI?

Rebase was conflicting so I've merged master; I hope that's okay.

> The paper uses a vector of doubly linked lists to find node candidates, so it avoids the linear search in your implementation.
> It would be nice to implement that, but if you prefer we can merge this PR as-is and optimize it afterwards.

Yeah, I would like to try this, good spot. I didn't read the paper thoroughly at the time. I still don't understand how the bin sort makes it "trivial" to find the next candidate node when constructing the node sequence 😅. But I'll have a play and hopefully figure it out.

ABorgna:
They store a cursor into the corresponding linked list for each node so they can delete the nodes in constant time.

I think not even the experimental [cursor API][1] for std's linked lists supports doing that, you'll have to roll your own.
You can pack the nodes directly into a `Vec<LLNode>` indexed by the `SeqGraphIx`s to avoid the cursor indirection.

[1]: https://github.com/rust-lang/rfcs/blob/master/text/2570-linked-list-cursors.md

tobz1000:
The new changeset is a lot more to review than the original, sorry 😁. This is mostly due to the linked list implementation.

I've added benchmarks and tested them against the original - the improvement is mostly good. It's a bit slower than before when there aren't many edges, but the time complexity is clearly better for many edges.

```
Tournament (|E| = O(|V|^2))
            old     new
10 nodes    3us     8us
50 nodes    0.16ms  0.16ms
100 nodes   1.8ms   0.6ms
200 nodes   15 ms   2.1ms
300 nodes   58ms    4.5ms

Fan Graph (|E| = O(|V|)
            old     new
10 nodes    2us     4us
50 nodes    9us     20us
200 nodes   54us    75us
```

src/feedback_arc_set.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Isomorphic to input graph; used in the process of determining a good node sequence from which to
+/// extract a feedback arc set.
+type SeqSourceGraph = StableDiGraph<(), (), SeqGraphIx>;
+type SeqGraphIx = usize;
+
+/// \[Generic\] Finds a [feedback arc set]: a set of edges in the given directed graph, which when
+/// removed, make the graph acyclic.
+///
+/// Uses a [greedy heuristic algorithm] to select a small number of edges in reasonable time, but
+/// does not necessarily find the minimum feedback arc set.
```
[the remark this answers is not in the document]
tobz1000: Changed to `O(|E|)` for the new implementation

src/feedback_arc_set.rs
```diff
@@ -0,0 +1,426 @@
...
+            s_2.push_front(nodes[sink_fas_ix].data().graph_ix);
+        }
+
+        while let Some(source_fas_ix) = buckets.sources.pop(&mut nodes) {
+            some_moved = true;
+            buckets.update_neighbour_node_buckets(source_fas_ix, &mut nodes);
+            s_1.push_back(nodes[source_fas_ix].data().graph_ix);
+        }
+
+        if let Some(list) = buckets
```
tobz1000, line 153: This seems like it could bring the complexity up to potentially `|E|^2`, but in the benchmark tests at least, this isn't the case.

I had a much more complex implementation using a linked list-of-lists so that the highest bucket is more immediately available, but the benchmark results were the same, and it would be harder to maintain.

ABorgna: Yeah, this seems ok.

I just tried adding a trim function at the end of `update_neighbour_node_buckets` that trims the empty lists at the end of `bidirectional_pve_dd` (since deltas increase one by one, the complexity gets amortized) and just got around a 1% decrease in the benchmarks.
Something similar could be done for the negative deltas by keeping track of the first non-empty list index, but it may complicate the implementation too much.

ABorgna (approved):
This is great :)
Seems that time constants are a bit higher, but times are much better for non-tiny inputs.

I just have a nitpick, but it can be merged as-is.

tobz1000:
I've tried trimming the bucket lists (hopefully this is similar to what you tried): https://github.com/tobz1000/petgraph/commit/266f97976a1e8f1d5d08ca99a75cb43b7800d75a

The benchmark changes are within a margin of error to me it seems - but I've observed that it occasionally trims 5 or so empty lists off at once with the test graphs. I suppose for some topologies these trims might be even larger, and it might cause a non-negligible performance improvement. Up to you if you want to merge that change as well.

Thank you for reviewing this and giving some great input :)

ABorgna:
It doesn't seem to add a cost and may be useful for some graphs as you said, so lets add it to this branch and I'll merge the PR.
Thanks again !

PR #395, opened 2021-01-22
Improve StableGraph de-serialization with holes

s3bk:
The current function calls `insert` for every hole, which itself is `O(len - insert pos)`. Needless to say this is quite bad with lots of nodes and holes.
This replaces it with a `O(number of nodes + number of holes)` function that merges into a newly allocated `Vec`.
I wrote a swap based implementation as well, but it is slower.

This also adds a benchmark and makes the serialize-deserialize test more complex.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ rand = "0.5.5"
 odds = { version = "0.2.19" }
 defmac = "0.1"
 itertools = { version = "0.8", default-features = false }
-bincode = "*"
+bincode = "1.3.3"
 
 [features]
 default = ["graphmap", "stable_graph", "matrix_graph"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ rand = "0.5.5"
 odds = { version = "0.2.19" }
 defmac = "0.1"
 itertools = { version = "0.8", default-features = false }
+bincode = "*"
 
 [features]
 default = ["graphmap", "stable_graph", "matrix_graph"]
--- a/benches/serialize.rs
+++ b/benches/serialize.rs
@@ -3,9 +3,9 @@
 extern crate petgraph;
 extern crate test;
 
-use test::Bencher;
 use petgraph::prelude::*;
-use rand::{Rng};
+use rand::Rng;
+use test::Bencher;
 
 const NUM_NODES: usize = 1_000_000;
 const NUM_EDGES: usize = 100_000;
@@ -33,13 +33,10 @@ fn make_stable_graph() -> StableGraph<u32, u32> {
     g
 }
 
-
 #[bench]
 fn serialize_bench(bench: &mut Bencher) {
     let graph = make_stable_graph();
-    bench.iter(|| {
-        bincode::serialize(&graph).unwrap()
-    });
+    bench.iter(|| bincode::serialize(&graph).unwrap());
 }
 
 #[bench]
--- a/benches/serialize.rs
+++ b/benches/serialize.rs
@@ -13,10 +13,12 @@ const NUM_HOLES: usize = 1_000_000;
 
 fn make_stable_graph() -> StableGraph<u32, u32> {
     let mut g = StableGraph::with_capacity(NUM_NODES + NUM_HOLES, NUM_EDGES);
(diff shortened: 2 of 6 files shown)


bluss:
Nice, I definitely prefer this to something that breaks the data format. I'm not sure how breaking changes in serialization formats are usually handled, it's an unfortunate thing to have to do

src/graph_impl/stable_graph/serialization.rs
```diff
@@ -197,30 +196,47 @@ where
...
+            node_pos = hole_pos.index() + 1;
+            debug_assert_eq!(nodes.len(), node_pos);
+        }
+        nodes.extend(compact_nodes);
+
+        /*
+        // smart implementation using swap.
+        // is actually slower in the benchmark.
+        
+        let mut nodes = input.nodes;
```
bluss, line 224: Thanks for showing the code but before integration the commented-out code should go. You could put in in a commit message (?)

src/graph_impl/stable_graph/serialization.rs
```diff
@@ -197,30 +196,47 @@ where
...
         if edges.len() >= <Ix as IndexType>::max().index() {
             Err(invalid_length_err::<Ix, _>("edge", edges.len()))?
         }
+        
+        let mut nodes = Vec::with_capacity(input.nodes.len() + node_holes.len());
+
+        let mut compact_nodes = input.nodes.into_iter();
+        let mut node_pos = 0;
+        for hole_pos in node_holes.iter() {
+            nodes.extend(compact_nodes.by_ref().take(hole_pos.index() - node_pos));
```
bluss, line 210: `hole_pos` is arbitrary untrusted input and deserialization's role is to validate it; here `hole_pos - node_pos` can panic in debug builds and wrap around in release builds - shouldn't it be an error at deserialization if the hole position is invalid?

src/graph_impl/stable_graph/serialization.rs
```diff
@@ -197,30 +196,47 @@ where
...
+            nodes.push(Node {
+                weight: None,
+                next: [EdgeIndex::end(); 2],
+            });
+            node_pos = hole_pos.index() + 1;
+            debug_assert_eq!(nodes.len(), node_pos);
```
bluss, line 216: Debug assertions are nice but here it's perpelexing since it doesn't mesh with the aforementioned mindset that the input is untrusted and arbitrary. We don't panic on invalid input, we return an error.

src/graph_impl/stable_graph/serialization.rs
s3bk: Yep. fixed.

src/graph_impl/stable_graph/serialization.rs
s3bk: This debug assertion is really just to double-check the indices.

Cargo.toml
```diff
@@ -39,6 +39,7 @@ rand = "0.5.5"
 odds = { version = "0.2.19" }
 defmac = "0.1"
 itertools = { version = "0.8", default-features = false }
+bincode = "*"
```
ABorgna, line 42: Set a version for the dependency.

benches/serialize.rs
```diff
@@ -0,0 +1,50 @@
...
+    while g.node_count() > NUM_NODES {
+        let idx = rng.gen_range(0, indices.len());
+        g.remove_node(indices[idx]);
+    }
+
+    g
+}
+
+#[bench]
+fn serialize_bench(bench: &mut Bencher) {
```
ABorgna, line 37: Mention that this uses stable graphs in the name of the benchmark, something like `stable_graph_serialize` (and `stable_graph_deserialize`).

ABorgna (changes requested):
Hi! This looks nice and can be merged before deciding on any format-breaking changes.
I just have a couple of comments, in addition to removing the commented code.

ABorgna:
@s3bk Could you resolve the conflict and address bluss' comment? We can merge this after that.

s3bk:
@ABorgna I think it should be okay now.

ABorgna (approved):

PR #397, opened 2021-01-25
Shortest distances based on BFS

PrototypeRailGun:
The lengths of the shortest paths from the start vertex to all the others.

The algorithm is based on BFS, and the path length is equal to the number of edges in this path.
If it is impossible to get from the vertex x to the vertex y, then the distance from x to y will be equal to f32::INFINITY.
Time complexity: **O(|V| + |E|)**, where **|V|** is the number of vertices, and **|E|** is the number of edges in the graph.


PrototypeRailGun:
I need help to understand why the rustfmt test fails.

PR #402, opened 2021-03-18
Add EdgeIndexable public trait

mtreinish:
Just as there is a NodeIndexable public trait that lets users handle the
interaction with node indices there was no corresponding trait for
edges. This commit adds a new public trait EdgeIndexable which mirrors
NodeIndexable but works with edge indices instead. It is then
implemented for StableGraph and frozen (the graphs where there were
NodeIndexable impls and EdgeIndexable made sense).


ABorgna:
Hi, the failing test in master is fixed now. Could you rebase your changes and fix the conflict? Thanks.

ABorgna:
This works nicely. Thanks !

PR #413, opened 2021-04-16
Modified Tarjan SCC to reduce memory usage.

saolof:
Reduced heap memory usage, improved cache locality, and made the code somewhat cleaner for a possible future rewrite to an imperative implementation. The changes to a more modern version of Tarjan's algorithm were directly based on algorithm 3 in "A Space-Efficient Algorithm for Finding Strongly Connected Components" by David J. Pearce, while also aiming to avoid changing the structure of the code too much.

In addition to saving memory, this version has the property that it builds a lookup
table for the components at the same time as it outputs them. If v belongs to the kth output component, then
` self.nodes[g.to_index(v)] == Some(k) ` . This could be used to save work in
functions that may want to call it such as condensation (where the work needed to make the node map can be completely avoided, while also creating a condensation that is topologically sorted).

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -7,7 +7,6 @@
 pub mod dominators;
 pub mod tred;
 
-use std::cmp::min;
 use std::collections::{BinaryHeap, HashMap};
 
 use crate::prelude::*;
@@ -330,9 +329,7 @@ where
 
 #[derive(Copy, Clone, Debug)]
 struct NodeData {
-    index: Option<usize>,
-    lowlink: usize,
-    on_stack: bool,
+    rootindex: Option<usize>
 }
 
 /// A reusable state for computing the *strongly connected components* using [Tarjan's algorithm][1].
@@ -341,6 +338,7 @@ struct NodeData {
 #[derive(Debug)]
 pub struct TarjanScc<N> {
     index: usize,
+    componentcount: usize,
     nodes: Vec<NodeData>,
     stack: Vec<N>,
 }
@@ -351,11 +349,13 @@ impl<N> Default for TarjanScc<N> {
     }
 }
 
+
 impl<N> TarjanScc<N> {
     /// Creates a new `TarjanScc`
     pub fn new() -> Self {
         TarjanScc {
-            index: 0,
+            index: usize::MAX, // Invariant: index > componentcount at all times.
+            componentcount: 0, // Will hold if index is initialized to number of nodes - 1 or higher.
             nodes: Vec::new(),
             stack: Vec::new(),
         }
@@ -382,9 +382,7 @@ impl<N> TarjanScc<N> {
         self.nodes.resize(
             g.node_bound(),
             NodeData {
-                index: None,
-                lowlink: !0,
-                on_stack: false,
+                rootindex: None,
             },
         );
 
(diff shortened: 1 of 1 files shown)


saolof:
Another quick comment: because of the way it builds up a component lookup table, this version would also enable a sort_scc_stable function (returning a list of SCCs where the nodes appear in the order given by the original graph or some iterator over its nodes), which would be very useful in some situations. 

Basically, since a node will be assigned to its proper SCC when visit returns from the top level loop, that means you can bucket sort nodes by their rindex value in the top level loop instead of when you pop them from the stack.

src/algo/mod.rs
```diff
@@ -475,7 +468,7 @@ where
     let mut sccs = Vec::new();
     {
         let mut tarjan_scc = TarjanScc::new();
-        tarjan_scc.run(g, |scc| sccs.push(scc.iter().cloned().rev().collect()));
+        tarjan_scc.run(g, |scc| sccs.push(scc.iter().cloned().collect()));
```
ABorgna, line 471: You could cite the algorithm's paper somewhere.

src/algo/mod.rs
```diff
@@ -330,9 +329,7 @@ where
 
 #[derive(Copy, Clone, Debug)]
 struct NodeData {
-    index: Option<usize>,
-    lowlink: usize,
-    on_stack: bool,
+    rootindex: Option<usize>,
```
ABorgna, line 332: Since you are optimizing for memory here, maybe use `Option<NonZeroUsize>`? Index will always be non-zero due to the invariant.

src/algo/mod.rs
```diff
@@ -355,7 +353,8 @@ impl<N> TarjanScc<N> {
     /// Creates a new `TarjanScc`
     pub fn new() -> Self {
         TarjanScc {
-            index: 0,
+            index: std::usize::MAX, // Invariant: index > componentcount at all times.
```
ABorgna, line 356: Why does index count downwards instead of upwards from 1, as in the paper's pseudocode?

ABorgna (commented):
Really nice !

In addition to the comments, it would be nice to have some benchmarks to compare the changes.

src/algo/mod.rs
saolof: Because that lets you have componentcount counting upwards, which makes it possible to do `self.nodes[g.to_index(v)] == Some(k)` when v belongs to the kth component. On the other hand, if we're going for the `Option<NonZeroUsize>` route and need a subtraction anyway since zero is ruled out, we can just flip them back to the original setup and have  `self.nodes[g.to_index(v)] == Some(usize::MAX - k)` instead.

saolof: Okay, Rustc optimizes usize::MAX - k properly to the ones complement of k, so that also basically costs nothing.
https://godbolt.org/z/GM8z7neEK

ABorgna: Great, that works then.

ABorgna:
I ran #421's benchmarks, and I'm seeing some noticeable performance regressions (I include Kosaraju's for reference): 

```
 name                        master ns/iter  new_tarjan ns/iter  diff ns/iter  diff %  speedup
 bigger_kosaraju_sccs        6,000           6,047                         47   0.78%   x 0.99
 bigger_tarjan_sccs          3,299           3,879                        580  17.58%   x 0.85
 full_kosaraju_sccs          1,047           1,047                          0   0.00%   x 1.00
 full_tarjan_sccs            538             658                          120  22.30%   x 0.82
 sccs_kosaraju_graph         2,687           2,607                        -80  -2.98%   x 1.03
 sccs_kosaraju_stable_graph  2,691           2,669                        -22  -0.82%   x 1.01
 sccs_tarjan_graph           1,809           2,134                        325  17.97%   x 0.85
 sccs_tarjan_stable_graph    1,755           2,087                        332  18.92%   x 0.84
```

I'll try to do some profiling.

ABorgna:
I pushed a small change to your branch, it checks if the node has been visited _before_ calling `visit` and hence avoids the recursive calls.
Now the benchmarks are much nicer :)

```
 name                        master ns/iter  new_tarjan_check ns/iter  diff ns/iter   diff %  speedup
 bigger_kosaraju_sccs        5,791           5,402                             -389   -6.72%   x 1.07
 bigger_tarjan_sccs          3,002           2,817                             -185   -6.16%   x 1.07
 full_kosaraju_sccs          1,156           976                               -180  -15.57%   x 1.18
 full_tarjan_sccs            487             438                                -49  -10.06%   x 1.11
 sccs_kosaraju_graph         2,405           2,367                              -38   -1.58%   x 1.02
 sccs_kosaraju_stable_graph  2,263           2,261                               -2   -0.09%   x 1.00
 sccs_tarjan_graph           1,460           1,249                             -211  -14.45%   x 1.17
 sccs_tarjan_stable_graph    1,478           1,223                             -255  -17.25%   x 1.21
```

ABorgna:
Uhm, I'm getting pretty noisy benchmarks (full_kosaraju_sccs there got -15% with no code changes).
But the Tarjan implementations is consistently faster with your implementation.

If you are OK with this I think the PR is ready to merge.

saolof:
Sounds good. When I have more time I'll work on an imperative version that can't stack overflow and microoptimize it in a followup PR. There's also a few other functions that depend on the SCC's such as condensation that could be rewritten to take advantage of the new node_component_index method.

PR #418, opened 2021-04-24
Implement Debug and Clone for iterators

### LINKED ISSUE #416: Struct `petgraph::graph::Externals` does not implement Clone
### Summary

The iterator struct `Externals` seems to have missed some trait implementations that it could have, most critically Clone.

### Motivation

Some cases, like the `cycle`-iterator adapter requires, require the iterator to be Clone which would make using the `Externals` struct a bit easier if it had that trait.
Also adding the Debug trait might be good too.

### Details

I don't see any problems from just adding an auto-derive for it, since the internal `iter::Enumerate` already has Clone (and Debug) too.
It also has some other traits implemented, such as `DoubleEndedIterator`, `ExactSizeIterator` etc, but I'm not sure if implementing those is necessary or compatible with the future API.

I will try to make a fork and a pull request for the change at some point unless someone wants to do it before me.

TheZalli:
I wanted to implement the Debug and Clone traits for `graph::Externals` (#416), but when looking at the source code I noticed that many iterator types didn't implement them either even when they could. So I went over every iterator type that I could find and added those implementations to where I could.

3 iterators that I found used an internal iterator that didn't implement them both so I skipped them but added a comment. They could be solved by making them a custom iterator but I feel like that's out of scope for this PR.

Fixes #416

### DIFF
--- a/src/adj.rs
+++ b/src/adj.rs
@@ -29,6 +29,7 @@ impl (Iterator) for
 /// An Iterator over the indices of the outgoing edges from a node.
 ///
 /// It does not borrow the graph during iteration.
+#[derive(Debug, Clone)]
 struct OutgoingEdgeIndices <Ix> where { Ix: IndexType }
 item: EdgeIndex<Ix>,
 iter: std::iter::Map<std::iter::Zip<Range<usize>, std::iter::Repeat<NodeIndex<Ix>>>, fn((usize, NodeIndex<Ix>)) -> EdgeIndex<Ix>>,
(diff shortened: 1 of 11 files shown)


ABorgna (approved):
[adj::EdgeIndices](https://github.com/petgraph/petgraph/pull/418/files#diff-a56d69a050eab3b56a403d052b255d3284f97ec8b9086ce93dbf78282f7fc70fR97) is missing a derive.

Lgtm. If we detect any other missing trait later it can be added in a minor revision.

TheZalli:
@ABorgna thanks! Seems like I missed that, but now it's there.

PR #421, opened 2021-04-26
Add benchmarks for Tarjan's SCC algorithm

ABorgna:
Rename the current SCC benchmarks to indicate they are using Kosaraju's algorithm, and add similar benchs for the Tarjan implementation.

The goal is to have a baseline for #413.

### DIFF
--- a/benches/matrix_graph.rs
+++ b/benches/matrix_graph.rs
@@ -198,9 +198,15 @@ fn full_neighbors_in(bench: &mut Bencher) {
 }
 
 #[bench]
-fn full_sccs(bench: &mut Bencher) {
+fn full_kosaraju_sccs(bench: &mut Bencher) {
     let a = parse_matrix::<Directed>(FULL);
     bench.iter(|| algo::kosaraju_scc(&a));
+}
+
+#[bench]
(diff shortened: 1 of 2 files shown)


saolof:
Adding ready-made benchmarks to a project is great and makes the contribution process a lot easier. One thing I noticed from the benchmarks in the other thread is that the tarjan_scc function is substantially faster than the Kosaraju implementation, roughly by a factor of 2 (makes sense, we're looking at one traversal vs two). 

This suggests that making Tarjan imperative so that it can't stack overflow and can replace Kosaraju as the default algorithm, is probably the right path to making SCC's faster.

ABorgna:
Yes, but the Tarjan implementation requires `NodeIndexable`, so it cannot be used with `GraphMap`. The `scc` method is deprecated, so there is no default implementation.
We could perhaps mention that Tarjan's tends to be a faster option in Kosaraju's documentation.

It will be nice to see the benchmarks for the imperative implementation :)

ABorgna:
I'll self-approve this since it's a simple change, and may be useful for contributors.

bluss:
IMO self-approve is a good way to work, when you feel like it is appropriate.

PR #424, opened 2021-05-10
Simplify/optimize petgraph::dot a bit

coolreader18:


coolreader18:
Oh, sure, sorry - so the 'optimization' part is mainly that `graph_fmt` was doing a linear search on `self.config` inside a loop (which, in retrospect, it was probably hoisting out of the loop anyway) as well as getting rid of the indirection with the `dyn FnMut` trait objects. The main thing though is that (IMO) this version is a lot simpler, with less trait bounds (some of them were redundant) and generics everywhere. Plus I got rid of some weird things that I'm sure were vestigial, like `graph_fmt` taking `g: G` even though it had access to `self.graph`.

coolreader18:
Also I could get rid of the macro + `Configs` struct if that seems too extra, I originally was gonna have `let mut config_var = false; for x in config { match x { ConfigVar => config_var = true } }` and then I realized it would be easier to make that its own struct, then that I could generate the struct + the configuration "parsing" with a macro, and then that I could store that inside `Dot` since `Dot.config` isn't mutable once it's created.

src/dot.rs
```diff
@@ -109,40 +112,62 @@ pub enum Config {
     /// Do not print the graph/digraph string.
     GraphContentOnly,
     #[doc(hidden)]
-    _Incomplete(()),
+    // can never be constructed
+    _Incomplete(std::convert::Infallible),
```
bluss, line 116: While interesting and something that has been considered before, having an inconstructible variant in some situations allows the user to omit it from match, which is sort of the opposite of what we want..

ABorgna (changes requested):
lgtm, it's definitely cleaner than before. The config struct seems like a good choice.
I agree with @bluss in that the Infallible should be removed.

Replacing the config flags by an EnumSet would be a nice change for 0.6, if we are ok with adding an extra dependency.

coolreader18:
Are you sure the `_Incomplete(Infallible)` can be omitted from match? It seems to require it for me: `non-exhaustive patterns: _Incomplete(_) not covered`

And yeah, having it be a bitset wouldn't even necessarily need `enumset` if you're ok with doing it manually, but it still would be a breaking change.

bluss:
Hm, it might have been a misfeature that has since been fixed as a bug! Sorry for the confusion. #[non_exhaustive] is the way to go, anyway

ABorgna (approved):
Let's merge this now and open another issue for bumping the MSRV for 0.6, so we can use non_exhaustive.

ABorgna:
And thanks for the contribution :)

PR #426, opened 2021-05-11
Fixed broken documentation links for Directed and Undirected enums.

john-hartley:
I'm not sure if there's an automated way to test for broken documentation links, as I'm new to Rust, so I've verified locally using `cargo doc --open`. 

Thanks for the crate all the same. I was tearing my hair out trying to figure out how to work with graphs.

### DIFF
--- a/src/adj.rs
+++ b/src/adj.rs
@@ -141,15 +141,15 @@ iterator_wrap! {
 /// Can be interpreted as a directed graph
 /// with unweighted nodes.
 ///
-/// This is the most simple adjacency list you can imagine. `Graph`, in contrast,
+/// This is the most simple adjacency list you can imagine. [`Graph`](../graph/struct.Graph.html), in contrast,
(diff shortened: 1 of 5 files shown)


john-hartley:
No problem!

I'm happy to fix the other errors, but I've run into a problem installing `cargo-deadlinks`, for which I've created an [issue](https://github.com/briansmith/ring/issues/1282) in one of its dependencies. Once I get that sorted, then I'll fix the rest.

john-hartley:
I've fixed the remaining problems reported by `cargo deadlinks`.

Where I've added `#[doc(hidden)]` to some methods it's because the traits that declare those methods have also marked them as `#[doc(hidden)]`. `deadlinks` was reporting that they don't exist because `#[doc(hidden)]` was applied to the trait but not the implementation.

ABorgna:
Great ☺
Thanks for the contribution.

PR #430, opened 2021-05-16
Fix double assigment in bellman_ford algo

oxarbitrage:
No big deal but i think the `i` removed here is not needed as it is populated before in `for i in g.node_identifiers()`


ABorgna:
Hi, thanks for the contribution :)

PR #437, opened 2021-06-16
Fix a formatting issue in the algo module's docs

abreis:
The (long) URL in the doc comment of `fn all_simple_paths` causes a formatting issue on  <https://docs.rs/petgraph/0.5.1/petgraph/algo/index.html> by preventing a line wrap.

This commit moves it to a new paragraph so that it gets omitted from the module's short description.

image

### DIFF
--- a/src/simple_paths.rs
+++ b/src/simple_paths.rs
@@ -10,10 +10,10 @@ use crate::{
     Direction::Outgoing,
 };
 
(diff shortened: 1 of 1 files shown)


ABorgna:
I could not reproduce the broken wrapping in firefox, but you are right in that it should be omitted in the short description.

Issue #428, opened 2021-05-14
Bump MSRV for release 0.6


ABorgna:
Since we are planning on doing a major release, we can update the Minimum Supported Rust Version (current one is 1.37).

- The last version of `fixedbitset` requires 1.39.
- #424 discusses using the `non_exaustive` attribute, which requires 1.40.

Should we go for 1.40? Anyone proposes something more recent?

mtreinish:
If possible could we avoid going above 1.41 (so 1.40 works quite well for me)? 1.41 is the version of rustc that debian buster is currently shipping and at least for some of my downstream user base it's very important to be able to compile the library I maintain that's built on petgraph with the debian version of rustc.

ABorgna:
Another datapoint:

- The no_std (#370) will add [`hashbrown`](https://github.com/rust-lang/hashbrown) as a dependency, which has a MSRV of **1.49**. hashbrown0.9.1 (from 2020-09-28) still supported 1.36, so we'd have to use that version if we keep to something lower.

ABorgna:
- Looks like IndexMap may require 1.49 soon since it also depends on hashbrown https://github.com/bluss/indexmap/pull/181.

mtreinish:
Unless you actually require the latest version (for api compatibility reasons, new features, or something else) you can just set a version range, something like: https://github.com/PyO3/pyo3/blob/main/Cargo.toml#L30. That way downstream crates that use it can use the version that meets their constraints. I actually use hashbrown in my crate (for the rayon iterator support) but have it fixed at 0.9 for rustc 1.41 support.

ABorgna:
I set the new version to 1.41 :)
We'll deal with the new indexmap and hashbrown versions when we need them.

PR #440, opened 2021-06-24
Fix minor clippy warnings

barafael:
I noticed there is quite a number of clippy warnings here. I fixed some of them; let me know what you think :)

BTW, I fixed the ones I deemed safe to change. Some of the fixes are targeting a relatively new version of Rust, for example introducing the `matches!` macro might break some builds with older compilers, so I didn't touch them.
Some of the remaining ones seem rather hard to fix to me, unfortunately :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,10 +18,16 @@ categories = ["data-structures"]
 
 edition = "2018"
 
+[package.metadata.docs.rs]
+features = ["serde-1", "quickcheck"]
+
+[package.metadata.release]
+no-dev-version = true
+
 [lib]
+bench = false
 
 name = "petgraph"
-bench = false
 
 [profile.release]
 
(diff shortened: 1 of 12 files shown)


src/matrix_graph.rs
```diff
@@ -128,10 +128,10 @@ impl<T: Zero> Nullable for NotZero<T> {
     }
 }
 
-impl<T: Zero> Into<Option<T>> for NotZero<T> {
```
ABorgna, line 131: The CI is breaking here because this is only allowed since 1.41 [1].
We are going to raise the MSRV (#428), so it shouldn't be a problem.

[1]: https://blog.rust-lang.org/2020/01/30/Rust-1.41.0.html#relaxed-restrictions-when-implementing-traits).

ABorgna (commented):
Thanks for these fixes :)

The merge will have to wait until we raise the MSRV in #428.

barafael:
Oh, I oversaw that one. In that case, we can either remove the breaking fixes, or add also the`matches!` fixes

ABorgna:
I went ahead and made the PR for the version bump to 1.41 (#444).
It may take a bit because it requires changing the CI config, but it should be merged soon.

ABorgna:
Done! It should work now if you rebase.

barafael:
While rebasing, I also ran `tomlfmt` on `Cargo.toml`. Please let me know if there are changes you want me to revert, like the unstable sort.

Some of the `matches!` stuff will be super nice in some later MSRV:
```rust
    self != &Label::None
        && !match self {
            Label::Flag(_) => true,
            _ => false,
        };
    // becomes:
    self != &Label::None && !matches!(self, Label::Flag(_))
```

Cargo.toml
```diff
@@ -30,37 +36,31 @@ debug = true
...
+# feature flags for testing use only
+all = ["unstable", "quickcheck", "matrix_graph", "stable_graph", "graphmap"]
 default = ["graphmap", "stable_graph", "matrix_graph"]
-graphmap = []
-serde-1 = ["serde", "serde_derive"]
-stable_graph = []
-matrix_graph = []
 
 # For unstable features
 generate = []
```
ABorgna, line 58: I'd move the comment here.
```suggestion
generate = [] # For unstable features
```

ABorgna (commented):
I didn't realise that `matches!` was added in 1.42 :/
We'll have to leave it for the next MSRV bump.

About the `tomlfmt` run, I think it's ok.

Cargo.toml
barafael: Fixed it :)

PR #444, opened 2021-06-26
Bump MSRV to 1.41, update dependencies

ABorgna:
Updates the MSRV as discussed in #428, and updates some dependencies with it.

Unfortunately we can't update quickcheck to 1.0 yet because we need to generate floats in the 0..1 range in various tests. See https://github.com/BurntSushi/quickcheck/issues/267.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - rust: 1.37.0  # MSRV
+          - rust: 1.41.0  # MSRV
           - rust: stable
             features: unstable quickcheck
             test_all: --all
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,17 +29,17 @@ bench = false
 debug = true
 
 [dependencies]
-fixedbitset = { version = "0.3.0", default-features = false }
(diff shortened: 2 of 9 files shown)


ABorgna:
@bluss Seems the "test (1.37.0)" check is set as required for the master branch.
I cannot update it as I don't have enough permissions in this repo, can you change it? Maybe require the non-numbered tests instead.

bluss:
indexmap will soon require Rust 1.49, because the latest hashbrown does

bluss:
Ok I found the setting for required tests, I've updated it to stable. I think @XVilka must have set it (and I guess you can set it too, not sure?)

ABorgna:
> Ok I found the setting for required tests, I've updated it to stable. I think @XVilka must have set it (and I guess you can set it too, not sure?)

Thanks. I could not change it because I'm not in the petgraph organization (you could add me if that's ok).

> indexmap (1.7 I guess) will soon require Rust 1.49, because the latest hashbrown does (just info of what's to come)

Yes. That was considered in #428, but it's not a pressing matter to update it.
@mtreinish suggested using a version range for the dependency, but I'm not sure the dependency resolution would work as intended (esp. since we don't have https://github.com/rust-lang/rust/issues/65262 yet).

bluss:
Sure, added

PR #455, opened 2021-07-07
Fix the 0.6 MSRV as 1.41

cuviper:


RELEASES.rst
```diff
@@ -521,4 +521,4 @@ Version 0.1.6 (2015-06-04)
 Version 0.1.4 (2015-05-20)
 ==========================
 
-- Add back DfsIter, BfsIter
\ No newline at end of file
+- Add back DfsIter, BfsIter
```
cuviper, line 524: FWIW, this was the GitHub editor's doing -- I think it's a "no newline at end of file" change.

ABorgna (approved):
Ouch, thanks for double checking that :)

PR #458, opened 2021-08-01
Add methods Graph::nodes and Graph::nodes_mut

mankinskin:
methods for iterators of `(NodeIndex, &N)` and `(NodeIndex, &mut N)`

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -1013,6 +1013,20 @@ where
         }
     }
 
+    /// Return an iterator over the node indices and weights of the graph.
+    ///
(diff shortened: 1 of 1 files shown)


mankinskin:
Apparently this is also intended by `[NodeReferences`](https://docs.rs/petgraph/0.6.0/petgraph/graph/struct.NodeReferences.html), however there is no way of creating a borrowed iterator, there is only the [`IntoNodeReferences`](https://docs.rs/petgraph/0.6.0/petgraph/visit/trait.IntoNodeReferences.html) trait, which moves the graph. Thus there is not iterator over mutable references either.

bluss:
IntoNodeReferences does not move the graph, it's implemented on `&G` where G is a graph.

PR #461, opened 2021-08-03
Export missing graph::{EdgeWeights,NodeWeights}

stevenengler:
The `EdgeWeights` and `NodeWeights` structs weren't exported, so [`node_weights()`](https://docs.rs/petgraph/0.6.0/petgraph/graph/struct.Graph.html#method.node_weights) and [`edge_weights()`](https://docs.rs/petgraph/0.6.0/petgraph/graph/struct.Graph.html#method.edge_weights) weren't usable.


bluss:
They are usable but not fully usable. In the long run the awkward graph_impl module should probably go away. (Separate PR/issue).

PR #472, opened 2021-11-30
Fix subgraph isomorphism

### LINKED ISSUE #401: is_isomorphic_subgraph fails on certain graphs
### Test Case

```
use petgraph::graph::Graph;
use petgraph::algo::is_isomorphic_subgraph;

fn isomorphism_check() {
    let mut graph0 = Graph::<&str, &str>::new();
    let a = graph0.add_node("a");
    let b = graph0.add_node("b");
    let c = graph0.add_node("c");
    let e1 = graph0.add_edge(a, b, "");
    let e1 = graph0.add_edge(b, c, "");

    let mut graph1 = Graph::<&str, &str>::new();
    let x = graph1.add_node("x");
    let y = graph1.add_node("y");
    let z = graph1.add_node("z");
    let w = graph1.add_node("w");
    let e2 = graph1.add_edge(x,y,"");
    let e2 = graph1.add_edge(y,z,"");
    let e2 = graph1.add_edge(x,w,"");

    assert!(is_isomorphic_subgraph(&graph0, &graph1));
}
```
### Steps To Reproduce

* This bug only happens with the version of petgraph obtained by cloning the github repository.  It looks like subgraph isomorphism is a new feature that has not been released yet but has been merged into the repository.  However, when it was merged, only one test case was given.  It fails on other simple graphs.
* Once that version is used, the test above fails.

### Actual Results

The code above will fail on the assert, despite the fact that the graphs are subgraph isomorphic.  Graph0 is just a string of three nodes, and graph1 is a string of 3 nodes with an extra node added to the root.

### Expected Results

One would expect that the assert would not fail.

cojuer:
Cardinality check was incorrect in case of subgraph isomorphism.

Fixes #401

### DIFF
--- a/src/algo/isomorphism.rs
+++ b/src/algo/isomorphism.rs
@@ -555,6 +555,7 @@ mod matching {
         mut st: &mut (Vf2State<'_, G0>, Vf2State<'_, G1>),
         node_match: &mut NM,
         edge_match: &mut EM,
+        match_subgraph: bool,
     ) -> Option<bool>
     where
         G0: NodeCompactIndexable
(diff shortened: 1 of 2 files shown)


sergiosmcr:
Hello. 
I am interested in these changes. Can anyone review and merge this please.
Not sure who to tag, what about @jkeljo (author of the most recent commit in master)

ABorgna:
Thanks for the fix !

PR #473, opened 2021-12-06
Algorithms for finding bridges and articulation points.

starovoid:
Two algorithms are implemented: for finding [bridges](https://www.geeksforgeeks.org/bridge-in-a-graph/) and for finding [points of articulation](https://www.geeksforgeeks.org/articulation-points-or-cut-vertices-in-a-graph/) are implemented, each of which expects the argument to be a simple undirected graph. The algorithms are tested with all types of graphs.

### DIFF
--- a/benches/bridges_and_ap.rs
+++ b/benches/bridges_and_ap.rs
@@ -1,0 +1,45 @@
+#![feature(test)]
+
+extern crate petgraph;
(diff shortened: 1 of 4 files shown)


ecashin:
Hi.  Why was the pull request closed?  I don't see any conversation about it.

PR #476, opened 2021-12-14
Fix IntoEdgesDirected implementation for NodeFiltered when direction is Incoming

jkeljo:
Calling it with Direction::Incoming would not filter the edges at all.
Existing tests didn't catch it because they did not have an edge going
from an excluded node to an included one.


ABorgna:
Thanks for the fix!

PR #479, opened 2022-01-16
[FIX-BIPARTITE-CHECK] check should identify disconnected cycles

a1t20v21:
The existing function do not check disconnected components,


a1t20v21:
Can we merge this PR? This will fix the CI failures too.

PR #485, opened 2022-02-25
Add a version of condensation which remembers node indices rather than their weight

FlorentBecker:
Currently, there is no way to get the original nodes of g from the nodes of `algo::condensation(g, _)`. This PR adds an `algo::condensation_indices` which has vector of node indices from g as the node weights of its return value. This helps when you use the condensed version as a guide to work with the original graph (for instance for doing layout). The edge weights are `()` since there's hardly another labeling which makes sense with `make_acyclic = true` (as well as `make_acyclic = false`).

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -634,6 +634,91 @@ where
     condensed
 }
 
+
+/// [Graph] Condense every strongly connected component into a single node and return the result.
+///
+/// The nodes in the condensed version of the graph are labelled with
+/// the indices of the nodes they stand for.  The edges of the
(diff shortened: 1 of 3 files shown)


NickHu:
You can do this manually by storing node indices as weights in the original graph, possibly tupled with any other node weight data required

FlorentBecker:
The upside of the `condensation_indices` function compared to storing indices as weights is that you don't need to copy the graph: `condensation_indices` takes a `&Graph` argument rather than a `Graph`. Thus, `condensation` can be expressed using `codensation_indices`, but not quite the other way around.

PR #487, opened 2022-03-16
Fix Floyd-Warshall algorithm behavior toward undirected graphs

At-sushi:
Feature in this patch:

- Fixed a bug that Floyd-Warshall algorithm doesn't work properly to undirected graphs. (It is supposed to work bidirectional when the graph is undirected.)

This patch adds a reference to GraphProp trait into the function petgraph::algo::floyd_warshall.


ABorgna:
LGTM. Thanks !

ABorgna:
LGTM. Thanks !

PR #491, opened 2022-05-10
docs: Clarify when graph methods return None?

beyarkay:
This is a docs change, and should not affect any functionality.

While looking about the API, I didn't quite understand why I had to `unwrap()` every time I wanted to access an edge or a node. I understand now it's because that weight or node might not exist, but that was only clear to me after looking at the source code.

This change adds a line to the relevant docstrings that explain when a `Some()` will be returned vs when a `None` will be returned.

The docstring for `remove_node` already had a similar line in it:
```
/// Remove `a` from the graph if it exists, and return its weight.
/// If it doesn't exist in the graph, return `None`.
```
So I made sure the lines I added were phrased similarly to keep the code base consistent, like 
```
/// If ____ doesn't exist in the graph, return `None`.
```

Hope this is helpful, let me know about any changes you'd like made!

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -534,6 +534,7 @@ where
 
     /// Access the weight for node `a`.
     ///
+    /// If node `a` doesn't exist in the graph, return `None`.
     /// Also available with indexing syntax: `&graph[a]`.
     pub fn node_weight(&self, a: NodeIndex<Ix>) -> Option<&N> {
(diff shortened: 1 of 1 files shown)


beyarkay:
Is this repo still maintained?

ABorgna:
Hi, sorry for the delay. Thanks for adding the clarifications :)

beyarkay:
No worries! Thanks for the great project

PR #493, opened 2022-05-22
Fix CI by fixing indexmap minor version

ABorgna:
`indexmap` changed the MSRV in a minor version, so 1.8 does not compile in rust 1.41

This is a quick fix pinning `indexmap` to `~1.7`.
Ideally, we should bump our MSRV in the next release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ debug = true
 
 [dependencies]
 fixedbitset = { version = "0.4.0", default-features = false }
-indexmap = { version = "1.6.2" }
+indexmap = { version = "~1.7" }
 quickcheck = { optional = true, version = "0.8", default-features = false }
 serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }


cuviper:
This is harmful to anyone who wants to use newer `indexmap` and newer `petgraph` at the same time, if they don't need that MSRV.

The way I've handled this for crates like `rayon` with a conservative MSRV is to downgrade the dependency in CI for compat testing, while leaving it unrestricted for users. That does mean that users of the old MSRV also have to deal with it though, like `cargo update -p indexmap --precise 1.7.0`, or add their own `indexmap = "~1.7"` version lock even though it's otherwise not their direct dependency. I feel that's a reasonable compromise.

ABorgna:
You are totally right. I guess we should just bump the MSRV since that's the plan and it'd take the same amount of work.

shepmaster:
> The way I've handled this for crates like `rayon` with a conservative MSRV is to downgrade the dependency in CI for compat testing, while leaving it unrestricted for users

[Ditto for SNAFU](https://github.com/shepmaster/snafu/blob/b552522ca92031a4450c914dce114013af383b7c/compatibility-tests/v1_34/Cargo.toml#L10-L12).

As a concrete pain point, the current decision prevents the playground from upgrading petgraph / indexmap (and by laziness all the other dependencies).

ABorgna:
Ok, I'll loose the dependency version and pin it in CI.
I'll probably add a comment about it in the readme when stating our MSRV.

ABorgna:
This is fixed now. Thanks for the comments, and sorry for the inconvenience :)

PR #496, opened 2022-06-06
Add serde support for graphmap

Indeximal:
This PR adds serde-1 support for the `GraphMap` struct. Unlike #341 two years ago, I addressed the issues of internal data and format stability by depending on the already existing serialization of the `Graph` struct.
This might not be the most performant solution and require the edges to be `Clone` as well, but was sufficiently easy to implement and is less error-prone.
I will try these changes out on my own projects in the coming weeks and obviously fix any issues I find. Let me know if any of you spot any mistakes or have improvement suggestions.

### DIFF
--- a/tests/graphmap.rs
+++ b/tests/graphmap.rs
@@ -326,6 +326,26 @@ fn test_into_graph() {
 }
 
 #[test]
+fn test_from_graph() {
+    let mut gr: Graph<u32, u32, Directed> = Graph::new();
+    let node_a = gr.add_node(12);
+    let node_b = gr.add_node(13);
+    let node_c = gr.add_node(14);
+    gr.add_edge(node_a, node_b, 1000);
+    gr.add_edge(node_b, node_c, 999);
+    gr.add_edge(node_c, node_a, 1111);
+    gr.add_node(42);
+    let gr = gr;
+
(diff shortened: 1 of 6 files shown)


tests/graphmap.rs
```diff
@@ -119,7 +119,7 @@ fn remove_node() {
     graph.remove_node(2);
 
     let neighbors: Vec<u32> = graph.neighbors(1).collect();
-    assert_eq!(neighbors, []);
+    assert_eq!(neighbors, [] as [u32; 0]);
```
Indeximal, line 122: This was necessary, because the new serde_json test dependency also implemented partial_eq for those types which lead to a conflict.

ABorgna (approved):
Thanks for the PR ! It's a nice straightforward implementation. Using the same serialized format as `Graph` is the best option I think.
Did you compare the performance difference with your previous PR ?

Indeximal:
No, I couldn't get @schulzch's PR to merge into the new version and abandoned testing it. I might try again tho.
Also, I thought it would be useful to move the graphmap building of the deserialization logic into a function `from_graph` that can be used as a standalone. This is implemented, but lacking tests at the moment.
I have discovered that there is a seralization_tests module, to which I've added my tests, which avoids adding the serde dev-dependency to the main module. However, cargo does not run these tests automatically, which has to be adressed, but I know too little about cargo tests.

Issue #497, opened 2022-06-08
Add functions to obtain isomorphisms between graphs


vlvrd:
### Summary

Add functions to the `isomorphism` module that return the possible mappings between isomorphic graphs.

### Motivation

Currently, Petgraph is only comparing graphs to indicate whether or not graphs are isomorphic. It doesn't allow users to explore the isomorphism solutions, which may be of interest too.

### Details

I suggest updating the `try_match` function from the isomorphism `matching` private module to return an iterator that contains the isomorphisms between two graphs, then adding public `isomorphism` functions to return those mappings. This is similar to what NetworkX does in the functions [isomorphisms_iter](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.isomorphism.GraphMatcher.isomorphisms_iter.html#networkx.algorithms.isomorphism.GraphMatcher.isomorphisms_iter) and [subgraph_isomorphisms_iter](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.isomorphism.GraphMatcher.subgraph_isomorphisms_iter.html#networkx.algorithms.isomorphism.GraphMatcher.subgraph_isomorphisms_iter).

I'm going to implement this myself in my fork since I need these mappings anyways, so if there's interest on whether or not these changes should make their way upstream I'll be happy to discuss them here and hopefully send a PR later.

PR #500, opened 2022-06-23
Add function that returns an iterator with subgraph isomorphisms

vlvrd:
This PR adds `subgraph_isomorphism_iter` to the `isomorphism` module to provide users with the capability of using the isomorphism solutions however they may need them. I will use these changes in my own project so I'll come back to this if there's something to be fixed. Please let me know if you have improvement suggestions.

This PR addresses issue #497 and includes @cojuer subgraph isomorphism fix from PR #472.

### DIFF
--- a/src/algo/isomorphism.rs
+++ b/src/algo/isomorphism.rs
@@ -1,3 +1,5 @@
+use std::convert::TryFrom;
+
 use crate::data::DataMap;
 use crate::visit::EdgeCount;
 use crate::visit::EdgeRef;
@@ -754,6 +756,49 @@ mod matching {
                 self.match_subgraph,
                 &mut self.stack,
             )
+        }
+
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            // To calculate the upper bound of results we use n! where n is the
+            // number of nodes in graph 1. n! values fit into a 64-bit usize up
+            // to n = 20, so we don't estimate an upper limit for n > 20.
+            let n = self.st.0.graph.node_count();
+            const MAX_N: usize = 20;
+
+            if n > MAX_N {
+                return (0, None);
+            }
+
+            // We hardcode n! values into an array that accounts for architectures
+            // with smaller usizes to get our upper bound.
+            let upper_bounds: Vec<Option<usize>> = vec![
+                1u64,
+                1,
+                2,
+                6,
+                24,
+                120,
+                720,
+                5040,
+                40320,
+                362880,
+                3628800,
+                39916800,
+                479001600,
+                6227020800,
+                87178291200,
+                1307674368000,
+                20922789888000,
+                355687428096000,
+                6402373705728000,
+                121645100408832000,
+                2432902008176640000,
+            ]
+            .iter()
+            .map(|n| usize::try_from(*n).ok())
+            .collect();
+
+            (0, upper_bounds[n])
         }
     }
 }
--- a/src/algo/isomorphism.rs
+++ b/src/algo/isomorphism.rs
@@ -763,11 +763,6 @@ mod matching {
             // number of nodes in graph 1. n! values fit into a 64-bit usize up
             // to n = 20, so we don't estimate an upper limit for n > 20.
             let n = self.st.0.graph.node_count();
-            const MAX_N: usize = 20;
-
-            if n > MAX_N {
-                return (0, None);
-            }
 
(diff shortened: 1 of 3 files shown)


src/algo/isomorphism.rs
```diff
@@ -631,8 +667,94 @@ mod matching {
...
+        EM: EdgeMatcher<G0, G1>,
+    {
+        st: (Vf2State<'a, G0>, Vf2State<'b, G1>),
+        node_match: &'c mut NM,
+        edge_match: &'c mut EM,
+        match_subgraph: bool,
+        stack: Vec<Frame<G0, G1>>,
+    }
+
+    impl<'a, 'b, 'c, G0, G1, NM, EM> GraphMatcher<'a, 'b, 'c, G0, G1, NM, EM>
```
ABorgna, line 699: Can you implement Debug and Clone for the iterator ?

src/algo/isomorphism.rs
```diff
@@ -631,8 +667,94 @@ mod matching {
...
+            let stack = vec![Frame::Outer];
+            Self {
+                st: (Vf2State::new(g0), Vf2State::new(g1)),
+                node_match,
+                edge_match,
+                match_subgraph,
+                stack,
+            }
+        }
+    }
```
ABorgna, line 730: Add a `size_hint` upper bound when the graphs are small enough.

ABorgna (changes requested):
Thanks for this ! I just have a couple requests for the iterator object.

Clippy is complaining about all the `&mut st` passed as parameters. Can you replace them with `st` ?

src/algo/isomorphism.rs
sergiosmcr: Are you referring to [size_hint method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint)?
Hmm. I don't think this is deterministic as it depends on the graph. The worst case scenario where every permutation is an isomorphism has a time complexity of O(n! * n) according to [this paper](https://www.researchgate.net/publication/224279054_Performance_evaluation_of_the_VF_graph_matching_algorithm). So should the upper bound for the size_hint be `n! * n` ? If so, should we return None if n is too large (when calculating `n!` is too expensive)?

This is a screenshot of the relevant image in that article:
image

ABorgna: Yes. It's mostly interesting for graphs with just a couple nodes, but you can encode the values that fit in an usize, for n≤19.

sergiosmcr: You are right. Time complexity is different from the actual number of solutions. I said that because I couldn't find a quick reference to talk about it. The formula you suggest calculates the number of permutations which is definitely the worst case when n=m.

ABorgna:
I'll merge #472 in the meantime, since there's been some interest for the fix.

vlvrd:
Thanks @ABorgna! Sorry I got a bit sidetracked, I just updated my branch with that merge and two commits with the size hint and removing the unused `mut`. Regarding the Debug and Clone traits, do you want me to implement them for the `GraphMatcher` struct or for the item of the iterator? Since the item is a `Vec<usize>` it already implements those traits.

sergiosmcr (approved):

src/algo/isomorphism.rs
```diff
@@ -638,8 +669,137 @@ mod matching {
...
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            // To calculate the upper bound of results we use n! where n is the
+            // number of nodes in graph 1. n! values fit into a 64-bit usize up
+            // to n = 20, so we don't estimate an upper limit for n > 20.
+            let n = self.st.0.graph.node_count();
+            const MAX_N: usize = 20;
+
+            if n > MAX_N {
+                return (0, None);
+            }
```
ABorgna, line 770:
```suggestion
```

src/algo/isomorphism.rs
```diff
@@ -638,8 +669,137 @@ mod matching {
...
+                20922789888000,
+                355687428096000,
+                6402373705728000,
+                121645100408832000,
+                2432902008176640000,
+            ]
+            .iter()
+            .map(|n| usize::try_from(*n).ok())
+            .collect();
+
```
ABorgna, line 800:
```suggestion

            if n > upper_bounds.len() {
                return (0, None);
            }
            
```

ABorgna (approved):

Issue #501, opened 2022-07-06
[Bug] `graph.neighbors_undirected(n0).detach()` returns node index 4294967295 


jerry73204:
### Steps To Reproduce

The bug is related to `graph.neighbors_undirected(n0).detach()`, which gives a neighbor walker surrounding a specified node.
To reproduce, get the walker first. Then, remove some of the neighbors `remove_edge()` before walking through the neighbors.
In the later walking loop, it returns the node index 4294967295 if the node is removed earlier.

```rust
use petgraph::{graph::NodeIndex, stable_graph::StableUnGraph};

#[test]
fn neighbor_detach_test() {
    let mut graph = StableUnGraph::default();
    let n0: NodeIndex = graph.add_node(());
    let n1 = graph.add_node(());
    let n2 = graph.add_node(());

    let e01 = graph.add_edge(n0, n1, ());
    let e02 = graph.add_edge(n0, n2, ());
    let e12 = graph.add_edge(n1, n2, ());

    println!("Create edges:");
    println!("{}: {} -- {}", e01.index(), n0.index(), n1.index());
    println!("{}: {} -- {}", e02.index(), n0.index(), n2.index());
    println!("{}: {} -- {}", e12.index(), n1.index(), n2.index());
    println!();

    let mut neighbors = graph.neighbors_undirected(n0).detach();
    graph.remove_edge(e01);
    graph.remove_edge(e02);

    while let Some((eidx, nidx)) = neighbors.next(&graph) {
        let neighbor = graph.node_weight(nidx);
        if neighbor.is_none() {
            println!("node {} is not found", nidx.index());
        }

        let edge = graph.remove_edge(eidx);

        if edge.is_none() {
            println!("edge {} is not found", eidx.index());
        }
    }
}
```

### Actual Results

The test code prints the following:

```
Create edges:
0: 0 -- 1
1: 0 -- 2
2: 1 -- 2

node 4294967295 is not found
edge 1 is not found
node 4294967295 is not found
edge 0 is not found
```

### Expected Results

4294967295 is a special value (-1 in signed integer) indicating the node does not exist. It is expected not being returned from the walker.

PR #504, opened 2022-07-24
fixes WalkNeighbors::next advancing to removed edge

aneksteind:
I'm not sure if this is what you'd like, but figured it would at least start a conversation that will lead us to that.

Addresses bug outlined in issue [#501](https://github.com/petgraph/petgraph/issues/501)

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -1559,7 +1559,8 @@ where
     /// neighbors and edges from the origin node.
     ///
     /// Note: The walker does not borrow from the graph, this is to allow mixing
-    /// edge walking with mutating the graph's weights.
+    /// edge walking with mutating the graph's weights. This means that any
+    /// subsequent modification of the graph's structure invalidates the walker.
     pub fn detach(&self) -> WalkNeighbors<Ix> {
         WalkNeighbors {
             skip_start: self.skip_start,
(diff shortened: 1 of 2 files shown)


src/graph_impl/mod.rs
```diff
@@ -2010,14 +2010,16 @@ impl<Ix: IndexType> WalkNeighbors<Ix> {
         &mut self,
         g: &Graph<N, E, Ty, Ix>,
     ) -> Option<(EdgeIndex<Ix>, NodeIndex<Ix>)> {
+        let removed = |e: &Edge<E, Ix>| e.node == [NodeIndex::end(), NodeIndex::end()];
+
         // First any outgoing edges
         match g.edges.get(self.next[0].index()) {
-            None => {}
-            Some(edge) => {
+            Some(edge) if !removed(edge) => {
```
ABorgna, line 2017: An `if let` followed by and `if` may be more compact.

ABorgna (changes requested):
While this change is a more sane behaviour, I think the main problem is that `detach` is underspecified when the graph connectivity changes. It may be the case that some edges are deleted, but others are added with the previous id.

For example, this example modified from the one in the original issue incorrectly says that n0 has 2 and 3 as neighbours:
```rust
use petgraph::{graph::NodeIndex, stable_graph::StableUnGraph};

fn main() {
    let mut graph = StableUnGraph::default();
    let n0: NodeIndex = graph.add_node(());
    let n1 = graph.add_node(());
    let n2 = graph.add_node(());
    let n3 = graph.add_node(());

    let e01 = graph.add_edge(n0, n1, ());
    let e02 = graph.add_edge(n0, n2, ());
    let e12 = graph.add_edge(n1, n2, ());

    println!("Create edges:");
    println!("{}: {} -- {}", e01.index(), n0.index(), n1.index());
    println!("{}: {} -- {}", e02.index(), n0.index(), n2.index());
    println!("{}: {} -- {}", e12.index(), n1.index(), n2.index());
    println!();

    let mut neighbors = graph.neighbors_undirected(n0).detach();
    graph.remove_edge(e01);
    graph.remove_edge(e02);

    println!(
        "Detach and remove edges {} and {}",
        e01.index(),
        e02.index()
    );
    println!();

    let e13 = graph.add_edge(n1, n3, ());
    let e23 = graph.add_edge(n2, n3, ());

    println!("Create edges:");
    println!("{}: {} -- {}", e13.index(), n1.index(), n3.index());
    println!("{}: {} -- {}", e23.index(), n2.index(), n3.index());
    println!();

    let n0_neighs: Vec<_> = std::iter::repeat(())
        .map_while(|_| Some(neighbors.next(&graph)?.1.index()))
        .collect();

    println!("Node {} neighbors:", n0.index());
    println!("{:?}", n0_neighs);
}
```
There doesn't seem to be a way to fix this.

I'm OK with merging this change, but more importantly we should state that modifying the graph structure invalidates the Walker in the docs for `detach` (both for StableGraph and for Graph) .
I also think the test should be removed as it shows an invalid operation.

aneksteind:
Yeah, I agree with you that this is fundamentally about the design of `WalkNeighbors`.

src/graph_impl/mod.rs
```diff
@@ -2010,20 +2011,21 @@ impl<Ix: IndexType> WalkNeighbors<Ix> {
         &mut self,
         g: &Graph<N, E, Ty, Ix>,
     ) -> Option<(EdgeIndex<Ix>, NodeIndex<Ix>)> {
+        let edge_present: for<'r> fn(&'r &Edge<E, Ix>) -> bool =
```
ABorgna, line 2014: This check should be in StableGraph, as before. Graph automatically compresses the nodes on deletion, so there shouldn't be empty ones. (`NodeIndex::end()` may be a valid identifier).

The comment addition should be the only thing needed for `Graph`.

aneksteind: Sorry, I didn't see at this should be moved to `stable_graph::WalkNeighbors.next` -- that just calls this function with a one-liner to `self.inner.next`. Would you prefer to change that instead? As far as `Graph` goes, if `NodeIndex::end()` is present for both nodes then it seems like there wouldn't be any adverse consequences for detecting that here, since it would imply no edges remain anyway. Have I missed something?

PR #516, opened 2022-10-08
Make ensure_len use resize_with

jujhar16:
This change has greatly improved the performance of my application, in which I create large `MatrixGraph`s.

### DIFF
--- a/src/matrix_graph.rs
+++ b/src/matrix_graph.rs
@@ -897,12 +897,7 @@ fn extend_lower_triangular_matrix<T: Default>(
 
 /// Grow a Vec by appending the type's default value until the `size` is reached.
 fn ensure_len<T: Default>(v: &mut Vec<T>, size: usize) {
-    if let Some(n) = size.checked_sub(v.len()) {
-        v.reserve(n);
-        for _ in 0..n {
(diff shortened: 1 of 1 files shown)


XVilka (approved):
Looks good, and works since Rust 1.33:

Resizes the Vec in-place so that len is equal to new_len.

If new_len is greater than len, the Vec is extended by the difference, with each additional slot filled with the result of calling the closure f. The return values from f will end up in the Vec in the order they have been generated.

If new_len is less than len, the Vec is simply truncated.

This method uses a closure to create new values on every push. If you’d rather [Clone](https://doc.rust-lang.org/std/clone/trait.Clone.html) a given value, use [Vec::resize](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize). If you want to use the [Default](https://doc.rust-lang.org/std/default/trait.Default.html) trait to generate values, you can pass [Default::default](https://doc.rust-lang.org/std/default/trait.Default.html#tymethod.default) as the second argument.

https://doc.rust-lang.org/std/vec/struct.Vec.html

XVilka:
Could you please rebase to trigger the CI?

XVilka (approved):

PR #533, opened 2023-01-16
Add reverse method to StableGraph

nicholasgodfreyclarke:
Add reverse() method to StableGraph.
The implementation here is basically the same to the reverse method on Graph.


ABorgna:
Thanks !

Issue #518, opened 2022-10-13
[Feature request] Generate complete graph


jonesmartins:
### Summary

Generate complete graph from a given size, or from given node values.

### Motivation

Ideally, it would give an optimized implementation of a complete graph instead of leaving it for the user to implement.

### Details

* What code needs to change? New traits? Which modules?

* Is this a new graph algorithm? Include links to papers, Wikipedia, or other
  implementations of this algorithm that exist.

An example from Python's NetworkX: https://networkx.org/documentation/stable/_modules/networkx/generators/classic.html#complete_graph

* Are you willing to implement this yourself? 

Yes. I'm new to Rust, though. I'm not sure I'd write the best implementation for it.

PR #543, opened 2023-03-03
[Feature request] Generate complete graph https://github.com/petgraph/petgraph/issues/518

insideoutclub:
I'd welcome any comments particularly on the complete_graph interface and where it should live in the module hierarchy. I envision generators for other kinds of graphs being added to the generators module. Does the generators module need a new name given that there's already a generator module?


insideoutclub:
I'm still experimenting with the right interface for complete_graph.

Issue #551, opened 2023-05-11
Proposal: Future of this crate


indietyp:
Hi!

Thank you all for creating such an incredible library. I use `petgraph` nearly on a day-to-day basis in both my professional and personal projects. After observing progress (and the different issues) over the last couple of months, I thought that it might make sense for me to contribute to `petgraph` itself, not only potentially adding some new functionality but also maybe lead `petgraph` into a new direction while maintaining its usefulness and broad adoption.

After thinking about it, I have come up with some action points I'd like to realize, and I know they are pretty drastic. Nevertheless, to not duplicate work or implementation, it makes sense to build onto the foundation of the existing `petgraph`.

If this is suitable for this crate, please let me know, and I'll start working on it. I still have some open questions, like: 

* How would we go about code review/PRs? Do I create intermittent PRs that add functionality, slowly transforming everything, or have one big blob that will then be reviewed/merged?
* Should this be done on the main branch? (maybe a dedicated `next` branch might make sense)
* I want to commit to being a long-term maintainer of the crate. Is this alright? Is this okay?

Thank you so much for considering all of this! (and I guess this is where I need to mention @bluss?) I look forward to working with you all!

I am open to having a lively discussion and going more in-depth, potentially leading to a change in plans.

### 1. Reorganization

Split up the crate:
* `petgraph-core`: Includes all traits. To reduce churn, we should consider stabilizing this first and avoid breaking changes. This way, other crates can feel comfortable implementing their algorithms onto it.
* `petgraph-graph`: includes `Graph` and `StableGraph` (as they are commonly used interchangeably, I'd bundle them in a single crate)
* `petgraph-csr`: includes `CSR` graph
* `petgraph-matrix`: includes `MatrixGraph` (name might need changes, to not confuse with matrices in a non-graph type sense)
* `petgraph-algorithms`: includes all algorithms currently in petgraph; we might want to feature gate them (not only `scc` and friends, but also, if possible, `Bfs`)
* `petgraph-dot`: support for `dot` reading (and maybe writing in the future)
* `petgraph`: re-export of all sub-crates, main entry point for binaries

(names are ofc up to debate)

### 2. GATification

This would raise the MSRV to 1.65 but enable a lot more flexibility; every trait that could benefit from this should be converted into its GAT form, allowing things like mutable vs. reference access.

### 3. Trait rework

Think if we can extend the current set of traits to enable more flexible approaches, things like: `Filter`, `FilterMap`, and `Map` as traits.

[...]

PR #560, opened 2023-05-14
Initial scaffolding

indietyp:
This does internal "rejuvination" (outlined under 7 of #551)

# Changes

* raise MSRV to `1.65`
* replace `Makefile` with `just`
	* permutations are tested using `cargo-hack`, tests are run through `nextest`
* `RELEASES.rst` has been converted to `CHANGELOG.md` (to be consistent with the rest of the Rust ecosystem)
	* now adheres to https://keepachangelog.com/en/1.1.0/
* `custom.css` is not properly added to the docs output
* `custom.css` and `graph-example.dot` have been moved to `doc/`
* I have added currently active maintainers in the list of authors
* I have added a `rust-toolchain.toml` and (currently empty) `rustfmt.toml`
	* I'd like to add a bit more of an "opinionated" touch to the `rustfmt.toml`, adding more settings, they are more of personal taste (but I think they improve rustfmt quite a lot), so I haven't applied them just yet. Please let me know if I can/should!
* ci has been revamped, with 3 stages, that use the justfile.
	* currently checks coverage but does _not_ upload it to codecov or similar, we might want to do that in the future.
* added an exhaustive list of clippy lints
* the generated `graph-example.svg` is now part of the repository, CI checks that the output of `dot` matches the current file, and will fail otherwise. This way it is possible for people to depend on the repository directly.

# Questions

What is the following snippet about?

```
[package.metadata.release]
no-dev-version = true
```

Is `package.metadata.docs.rs` needed?

I have gone through the git history and have tagged every release locally. I plan to publish those tags as referenced in the new `CHANGELOG.md` is this okay? Because we plan on publishing multiple libraries, I have adopted the common pattern of `petgraph@v<version>`


indietyp:
There are a lot of errors in CI right now, so for now, I will disable it for the `next` branch because we expect many things to change. I am not going to bother with lints right now, but will apply them gradually when changing things in the next steps.

indietyp:
it seems like it takes an extraordinarily long time to run a single run (194s); most of that time is spent on quickcheck. I wonder if we'd be able to make this faster by switching to proptest (which in contrast to `quickcheck` seems to be actively maintained), as well as selectively removing features/requiring features. This will need more testing. Even if we switch to proptest, we should still provide quickcheck integration! I think providing for both would be a net benefit.

We also have a high number of test permutations. I think that is because we currently do not use `dep:` dependencies, this will change in upcoming PRs.

Currently, it would take 15 hours to run all tests, and I don't think that's feasible 😅

bluss:
`package.metadata.release` is from cargo-release. If we don't want to use it, then it's not necessary. The docs.rs metadata - yes it's needed so that generated docs contains traits from those features. We can consider using all-features = true instead in the metadata, or cherry-pick features to use.

bluss (approved):
Looks good to me, nice way to go

indietyp:
I was thinking of using cargo-release anyway, so it's good to know that my suspicions were confirmed!

Issue #559, opened 2023-05-14
Trait Rework: New traits


indietyp:
This proposes several new traits that are useful in manipulating existing graphs.

```rust
pub trait GraphMap {
	type Output;

	fn map<'a, F, G, N, E>(&self, nodes: F, edges: G) -> Self::Output<N, E>;
}

// mutable access could be modelled over `for node in g.nodes_mut() { node = ... }`

pub trait GraphFilterMap {
	type Output;

	fn filter_map<'a, F, G, N, E>(&self, nodes: F, edges: G) -> Self::Output<N, E>;
}

pub trait GraphRetain {
	fn retain_nodes<F>(&mut self, visit: F);
	fn retain_edges<F>(&mut self, visit: F);
}
```

These are modeled after `Graph` (and `StableGraph`) as they are currently the most fully featured.

We might want to think about the following:

* `edge_connecting()`
* `contains_edge()`
* `find_edge()`
* `find_edge_undirected()`
* `externals()`
* `node_weights()`
* `node_weights_mut()`
* `neighbors_undirected()`
* `reverse()`
* `Entry` API

On `Graph`, there are the following duplicates:
* `node_indices()` -> `IntoNodeIdentifiers` (deprecate)
* `node_weight()` -> #556 `GraphAccess` (deprecate)
* `node_weight_mut()` -> #556 `GraphAccess` (deprecate)
* `edge_weight()` -> #556 `GraphAccess` (deprecate)
* `edge_weight_mut()` -> #556 `GraphAccess` (deprecate)
* `neighbors()` -> `IntoNeighbors` (deprecate)
* `neighbors_undirected()` -> `IntoNeighorsUndirected` (deprecate)

PR #562, opened 2023-05-18
Implement map_owned method

RosaUlisses:
This commit implement the map_owned method. It is equal to the map method, the unique difference is that map_owned require a `self` instead of `&self`.

Besides that I implemented tests for both functions (map method was not covered by tests).


indietyp:
Thank you for your PR; we're working on a significant version/rework (#551) that will likely change much of the underlying implementation. I plan to get back to this PR once things have stabilized.

One of the plans is to rework the traits and, with them, also add generic methods to map/filter (#559). This PR might be affected by this issue. This use case is something I will consider moving forward; it'd be great if you could outline your use case and why the current `map` is insufficient in #559. Thank you!

Some notes, nevertheless:

* Please remove the `.idea` folder
* you still have a `main.rs` file included

RosaUlisses:
I want to contribute in this library, what could I do to help you in this important rework of the library ?

indietyp:
Sure! Right now, quite a lot is in flux; in the future, some work might open up once the fundamentals are settled, but we're still in the conceptual stage, and the bandwidth of maintainers is currently quite limited. I'd feel uncomfortable assigning you a task of something I am still unsure about, or that might get scrapped later.

You could help us by looking at the issues linked in #551; it would be fascinating to know what people think about the proposals, what they feel is missing from the current version (or from the suggestions), and what you believe are opportunities for improvement for petgraph.

We're tracking progress in the `next` branch.

PR #567, opened 2023-06-20
0.7.0

indietyp:
This is the tracking branch for all changes for version `0.7.0`.

PRs for 0.7.0 will merge into this branch (`next/`); once ready, this will then be merged into `master`

PR #568, opened 2023-06-24
Update `indexmap` to v2

AlexTMjugador:
`indexmap` v2.0.0 has just been released, and several prominent crates in the Rust ecosystem such as `serde_json` are already using it. To minimize the number of projects for which Cargo would have to download and build two versions of `indexmap` due to their transitive dependencies, let's update the version `petgraph` depends on.

`indexmap` v2.0.0 does not bring substantial changes for this project, but it bumps our MSRV to 1.64. This should not be a concern, given that 0.7.0 (#567) will bump it to 1.65 anyway.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - rust: 1.41.0  # MSRV
-            # Fix an old minor version of some dependencies to ensure they
-            # compile with the MSRV
-            fixed_deps: -p indexmap --precise 1.7.0
+          - rust: 1.64.0  # MSRV
           - rust: stable
             features: unstable quickcheck
             test_all: --all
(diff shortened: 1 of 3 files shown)


indietyp:
MSRV upgrades are not considered breaking changes, so I see two ways forward:

1) we merge into `next/` and wait until 0.7.0 (which will take a while)
2) issue a new point release upgrading MSRV and indexmap.

AlexTMjugador:
For what it's worth, way 2 sounds better to me, as it avoids the described fragmentation and the inconveniences it might cause to downstream users.

XVilka (approved):
Agree on 2.

cwfitzgerald:
Getting this out would be great, is currently causing dupes in downstream libs!

indietyp:
I will try to get out a new point release this weekend!

cuviper:
@indietyp gentle ping, it would be nice to see that release!

indietyp:
Release tracked in https://github.com/petgraph/petgraph/pull/579, let's hope it works 🤞

PR #569, opened 2023-07-08
make last line also left-aligned

CobaltCause:
~~Not super confident this is the best way to have implemented this but it definitely works.~~ Reworked it to not allocate, which was my main concern.

Old behavior:

image

New behavior:

image

### DIFF
--- a/src/dot.rs
+++ b/src/dot.rs
@@ -248,6 +248,66 @@ where
     }
 }
 
+/// Detects whether the input has multiple lines
+struct MultilineDetector<W> {
+    writer: W,
+    saw_backslash: bool,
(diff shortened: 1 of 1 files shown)


indietyp:
0.7.0 plans to replace the built-in graphviz implementation with an external crate through a new `petgraph-io` crate.

CobaltCause:
Okay. Is there any reason not to merge this in the meantime though? I just noticed that a month after I opened this PR there was [another 0.6 release](https://github.com/petgraph/petgraph/commit/c78b5b13a816f1f0daf37f3964db0d33c6cbf38d)...

PR #571, opened 2023-07-22
Use `IndexMap::get_index_of`

cuviper:
Several calls to `IndexMap::get_full` were only using the index, which
can be more directly served by calling `get_index_of`.


bluss:
Thanks!

Issue #572, opened 2023-07-25
Rayonification: exposing parallel iterators in `GraphMap`.


daniel-levin:
### Summary
Long-time listener, first-time caller here. I am a happy user of `GraphMap`, as it is proving invaluable in a project I'm working on. However, it's a bit unwieldy to apply concurrent algorithms to it. Petgraph would be even better if it had easy to use parallel algorithms. Adding Rayon-based parallel iterators to petgraph's public API will let users take advantage of multi-core systems. **I'm happy to do this myself and submit PRs, but I'd like to gauge the community's interest first**.

### Motivation
The motivation is improving performance of `GraphMap` on multi-core systems. Under the hood, `GraphMap` uses indexmap, which [already exposes parallel iterators](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html#method.par_values), behind a feature flag, and that feature flag causes indexmap to take a dependency on rayon.

### Details

* What code needs to change? New traits? Which modules?
  * A dependency on Rayon must be added.
    * Behind a feature flag, just like indexmap, so that users can opt in and don't require a dependency on rayon.
  * The public API of `GraphMap` should expose parallel iterators over nodes and edges.
    *  by either hoisting indexmap's parallel iterators or through some layer of indirection that prevents the need to re-export indexmap's types.

* Is this a new graph algorithm? Include links to papers, Wikipedia, or other
  implementations of this algorithm that exist.
  * No.

* Are you willing to implement this yourself? Mentor someone else and help them
  implement it?
  * Yes.

dmyyy:
Interested in seeing something like this in petgraph

PR #573, opened 2023-07-27
Add Rayon-based parallel iterator over GraphMap nodes

daniel-levin:
See #572 for some context. In short, I have a use case for parallel iteration over the nodes in the underlying graph. I don't want every user of this crate to have to include Rayon, so I added it as a conditional dependency, as it is in indexmap. I wanted to create this PR early, to get an idea of the desirability of such a change - if this is something the petgraph team wants, I'll continue adding more.

The purpose of most of the code I've written so far is to expose a parallel iterator API that's doesn't re-export any types from indexmap and doesn't use any crate-private types.

I've also added a microbenchmark replicating my real-world workload.

I'm running the benchmarks on my laptop's i7-7500U, and even though [benchmarking on a laptop is a sin](https://lemire.me/blog/my-sayings/), the difference between serial and parallel is striking:

```
running 2 tests
test graphmap_parallel_bench ... bench:     707,148 ns/iter (+/- 325,762)
test graphmap_serial_bench   ... bench:   1,635,905 ns/iter (+/- 379,030)
```

Edit: ran the benchmarks on a m6g.2xlarge on EC2:
```
running 2 tests
test graphmap_parallel_bench ... bench:     422,702 ns/iter (+/- 4,520)
test graphmap_serial_bench   ... bench:   2,998,678 ns/iter (+/- 7,824)
```

Edit: I've edited the benchmarks to do something 'real', which still produces considerable speedups:

```
running 2 tests
test graphmap_parallel_bench ... bench:  35,571,906 ns/iter (+/- 2,379,332)
test graphmap_serial_bench   ... bench: 285,738,589 ns/iter (+/- 23,645,030)
```


src/rayon/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+/// Adds parallel iterators where possible, i.e. where nodes are [Send] + [Sync].
+pub trait NodesParIter<'a, N>
+where
+    N: Send + Sync,
+{
+    /// Returns a parallel iterator over this graph's nodes.
+    fn par_nodes(&'a self) -> ParNodes<'a, N>;
+}
+
+impl<'a, N, E, Ty> NodesParIter<'a, N> for GraphMap<N, E, Ty>
```
cuviper, line 36: Why use an extension trait for this? An inherent method on `GraphMap` should work fine, even defined from here in this module.

daniel-levin: I didn't know Rust lets you do that 😅

daniel-levin:
Good callout @cuviper ! I was not aware that you could define inherent methods in different modules, but now that I know that, I agree it's the right thing to do here. I've updated the code to do this.

Cargo.toml
```diff
@@ -38,10 +38,11 @@ debug = true
 
 [dependencies]
 fixedbitset = { version = "0.4.0", default-features = false }
-indexmap = "2.0"
+indexmap = { version = "2.0", default-features = true }
```
cuviper, line 41: What's the point of this? Defaults are on by default. :wink:

Cargo.toml
```diff
@@ -51,9 +52,10 @@ odds = { version = "0.4.0" }
 rand = "0.5.5"
 
 [features]
+rayon = ["graphmap", "dep:rayon", "indexmap/rayon"]
```
cuviper, line 55: I don't think this should force `graphmap` on -- that happens to be the only useful place for rayon functionality _right now_, but there could be others later.

src/graphmap.rs
```diff
@@ -59,7 +59,7 @@ pub type DiGraphMap<N, E> = GraphMap<N, E, Directed>;
 /// Depends on crate feature `graphmap` (default).
 #[derive(Clone)]
 pub struct GraphMap<N, E, Ty> {
-    nodes: IndexMap<N, Vec<(N, CompactDirection)>>,
+    pub(crate) nodes: IndexMap<N, Vec<(N, CompactDirection)>>,
```
cuviper, line 62: If these `pub(crate)` additions are not wanted, the `par_nodes` definition could be moved into this module or a submodule.

benches/graphmap.rs
```diff
@@ -0,0 +1,67 @@
...
+    gr
+}
+
+#[bench]
+fn graphmap_serial_bench(bench: &mut Bencher) {
+    let data = test_nodes();
+    let gr = test_graph(&data);
+    bench.iter(|| {
+        for n in gr.nodes() {
+            gr.edges_directed(n, Direction::Outgoing);
```
cuviper, line 53: This seems like a strange workload, creating an iterator without consuming it...

cuviper (commented):
I'm not a maintainer here, but since I was requested, I can add my 2 cents...

Cargo.toml
daniel-levin: D'oh 😣 chalk it up to inexperience with Cargo

benches/graphmap.rs
daniel-levin: Yeah this is peculiar. I'm going to consume the iterator for the sake of clarity of the benchmark. It's kinda subtle. The construction of the inner iterator (despite it not being consumed) is still time consuming enough to bottleneck the outer iterator - which happens to be a minimal reproducible example of my use case - and enough to manifest in the profiling data.

Cargo.toml
daniel-levin: Agreed.

daniel-levin:
@bluss @ABorgna Kindly requesting that you take a look.

Cargo.toml
ABorgna: Agreed with the suggestion,
```suggestion
indexmap = "2.0"
```

src/graphmap.rs
ABorgna: Yes, I say move `par_nodes` to here instead.

ABorgna (commented):
Nice! I don't have more to add to @cuviper's review; just move the code to `graphmap.rs`.

Cargo.toml
daniel-levin: Fixed this. At the time I misunderstood how feature flags worked.

src/graphmap.rs
daniel-levin: Done. I've removed the `rayon` module. Everything has been put into `graphmap.rs`, which on reflection, is quite a bit nicer than what I first proposed.

daniel-levin:
> Nice! I don't have more to add to @cuviper's review; just move the code to `graphmap.rs`.

Done! Thanks for the comments.

PR #575, opened 2023-07-30
Add edge_weight, edge_weight_mut, and edge_weight_safe methods for CSR graphs

Atila-M-Schrieber:
While other graph types have edge_weight methods, CSR graphs were lacking these features.
I have also created an edge_weight_safe method for safe accessing of edges between two nodes where you don't know if they are connected. 
I copied the doc comments from MatrixGraph, but I wasn't sure about what exactly was meant by the graph indexing syntax - I thought it was only for nodes.
Some methods which do the same things (has_edge vs contains_edge) are named differently, which might be something to check out.

### DIFF
--- a/src/csr.rs
+++ b/src/csr.rs
@@ -351,6 +351,19 @@ where
         &self.edges[p]
     }
 
+    /// Access the weight for edge `e`, mutably.
+    ///
+    /// Also available with indexing syntax: `&mut graph[e]`.
+    /// (NOTE TO MAINTAINERS - NOT SURE ABOUT THIS, NOT ENTIRELY SURE WHAT THIS MEANS)
+    ///
+    /// **Panics** if no edge exists between `a` and `b`.
+    pub fn edge_weight_mut(&mut self, a: NodeIndex<Ix>, b: NodeIndex<Ix>) -> &mut E {
+        let p = self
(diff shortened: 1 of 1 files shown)


indietyp:
0.7.0 completely reworks how graphs are managed through the `GraphStorage` and `DirectedGraphStorage` traits, these will ensure complete parity among implementations.

tinou98:
This feature would be interesting

But the implementation of `edge_weight_mut` in this PR is wrong for Undirected graphs, where we would need to update both values.

This might be a problem if you try to unify different graphs implementations, as such a version would need to return a proxy that would update both value when mutated.

But, a `set_edge_weight` version would be easier

```rust
    /// Update the weight for an existing edge between `a` and `b`.
    pub fn set_edge_weight(&mut self, a: NodeIndex<Ix>, b: NodeIndex<Ix>, weight: E) -> bool
    where
        E: Clone,
    {
        let pos = match self.find_edge_pos(a, b) {
            Ok(i) => i,
            Err(_) => return false, /* not found */
        };

        self.edges[pos] = weight.clone();
        if !self.is_directed() && a != b {
            let pos = self.find_edge_pos(b, a).expect("mirror edge to exist in Undirected graph");
            self.edges[pos] = weight;
        }
        true
    }
```

PR #576, opened 2023-08-10
Add element filter adapter

Phlosioneer:
Due to the implementation of several traits on
graph-refs, it is not possible to nest a
`NodeFiltered` within an `EdgeFiltered`, and
vice-versa. The inner graph must be a GraphRef,
which imposes lifetime constraints that can't be
resolved in many cases (e.g. as a field within
a struct, or as a return value from a function).

This PR adds the `ElementFiltered` adapter
address this problem, and a `NestedFilter` structure for the most common way to combine a node filter and an edge filter.

I can write more tests if needed, but it mostly reuses `NodeFiltered` and `EdgeFiltered` code. For example, I didn't need to define any new iterator structs. Testing deficiencies would probably apply for all the filters.

### DIFF
--- a/src/visit/filter.rs
+++ b/src/visit/filter.rs
@@ -581,3 +581,173 @@ NodeCount! {delegate_impl [[G, F], G, EdgeFiltered<G, F>, access0]}
 NodeIndexable! {delegate_impl [[G, F], G, EdgeFiltered<G, F>, access0]}
 EdgeIndexable! {delegate_impl [[G, F], G, EdgeFiltered<G, F>, access0]}
 Visitable! {delegate_impl [[G, F], G, EdgeFiltered<G, F>, access0]}
+
+/// Combines a node filter and an edge filter.
+///
+/// Whenever an edge is checked by this filter, it first checks if the
+/// source and target nodes should be included, and then it checks if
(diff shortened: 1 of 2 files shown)


indietyp:
The element filter has been replaced in `0.7.0` with a more traditional approach of using iterators and a `Graph::from_parts` method.

Phlosioneer:
I'm not sure I understand; do you mean that this use case will require cloning the subgraph?

indietyp:
No, in the new version (where you can check out the current docs by switching to `next` and generating the docs for `petgraph-core`) the graph has a method `fn into_parts(self) -> (impl Iterator<Item = DetachedNode>, impl iterator<Item = DetachedEdge>)`, and `fn from_parts(self, nodes: impl Iterator<Item = DetachedNode>, edges: impl Iterator<Item = DetachedEdge>)`.

You can them do the following:

```rust
let (nodes, edges) = graph.into_parts();

let nodes = nodes.filter(|node| node.weight % 2 == 0);

let graph = DiGraph::from_parts(nodes, edges);
```

Phlosioneer:
That makes sense. Thanks for clarifying.

PR #585, opened 2023-09-10
add Topo::with_initials API

cool-little-fish:
Sometimes the user may already know the root nodes, it's convenient to use `Topo::with_initials` to create a Topo instance directly.

### DIFF
--- a/src/visit/traversal.rs
+++ b/src/visit/traversal.rs
@@ -344,14 +344,19 @@ where
         topo
     }
 
-    /// Create a new `Topo` with initial nodes
+    /// Create a new `Topo` with initial nodes.
+    ///
+    /// If any incoming edges to the initial nodes will be ignored.
     pub fn with_initials<G, I>(graph: G, i: I) -> Self
     where
-        G: Visitable<NodeId = N, Map = VM>,
+        G: IntoNeighborsDirected + Visitable<NodeId = N, Map = VM>,
         I: IntoIterator<Item = N>,
     {
         Topo {
-            tovisit: i.into_iter().collect(),
+            tovisit: i
+                .into_iter()
+                .filter(|&n| graph.neighbors_directed(n, Incoming).next().is_none())
+                .collect(),
             ordered: graph.visit_map(),
         }
     }
--- a/src/visit/traversal.rs
+++ b/src/visit/traversal.rs
@@ -346,14 +346,14 @@ where
 
     /// Create a new `Topo` with initial nodes.
     ///
-    /// If any incoming edges to the initial nodes will be ignored.
-    pub fn with_initials<G, I>(graph: G, i: I) -> Self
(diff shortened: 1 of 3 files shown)


src/visit/traversal.rs
```diff
@@ -344,6 +344,18 @@ where
...
+    /// Create a new `Topo` with initial nodes
+    pub fn with_initials<G, I>(graph: G, i: I) -> Self
+    where
+        G: Visitable<NodeId = N, Map = VM>,
+        I: IntoIterator<Item = N>,
+    {
+        Topo {
+            tovisit: i.into_iter().collect(),
+            ordered: graph.visit_map(),
+        }
```
ABorgna, line 356: If the user passes a node with incoming edges, this would then yield duplicates.

I say mark the items in `i` in the `visit_map`. Then we'd process them instead as if the incoming edges didn't exist.

src/visit/traversal.rs
```diff
@@ -344,6 +344,18 @@ where
         topo
     }
 
+    /// Create a new `Topo` with initial nodes
```
ABorgna, line 347: Continuing with the other comment, document the behaviour here.
```suggestion
    /// Create a new `Topo` with initial nodes.
    ///
    /// If any incoming edges to the initial nodes will be ignored.
```

src/visit/traversal.rs
cool-little-fish: The items in `i` could be children of others nodes, then we may miss them if we just mark them in the `visit_map`. It's more reasonable to filter the nodes with incoming edges I think.

src/visit/traversal.rs
```diff
@@ -344,6 +344,23 @@ where
         topo
     }
 
+    /// Create a new `Topo` with initial nodes.
+    ///
+    /// If any incoming edges to the initial nodes will be ignored.
```
ABorgna, line 349:
```suggestion
    /// Nodes with incoming edges are ignored.
```

src/visit/traversal.rs
```diff
@@ -344,6 +344,23 @@ where
...
+    pub fn with_initials<G, I>(graph: G, i: I) -> Self
+    where
+        G: IntoNeighborsDirected + Visitable<NodeId = N, Map = VM>,
+        I: IntoIterator<Item = N>,
+    {
+        Topo {
+            tovisit: i
```
ABorgna, line 356: Slight rename to make the API clearer.
```suggestion
    pub fn with_initials<G, I>(graph: G, initials: I) -> Self
    where
        G: IntoNeighborsDirected + Visitable<NodeId = N, Map = VM>,
        I: IntoIterator<Item = N>,
    {
        Topo {
            tovisit: initials
```

ABorgna (approved):
That works too. Thanks!

I'll just add a couple doc changes.

PR #591, opened 2023-10-10
Supplement VF2 algorithm

Lucian-4a25:
This PR try to add more granular feasibility rule check which conform the paper of VF2 algorithm. In current VF2 algo implemention, `petgraph` only check the neighbors' sum of number which are added to outs or ins for new node pair, but according to VF2 's paper, we could check in a more granular way, and compare the possible new node pair's succ and prev neighbors which are in outs or ins or mapping or unknown nodes respectively.

And This PR also add one more benchmark test for isomorphism and modify one of integration test for isomorphism to fit the change of codes.

### DIFF
--- a/src/algo/isomorphism.rs
+++ b/src/algo/isomorphism.rs
@@ -30,14 +30,14 @@ mod state {
         /// out[i] is non-zero if i is in either M_0(s) or Tout_0(s)
         /// These are all the next vertices that are not mapped yet, but
         /// have an outgoing edge from the mapping.
-        out: Vec<usize>,
+        pub out: Vec<usize>,
         /// ins[i] is non-zero if i is in either M_0(s) or Tin_0(s)
         /// These are all the incoming vertices, those not mapped yet, but
         /// have an edge from them into the mapping.
         /// Unused if graph is undirected -- it's identical with out in that case.
-        ins: Vec<usize>,
-        pub out_size: usize,
-        pub ins_size: usize,
+        pub ins: Vec<usize>,
+        // pub out_size: usize,
+        // pub ins_size: usize,
         pub adjacency_matrix: G::AdjMatrix,
         generation: usize,
     }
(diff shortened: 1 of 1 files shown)


indietyp:
I am planning to completely re-implement the VF2 algorithm (in-fact replace it by VF2++) in an upcoming PR for 0.7.0

indietyp:
The codebase is currently still quite in flux and prone to breakage, but once more stable, any help would be very much appreciated!

Lucian-4a25:
The most difference between VF2++ and VF2 is that VF2++ compute a matching order before doing isomorphism check, they are basically same with the rest of part. In some cases, VF2 algo maybe more efficient  because it doesn't need to do a presort for nodes in G0, which is a time-consuming process. I think it's reasonable to keep them all in the future, give user to choose which algo is more fit for them.
I see that you want to rework the core trait of `petgraph`. That's cool, currently many definiton  of trait is not so reasonable. For example, trait  `IntoNeighbors` or `IntoNeighborsDirected` should have  definiton of borrowed Self as first parameter, instead of a consuming Self,  which I think it's design problem.

Lucian-4a25:
> Once VF2 is ported (which should be soon-ish), would you be willing to integrate your work into 0.7.0 instead? (although some modifications might be necessary)
> 
> I think it would also be helpful to have documentation stating when to use which algorithm in the future (something I plan for every category where we implement multiple, like #594)

Sure, I'm willing to help with it when 0.7 version is landed. I learn much from `petgraph`, it's a good chance for me to give back the help I get.

bluss:
> For example, trait IntoNeighbors or IntoNeighborsDirected should have definiton of borrowed Self as first parameter, instead of a consuming Self, which I think it's design problem.

The design is that Self is a reference to a graph - so the function call "consumes" a reference. It was designed for early rust, and the only way (?) to have iterators in the traits. Now we have better tools with GAT for traits.

OwenTrokeBillard:
I implemented VF2 for petgraph graphs in another crate: https://github.com/OwenTrokeBillard/vf2. It supports induced and non-induced subgraphs.

skius:
I have also become interested in the VF2++ algorithm. For reference, this seems to be a useful resource from someone who added VF2++ to networkx: https://gist.github.com/kpetridis24/535af6baf74953e55edc9770bea75200

PR #596, opened 2023-10-22
[Bug] Walker bug after `remove_edge`

yanivpas:
Hi,

When using `NeighborsWalker` to remove all edges for a given node, sometimes not all the edges are removed.
this happens because the walker keeps index to the next edge, which might be swapped by the remove_edge.

Code example to reproduce the error:
```rust
 #[test]
  fn test_detach_and_remove_edge() {
      let mut graph: Graph<(), ()> = Graph::new();
      let nodes: Vec<_> = (0..2u32).map(|_| graph.add_node(weight: ())).collect();
      let src_node: NodeIndex = graph.add_node(weight: ());
      let edge_to_remove: EdgeIndex = graph.add_edge(a: nodes[0], b: nodes[1], weight: ());
      graph.add_edge(a: src_node, b: nodes[0], weight: ());
      graph.add_edge(a: src_node, b: nodes[1], weight: ());
      graph.remove_edge(edge_to_remove);

      let mut walker: WalkNeighbors<u32> = graph.neighbors_directed(a: src_node, dir: petgraph::Direction::Outgoing).detach();
      while let Some(edge: EdgeIndex) = walker.next_edge(&graph) {
          graph.remove_edge(edge);
      }
      assert_eq!(graph.neighbors(src_node).count(), 0);
  }
```

The proposed solution is to create a new API for the walker to remove the edge, which will check if the next edge is the swapped one.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -725,19 +725,12 @@ where
         }
     }
 
-    /// Remove an edge and return its edge weight, or `None` if it didn't exist.
-    ///
-    /// Apart from `e`, this invalidates the last edge index in the graph
-    /// (that edge will adopt the removed edge index).
-    ///
-    /// Computes in **O(e')** time, where **e'** is the size of four particular edge lists, for
-    /// the vertices of `e` and the vertices of another affected edge.
-    pub fn remove_edge(&mut self, e: EdgeIndex<Ix>) -> Option<E> {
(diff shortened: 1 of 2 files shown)


indietyp:
Because of errors like these (and associated problems) 0.7.0 deprecates the walker trait and implementation in favor of a more conventional `drain` and `iterator` approach.

indietyp:
0.7.0 is still quite unstable (with the core now hopefully being stable) and is tracked here: https://github.com/petgraph/petgraph/pull/567. There is currently no chat/slack/IRC, although I have been thinking about opening a petgraph Discord for discussion if there's want.

PR #601, opened 2023-12-08
Replace `copyclone!` macro by `Clone` derivation

samueltardieu:
`Clone` can be derived at the same time as `Copy`: Rust will generate an implementation which does a bitwise copy of the value.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -170,20 +170,9 @@ pub mod graph {
 #[cfg(feature = "stable_graph")]
 pub use crate::graph_impl::stable_graph;
 
-macro_rules! copyclone {
-    ($name:ident) => {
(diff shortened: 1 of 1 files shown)


samueltardieu:
An [upcoming Clippy lint](https://github.com/rust-lang/rust-clippy/pull/11878) would tag the current `copyclone!()` macro as in some cases it attempts to dereference an reference to an uninhabited type (such as a `Directed` or `Undirected` instance). Although only unsafe code could produce such a reference, Clippy will flag the code as suspicious as dereferencing an uninhabited type is undefined behaviour.

samueltardieu:
Pinging @indietyp for review (as you seem to be the most active here nowadays).

PR #603, opened 2023-12-29
Improve overall speed.

indietyp:
This introduces the fixes that have been applied in `bm/dino-closure` (and more) to improve the speed of algorithms and lookup tremendously.


indietyp:
Using dimacs9 as an example, the dijkstra runs on 0.6.0 with the following speed:

```
dimacs9/florida/dijkstra
                        time:   [217.47 ms 220.04 ms 222.66 ms]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

indietyp:
The examples are not working; I think I found a way to significantly reduce the burden of implementations and bubbling of associated bounds associated with ids. Thanks to @TimDiekmann for the advice

PR #606, opened 2024-01-07
Stoer–Wagner algorithm for minimum cut

gbagan:
Hi,

I propose an implementation of Stoer–Wagner algorithm to solve the (global) minimum cut problem.
https://en.wikipedia.org/wiki/Stoer%E2%80%93Wagner_algorithm

The implementation runs in O(|V| |E| + |V|²*log(|V|).

I'm fairly new to Rust and Petgraph. So I would be happy to have some advice.

The algorithm requires modifying the graph (vertex merging). So, I works on a mutable copy of the graph with the type Graph. 
There may be a more elegant way to do this.

Since I need a max version of `MinScored`, I created the struct `MaxScored`.

I have added tests, quickcheck tests and benchmarks.

### DIFF
--- a/src/algo/minimum_cut.rs
+++ b/src/algo/minimum_cut.rs
@@ -148,10 +148,10 @@ where
     let len = seen_list.len();
     let last = seen_list[len-1];
     let before_last = seen_list[len-2];
-    let edges: Vec<_> = graph.edges(last).map(|e| e.weight().0).collect();
-    let weight = sum(graph.edges(last).map(|e| e.weight().1));
+    let cut: Vec<_> = graph.edges(last).map(|e| e.weight().0).collect();
+    let cut_weight = sum(graph.edges(last).map(|e| e.weight().1));
     
-    (edges, weight, last, before_last)
+    (cut, cut_weight, last, before_last)
 }
 
 fn minimum_cut_aux<N, EdgeId, K, Ix>(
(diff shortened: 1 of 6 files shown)


aalekhpatel07:
Thanks for this! Works like a charm. This definitely [helped](https://github.com/aalekhpatel07/advent-of-code-2023/blob/main/src/bin/day-25.rs) me with an Advent of Code problem, haha.

src/algo/minimum_cut.rs
```diff
@@ -0,0 +1,186 @@
...
+    graph: G,
+    edge_cost: F,
+    ) -> (Vec<G::EdgeId>, K)
+where
+    G: IntoNodeIdentifiers + IntoEdgeReferences,
+    F: Fn(G::EdgeRef) -> K,
+    G::NodeId: Eq + Hash,
+    K: Measure + Copy,
+{
+    let mut node_to_node = HashMap::new();
```
RaoulLuque, line 71: std::HashMap should be replaced by the hashbrown equivalent to be no-std compliant.

RaoulLuque (commented):
A rebase to the current version of Petgraph is probably necessary. In particular, to comply with no-std, HashMap from the standard library should/could be replaced with the one from HashBrown and similarly with BinaryHeap.

I would also like if the private helper functions had some brief docstrings like for example in https://github.com/petgraph/petgraph/blob/master/src/algo/ford_fulkerson.rs

RaoulLuque:
Hey, I wanted to check-in on whether you'd be willing to continue working on this PR, or whether I or someone else could possibly pick it up in the future :) ?

PR #605, opened 2024-01-03
Simplify trait definition

indietyp:
Follow up on #587 and discussions following #603.

1. `NodeId` and `EdgeId` can no longer be chosen and are just wrappers around usize (now in core)
2. Proposal (2) has been implemented, with `GraphMap` now renamed `EntryGraph` to reflect the changes.
3. Auxiliarly graph storage and sequential mapping of ids are no longer optional
4. Auxiliarly graph storage now accepts performance hints (currently unused
5. `ReverseGraphStorage` as helper for `EntryGraph` to facilitate lookup by key

The following hierarchy has been implemented:

```mermaid
graph TD;
	AuxiliarlyGraphStorage ---> GraphStorage
	SequentialGraphStorage ---> GraphStorage

	GraphStorage ---> ReverseGraphStorage
	GraphStorage ---> DirectedGraphStorage
	GraphStorage ---> RetainableGraphStorage
```

## Next Steps

* testing framework for graph implementation
* implementations of more algorithms on top of the new paradigm (and further performance improvements)
* unify `csr`, `matrix-graph`, `adjacency-matrix` into a single crate
* delete old `graph`, replace with `dino`
* `DinoGraph` make use of `Hints`
* rework `from_parts` and `into_parts` with a mapping and stable ids if ingesting from same storage (how to "tag" ids?)
* reworked serde support
* documentation, documentation, documentation
* potentially widen id to 16 bytes instead of 8 to include a VTable with tag and `Display` implementation (don't know if it is worth it tho)

PR #616, opened 2024-02-04
Use typos-cli for automatic spell-checking

cnpryer:
Hi! I've been reading through both `master` and `next`. Love what you're doing @indietyp!

I noticed in #605 you mentioned *documentation, documentation, documentation* as part of next steps. I figured it'd be helpful to add some tooling ([`typos`](https://github.com/crate-ci/typos)) to catch typos throughout both the code and, maybe more importantly, in any documentation that's generated.

I also ran `cargo fmt` and manually opted-out of any code-specific formatting in the diff. I did this because I noticed you're using some non-defaults for formatting options like `wrap_comments`.

I'm also happy to checkout upgrading [`actions/checkout@v4`](https://github.com/actions/checkout) if you're interested.

_____

On a side note, if my goal is to get familiar with this code-base, what do you think makes the most sense for me to spend more of my time with `master` or `next`?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,5 +131,13 @@ jobs:
 
       - name: Run tests
         run: just coverage --lcov --output-path lcov.info
+  typos:
+      name: "Spell check"
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v3
+        - uses: crate-ci/typos@master
+          with:
+            files: .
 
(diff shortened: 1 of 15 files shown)


indietyp:
Hello, thank you! I am currently in the middle of exam preparations, so the little free time I have isn't available for petgraph (which will hopefully be different in the near future). The changes look good. about the `rust-fmt` thing, we have a `rustfmt.toml` configuration file, that you be automatically picked up when you run `rustfmt`!

PR #617, opened 2024-02-26
Update to fixedbitset 0.5.0

james7132:
Finally got around to fixing the issues with fixedbitset. This PR updates fixedbitset to 0.5.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ name = "petgraph"
 debug = true
 
 [dependencies]
-fixedbitset = { version = "0.4.0", default-features = false }
+fixedbitset = { version = "0.5.0", default-features = false }
 indexmap = "2.0"
 quickcheck = { optional = true, version = "0.8", default-features = false }
 serde = { version = "1.0", optional = true }


bluss:
nice. breaking by being a public dependency so care needed, needs version bump in petgraph too, I think?

james7132:
If that's the case, should I remove those reexports? There may be more breaking changes in the near future with the SIMD acceleration work that has yet to land.

bluss:
Not a reexport but a trait impl, I don't see how it can be removed right now.

Maybe https://docs.rs/petgraph/0.6.4/petgraph/visit/trait.Visitable.html  with the associated type that links Graph to Graph::Map == FixedBitSet is the more clear example. I.e. that means FixedBitSet is in Graph's (an others) public API.

james7132:
Note the SIMD changes made it in without another minor version bump. This should be the only breaking change.

bluss:
Here is info, just for clarification:

- Fixedbitset is a public dependency
  - meaning: it's in the public api of petgraph
- 0.4.* to 0.5.* is not a default automatic upgrade in cargo

Those facts are enough to say this PR is a breaking change in petgraph, the contents of the library or the change in it doesn't matter.

james7132:
Right I was just mentioning that this is likely to be the only breaking change from updating FixedBitset, as the major changes have landed without needing to bump the minor version further.

JMS55:
Any update on this?

XVilka:
Thanks! Will be a part of the petgraph-0.6.6

PR #622, opened 2024-03-22
Allow alternative hash functions in GraphMap

daniel-levin:
Hi there @bluss and @ABorgna!

Today, `GraphMap` uses `IndexMap`'s default choice of hash function. Like many Rustaceans, I have a use case that is bottlenecked by hashing, which could take advantage of a faster hash function.

This change allows users of this crate to pick alternative hashers, as they might with an `IndexMap` or a `HashMap`, but only if they want to. Existing code will still work, because each `impl` has been made more generic, since `std::hash::RandomState` already implements `BuildHasher` and `Default`.

I took `dev-dependencies` on `ahash` and `fxhash`, and extended a benchmark I'd previously contributed:

```
running 3 tests
test graphmap_serial_bench        ... bench: 208,458,239 ns/iter (+/- 6,845,112)
test graphmap_serial_bench_ahash  ... bench: 171,139,041 ns/iter (+/- 3,972,136)
test graphmap_serial_bench_fxhash ... bench: 183,217,281 ns/iter (+/- 55,096,481)
```

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -5,11 +5,12 @@ use indexmap::map::Keys;
 use indexmap::map::{Iter as IndexMapIter, IterMut as IndexMapIterMut};
 use indexmap::IndexMap;
 use std::cmp::Ordering;
+use std::collections::hash_map::RandomState;
 use std::collections::HashSet;
 use std::fmt;
-use std::hash::{self, Hash};
+use std::hash::{self, BuildHasher, Hash};
+use std::iter::Copied;
 use std::iter::FromIterator;
-use std::iter::{Copied, DoubleEndedIterator};
 use std::marker::PhantomData;
 use std::mem;
 use std::ops::{Deref, Index, IndexMut};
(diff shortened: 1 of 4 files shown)


ABorgna (approved):
Awesome. Thanks!

src/graphmap.rs
```diff
@@ -166,21 +174,40 @@ where
...
     N: NodeTrait,
     Ty: EdgeType,
+    S: BuildHasher,
 {
     /// Create a new `GraphMap`
-    pub fn new() -> Self {
+    pub fn new() -> Self
+    where
+        S: Default,
+    {
```
cuviper, line 187: FYI: by leaving `S` open here, it means that type inference can't figure it out, and it won't use the default either. e.g. This used to work:

```rust
let mut graph = GraphMap::new();
// ... other code that insert stuff, inferring those types but *not* the hasher
```

Now to allow inference I have to write:

```rust
let mut graph = GraphMap::<_, _, _>::new(); // now 4th param `S` is defaulted
```

That's why `HashMap::new()` and `IndexMap::new()` only use the default hasher, while `default()` is fully generic.
See also: https://faultlore.com/blah/defaults-affect-inference/

cuviper: Don't get me wrong, this is a great feature overall! I would have just left `new()` on the old hasher for the reason above, but it would be a breaking change to go back on that now.

daniel-levin: Ack, thanks for the comment. I would have done it differently if I'd realized.

PR #624, opened 2024-04-01
Move Minimum Spanning Tree Algorithm to its own module

BryanCruz:
### Move Minimum Spanning Tree Algorithm to its own file

This will better organize stuff related to this algorithm, while also providing room to add other algorithms in addition to Kruskal's, such as Prim's (https://en.wikipedia.org/wiki/Prim%27s_algorithm) for example. 

This PR should not change any behavior, only move stuff around.

### DIFF
--- a/benches/min_spanning_tree.rs
+++ b/benches/min_spanning_tree.rs
@@ -1,0 +1,60 @@
+#![feature(test)]
+
+extern crate petgraph;
+extern crate test;
+
+use test::Bencher;
+
+#[allow(dead_code)]
+mod common;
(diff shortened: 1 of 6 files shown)


ABorgna (approved):
Hi, thanks for the PR.

The next mayor release of the crate will re-organize the algorithms module into their own crate (see #551, and the feature branch `next`), so this change may be short-lived.
But I'm always happy to merge changes that make the code cleaner. We'll eventually move any new algorithms to their new location for `0.7.0`.

PR #625, opened 2024-04-02
Add Prim's Algorithm for Minimum Spanning Tree

BryanCruz:
Prim's algorithm: https://en.wikipedia.org/wiki/Prim%27s_algorithm

This PR adds another algorithm to find the Minimum Spanning Tree of a graph, an alternative to Kruskal's.

1. Prim's algorithm has some limitations, such as the input graph must be undirected and it should not have disconnected components.
I'm wondering if this can be enforced with traits, if anyone can help on this.

2. I would like to suggest adding another algorithm as well, `min_spanning_forest_prim`, that iterates through this algorithm to find min spanning tree for all input graph components. `min_spanning_forest_kruskal` would be trivial to implement, since current `min_spanning_tree` function, using Kruskal's algorithm, already finds a min spanning forest. Any thoughts on this? 

This PR also aims to fix current `min_spanning_tree` benches, since they currently do not iterate through the created `MinSpanningTree`  structure, not giving the actual bench for Kruskal algorithm.

3. I added a simple function to iterate through the generated tree elements, but I'm not very familiar with Rust ecosystem, so I would appreciate if anyone knows a better way to force the iteration:
```Rust
// Current Bench:
bench.iter(|| (min_spanning_tree(&a), min_spanning_tree(&b)));

// Suggested Bench:
bench.iter(|| (iterate_mst_kruskal(&a), iterate_mst_kruskal(&b)));

// Force Tree Iteration:
fn iterate_mst_kruskal<G>(g: G) -> bool
where
    G: Data + IntoEdges + IntoNodeReferences + IntoEdgeReferences + NodeIndexable,
    G::NodeWeight: Clone,
    G::EdgeWeight: Clone + PartialOrd,
{
    let mst = min_spanning_tree(g);
    mst.into_iter().all(|_| true)
}
```


ABorgna:
Thanks for the PR!

1. Kruskal's doc states 
  ```The input graph is treated as if undirected.```
  perhaps we could do the same here
2. That could be useful.
    It is a bit inconsistent that Kruscal always returns a forest, but leaving it as undefined behaviour if there is more than one component sounds OK.

benches/min_spanning_tree.rs
```diff
@@ -9,52 +9,100 @@ use test::Bencher;
...
+}
+
+fn iterate_mst_prim<G>(g: G) -> bool
+where
+    G: Data + IntoEdges + IntoNodeReferences + IntoEdgeReferences + NodeIndexable,
+    G::NodeWeight: Clone,
+    G::EdgeWeight: Clone + PartialOrd,
+{
+    let mst = min_spanning_tree_prim(g);
+    mst.into_iter().all(|_| true)
```
ABorgna, line 107: Good catch. Using `black_box` is the best bet to avoid it getting optimised away.

```suggestion
fn iterate_mst_kruskal<G>(g: G)
where
    G: Data + IntoEdges + IntoNodeReferences + IntoEdgeReferences + NodeIndexable,
    G::NodeWeight: Clone,
    G::EdgeWeight: Clone + PartialOrd,
{
    for e in min_spanning_tree(g) {
        std::hint::black_box(e);
    }
}

fn iterate_mst_prim<G>(g: G)
where
    G: Data + IntoEdges + IntoNodeReferences + IntoEdgeReferences + NodeIndexable,
    G::NodeWeight: Clone,
    G::EdgeWeight: Clone + PartialOrd,
{
    for e in min_spanning_tree_prim(g) {
        std::hint::black_box(e);
    }
```

BryanCruz:
> 1. Kruskal's doc states
>    `The input graph is treated as if undirected.`
>    perhaps we could do the same here

Thanks for the review @ABorgna :) 

I updated the docs to let more explicit what will happen if input graph is directed or has more than 1 component, and applied the suggested change for benches.

I think we would need to convert a directed graph to an undirected one in order to Prim behave correctly on it. For example, a graph where edges are: (I) A-5->B , (II) B-10->C and (III) C-15->A,  the valid MST computed by Prim would be {I and II} (undirected), but in order to actually implement this behavior we would need some sort of inverse of `IntoEdgeReferences`, to access edges that goes to a certain node

So I don't think it's worth the complexity, since there are other algorithms specifically for directed graphs that could be implemented.

What do you think?

daehiff:
Hey, since we can use this feature in a different project: 
Is there any due date to this feature or any open points that need to addressed, if yes, do you need help with that?

BryanCruz:
Thanks for the interest in this feature @daehiff 😄 

I cleaned up this PR, it had some extra stuff from another PR I was working on.

I don't think there are any open points to be addressed, just waiting for a review now.

BryanCruz:
Limited `indexmap` version in project config since its latest release broke `petgraph` MSRV

ABorgna (approved):
Thanks!

Perhaps we should rename Kruskal to `min_spanning_tree_kruskal` to be more consistent, but that should be done later in a breaking release.

PR #627, opened 2024-04-11
Fixed clippy warnings

martinfrances107:
I ran cargo clippy and fixes most but not all of the warnings.

I manualy reviewed all the changes ... they all look like  improvements.

Nothing earth shaking lots of tiny blemishes removed.

By was of testing, I can asset that 
```
cargo test feature --all
```

passes.

When I say most but not all clippy warning are cleared. A  warning of this form remains 

```warning: docs for unsafe trait missing `# Safety` section```

but I think documentation issues are best dealt with in a separate issue 

Highlights :-

1) .any() is prefered over .find(), as it will break a "loop" on the first occurance.
```
-                        if let Some(_) = w_out_edges.find(|e| e.target() == nodeix(v)) {
+                        if w_out_edges.any(|e| e.target() == nodeix(v)) {
```

2) Was manually implementing a call to .find()

```
-        while let Some(edge) = self.edges.next() {
-            if edge.node[1] == self.target_node {
-                return Some(edge);
-            }
-        }
-
-        None
+      let target_node = self.target_node;
+      self.edges.by_ref().find(|&edge| edge.node[1] == target_node)
```

3) Made find_join() make more flexible by accepting slices rather
   than a strict vec.

```
-    label: &mut Vec<Label<G>>,
-    first_inner: &mut Vec<usize>,
+    label: &mut [Label<G>],
+    first_inner: &mut [usize],
```

4) Now using flap_map where possible
```
-        .map(|(i, &node)| {
+        .flat_map(|(i, &node)| {
```

5) using .by_ref() to repalce a while loop with something
   more idomatic.

```
-        while let Some(next) = self.iter.next() {
+        for next in self.iter.by_ref() {
```

6) removed the use of a deprecated function.

```
-        let weight = self.edges.remove(&Self::edge_key(a, b));
+        let weight = self.edges.swap_remove(&Self::edge_key(a, b));
```

### DIFF
--- a/src/algo/isomorphism.rs
+++ b/src/algo/isomorphism.rs
@@ -766,7 +766,8 @@ mod matching {
 
             // We hardcode n! values into an array that accounts for architectures
             // with smaller usizes to get our upper bound.
-            let upper_bounds: Vec<Option<usize>> = [1u64,
+            let upper_bounds: Vec<Option<usize>> = [
+                1u64,
                 1,
                 2,
                 6,
@@ -786,7 +787,8 @@ mod matching {
                 355687428096000,
                 6402373705728000,
                 121645100408832000,
-                2432902008176640000]
+                2432902008176640000,
+            ]
             .iter()
             .map(|n| usize::try_from(*n).ok())
             .collect();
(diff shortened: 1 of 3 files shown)


ABorgna (approved):
Thanks!

PR #630, opened 2024-04-11
Minor: Cargo doc fixup.

martinfrances107:
Clippy suggests using this markdown convention when discussing unsafe code.

A new heading and section is added.

This is how the updated page looks 

Safe


ABorgna (approved):
👍

PR #636, opened 2024-04-21
Add iterator over incident edges of a stable graph node.

ISibboI:


src/graph_impl/stable_graph/mod.rs
```diff
@@ -1457,6 +1473,127 @@ where
...
+                direction: Outgoing,
+            });
+        }
+
+        while let Some(Edge { node, weight, next }) = self.edges.get(self.next[1].index()) {
+            debug_assert!(weight.is_some());
+            let edge_index = self.next[1];
+            self.next[1] = next[1];
+            // In any of the "both" situations, self-loops would be iterated over twice.
+            // Skip them here.
```
ABorgna, line 1580:
```suggestion
            // Self-loops would be iterated over twice. Skip them here.
```

ABorgna (changes requested):
Thanks! This is quite useful.

Would you mind implementing the method on `Graph` too?

PR #640, opened 2024-04-26
Ford Fulkerson algorithm support.

geosarr:


starovoid:
Hi @geosarr, do I understand correctly that the classic Edmonds–Karp algorithm is implemented here?

geosarr:
Hi @starovoid, I am not familiar with the name Edmonds-Karp, but looking on the internet it seems like it. In the implemented algorithm, breadth first seach is used to find an augmented path, if any exists.

PR #648, opened 2024-06-14
fix adjacency matrix for csr and adjacency list

BryanCruz:
hi folks, I was using the `GetAdjacencyMatrix` trait in a personal project and found out that the current trait implementations in `CSR` and `List` weren't working as expected, so I'm raising this PR to address that.

I also added unit tests for graph types where this trait is implemented.

Problems being fixed:
* `List`: This data structure is directed, but the adjacency matrix was being built as if it was undirected, adding both `a->b` and `b->a` edges to the matrix if even if only one of them was present in the graph. Also, it was using `node_count` to index the adjacency matrix rows, but was using `edge_count` to query it, resulting in wrong results.
* `CSR`: Was not accounting for undirected graphs

### DIFF
--- a/src/adj.rs
+++ b/src/adj.rs
@@ -641,15 +641,12 @@ where
         for edge in self.edge_references() {
             let i = edge.source().index() * n + edge.target().index();
             matrix.put(i);
-
-            let j = edge.source().index() + n * edge.target().index();
-            matrix.put(j);
         }
         matrix
     }
 
     fn is_adjacent(&self, matrix: &FixedBitSet, a: NodeIndex<Ix>, b: NodeIndex<Ix>) -> bool {
-        let n = self.edge_count();
+        let n = self.node_count();
         let index = n * a.index() + b.index();
         matrix.contains(index)
     }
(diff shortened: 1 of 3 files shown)


XVilka (changes requested):
Thanks for the PR! Could you please rebase it? Also please update the PR description to describe what *exactly* was not working properly and is now fixed.

BryanCruz:
> Thanks for the PR! Could you please rebase it? Also please update the PR description to describe what _exactly_ was not working properly and is now fixed.

thanks for the review, made the requested changes

XVilka (approved):

XVilka:
@starovoid please take a look too

starovoid:
@XVilka 
It's definitely worth doing a new rebase. It also seems to me correct to split all the test functions into two parts, because the code below covers two non-trivial cases.

XVilka:
> @XVilka It's definitely worth doing a new rebase. It also seems to me correct to split all the test functions into two parts, because the code below covers two non-trivial cases.
> 
> ```rust
> test_adjacency_matrix(&directed_g);
> test_adjacency_matrix(&undirected_g);
> ```

Good idea! @BryanCruz please address this.

BryanCruz:
@XVilka @starovoid done

XVilka (approved):

PR #649, opened 2024-06-15
implement `Neighbors` via `WalkNeighbors` and add an `attach` method to `WalkNeighbors`

soooch:
This removes duplication of the logic for neighbors iteration and enables walkers to be "rehydrated" to their former level of usability.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -805,8 +805,8 @@ where
         let mut iter = self.neighbors_undirected(a);
         if self.is_directed() {
             let k = dir.index();
-            iter.next[1 - k] = EdgeIndex::end();
-            iter.skip_start = NodeIndex::end();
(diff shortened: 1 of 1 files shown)


soooch:
interested in PRing this into the next branch (#567) as well. But it looks like it's not very fleshed out yet in this area.

In particular, it appears `graph::stable::Neigbors` and `graph::Neighbors` have separate implementations such that `graph::stable::Neighbors` may add a couple additional debug asserts to iteration. The same is not true of `graph::stable::WalkNeighbors` and `graph::WalkNeighbors`. Here, the same iteration happens, but no additional debug_asserts are added to `graph::stable::WalkNeighbors::next`.

I assume this inconsistency is a mistake and will note that it may have been avoided if either Neighbors or WalkNeighbors implemented itself in terms of the other.

Issue #466, opened 2021-09-14
[Question / Feature Request] Adding support for reading in a graph from a dotfile


saresend:
### Summary
This might already be supported, but I can't seem to find it on docs.rs:
I'd like a way to read in a graph from a dotfile

### Motivation

My usecase is the following: I have a tool that outputs graphs to a dotfile format, and I'd like to be able to perform operations with petgraph on the output of those tools. 

As an example, one thing that gets output is a graph of all of the dependencies my project has, and given a set of dependencies I'd like to remove, I'd like to find the minimum cut such that my project no longer depends on those libraries. Currently, the approach I have is to visually inspect the output of graphviz for plausible ways to do this, but I'd like to be able to write a tool that finds where I can remove them a bit easier. There are a couple other usages I'd like to add, and for that I'd love to have a way to operate on the contents of a dotfile, with petgraph

### Details

* Are you willing to implement this yourself? Mentor someone else and help them
  implement it?

Happy to help add this myself, just want to verify this is behavior that petgraph is ok supporting

saona-raimundo:
Related to #446 
(I just noticed the similarity while looking at the issues)

XVilka:
It's a good idea, but I think it should be a separate crate/library to avoid unnecessary dependencies. On the other hand, such crate could be a part of petgraph organization as well.

daniel-pfeiffer:
Not exactly what you're asking for, but related: [RustDOT: `graph! { A -- A -- B; A -- C; }`](https://github.com/petgraph/petgraph/issues/613)

daniel-pfeiffer:
I have generalised those ideas and published them as crate [rust_dot](https://docs.rs/rust_dot/latest/rust_dot/). Still a plain macro, not yet a proc-macro, but built around that technology. Turns out this now allows parsing strings or files.

RaoulLuque:
I think this is resolved in 0.8.2 by #653 :)

PR #653, opened 2024-07-10
feat(parser): allow parsing graphs from Dot/Graphviz files

### LINKED ISSUE #446: Parse Graphviz .dot file to Graph?
### Summary

Extension to [petgraph::dot::Dot](https://docs.rs/petgraph/0.5.1/petgraph/dot/struct.Dot.html) to parse graph from dot file.

### Motivation

There currently is a way to write Graphs to dot files but there is no way to read a Graph from a dot file into your program.

### Details

The best I could find was [dot_parse](https://docs.rs/dot_parse/0.2.0/libdot_parse/struct.Graph.html#method.parse_from) which seems to implement a low level parser for dot files. Maybe this can help reading dot files to petgraph graphs.

Bromind:
This is a PR that adds support for Dot/Graphviz parsing. Parsing can happen both at runtime and at compile time (i.e. there are functions and macros).

It fixes #446 and #466 .

Internally, it uses the `dot_parser` crate (https://crates.io/crates/dot-parser), the implementation is simply a thin wrapper over that crate. This wrapper is a trait (`ParseFromDot`), introduced in a new module `crate::dot::dot_parser`. I also added a feature `dot_parser` to gate this module.

The modifications do *not* introduce breaking changes.

TODO List (this is just what I have on the top of my head right now, feel free to edit if you think of other items):
- [x] Add more tests.
- [x] Implement `ParseFromDot` for existing graph types.
- [x] Discuss on exposed (and re-exported) items.
- [x] Improve the documentation. (If need be, I'll also improve the doc on the `dot_parser` crate side).


Bromind:
I added tests for parsing from both a string and an existing file (I used `graph-example.dot` in the root of the project).

Bromind:
In addition, I implemented `ParseFromDot` for `StableGraph` (trivially, there's nothing to do).

Cargo.toml
```diff
@@ -40,6 +40,8 @@ quickcheck = { optional = true, version = "0.8", default-features = false }
 serde = { version = "1.0", optional = true }
 serde_derive = { version = "1.0", optional = true }
 rayon = { version = "1.5.3", optional = true }
+dot-parser = { version = "0.3", optional = true}
```
bluss, line 43: I'm sorry, I don't have full bandwidth to review well, but I hope others do.

dot-parser already optionally depends on petgraph. How does it all interact, when petgraph optionally depends on dot-parser?

Bromind: Hi bluss, 

dot-parser depends on petgraph for a single thing: to export parsed graphs as petgraph's `Graph`. (see here : https://codeberg.org/bromind/dot-parser/src/branch/master/src/canonical.rs#L156 ).

Basically, this is a poor attempt to do what I intend to do (in a cleaner way) with this PR: to get petgraphs directly from Dot files. However, implementing that in `dot-parser` is intrinsically bad (imho), since it wouldn't be very flexible in terms of petgraph's version (e.g. as of today, `dot-parser` exports `Graph`s from petgraph 0.6.5, which causes issues if the user wants/needs to use an other version). 

Eventually, if this PR gets accepted, I'll removed the dependency. 

From a more practical aspect, the current status is not so much of a problem, since the PR does not uses the `petgraph` feature of dot-parser, so petgraph should not even appear down the dependency tree/there is not cyclic dependency issue.

Bromind:
Hi, I've been looking to implement the trait for other graph types. However, I think it is reasonable to have it implemented only for `Graph` and `StableGraph`. The reason is that the trait extends `Create` which has only 3 implementors: `Graph`, `StableGraph` and `GraphMap`. For `GraphMap`, the node weight must implement Copy, which graph payloads from the Dot-parser library do not. 
Therefore, I think it simpler to keep it that way. If a user wants GraphMap, if think it is reasonible to leave them the burden to create a `Graph` or a `StableGraph` with the default weights (i.e. dot-parser's `Node<...>` and `Edge<...>`), to convert those types into the type they want (which they'll probably do anyways), and then to create a `GraphMap` from the `Graph`.

Bromind:
I would like your opinion on re-exported items. So far, this PR re-export/show the following types from the dot-parser crate:

- `Node` and `AList`, which appear in the weights of Nodes and Edges bounds of `Create`, when extending the trait with `
- `AstGraph`, i.e. dot_parser::ast::Graph. It is not re-exported, but it is visible in the signature of `from_dot_graph`.

I'm also thinking about making pest errors directly visible, or nesting them in a wrapper.

Is there any opinion on those re-exports? I haven't think about it too much yet, and I'm not really happy with the current situation. Any suggestion?

Bromind:
So, regarding my previous comment about reexported items from the dot-parser library, I hid them behind type aliases. I think it is a good tradeoff. Also, parsing errors are now used: if an incorrect DOT string/file is used, an error will be displayed. This is simply using errors from the dot-parser lib and hiding them in ad-hoc opaque `DotParsingError`.

starovoid:
Hi, is it possible to add more tests with graph isomorphism checks? This will make it possible to more strictly check the correctness of parsing in order to avoid possible regressions in the future.

src/dot.rs
```diff
@@ -324,6 +324,117 @@ where
...
 }
 
+#[cfg(feature = "dot_parser")]
+#[macro_use]
+pub mod dot_parser {
+    use dot_parser::canonical::Node;
+    use dot_parser::ast::AList;
+    use dot_parser::canonical::Graph as CGraph;
+    use dot_parser::ast::Graph as DotGraph;
+    use dot_parser::ast::PestError as ParsingError;
```
starovoid, line 334: Please add re-export of `ast` and `canonical` modules so that the user can write `ParseFromDot` implementations for their own types in downstream crates.
Since all the elements are contained under a `feature`, it won't become unnecessarily verbose.

starovoid (changes requested):

Bromind:
> Hi, is it possible to add more tests with graph isomorphism checks? This will make it possible to more strictly check the correctness of parsing in order to avoid possible regressions in the future.

I added a few more tests in https://github.com/petgraph/petgraph/pull/653/commits/8ff3d6e96cfcabaab76bff0f5a885c9840b3d997 . Only 3 are related to isomorphism (checking isomorphism, isomorphism of subgraph, and not isomorphism), but I also included checks with cyclic graphs.

Bromind:
I just tested the semver-checks failure, and it definitively seems to be a false positive. In a new and separate project, I wrote a dummy graph to implement `Create` and `Build`, and both work like a charm. So I think we could ignore this warning.

(For reference: https://github.com/obi1kenobi/cargo-semver-checks/issues/1200)

```rust
use petgraph::data::{Build, Create};
use petgraph::visit::{Data, GraphBase, NodeCount};
use petgraph::dot::dot_parser::{ParseFromDot, DotNodeWeight, DotAttrList};
use std::marker::PhantomData;

#[derive(Default)]
struct MyDummyGraph<'a> {
    a: PhantomData<&'a ()>
}

impl<'a> GraphBase for MyDummyGraph<'a> {
    type EdgeId = ();
    type NodeId = ();
}

impl<'a> Data for MyDummyGraph<'a> {
    type NodeWeight = DotNodeWeight<'a>;
    type EdgeWeight = DotAttrList<'a>;
}

impl<'a> NodeCount for MyDummyGraph<'a> {
    fn node_count(&self) -> usize {
        0
    }
}

impl<'a> Build for MyDummyGraph<'a> {
    fn add_node(&mut self, _data: <Self as Data>::NodeWeight) -> <Self as GraphBase>::NodeId {
        return ();
    }

    fn update_edge(
        &mut self,
        _from: <MyDummyGraph as GraphBase>::NodeId,
        _to: <MyDummyGraph as GraphBase>::NodeId,
        _weight: <MyDummyGraph as Data>::EdgeWeight,
    ) -> <Self as GraphBase>::EdgeId {
        return ();
    }
}

impl<'a> Create for MyDummyGraph<'a> {
    fn with_capacity(_n: usize, _e: usize) -> Self {
        Self::default()
    }
}

impl<'a> ParseFromDot<'a> for MyDummyGraph<'a> {}

fn main() {
    println!("Hello, world!");
}

```

src/dot.rs
Bromind: About that, I tested and a user can implement `ParseFromDot` (see example in https://github.com/petgraph/petgraph/pull/653#issuecomment-2784535332, where this is basically what I did). It doesn't seem that exporting them is needed. That being said, if you want, I can re-export them all.

starovoid:
Please apply `cargo fmt` to satisfy the formatting lint

src/dot.rs
starovoid: Okay, let's not complicate the API with re-exports until someone specifically asks for it.

src/dot.rs
```diff
@@ -324,6 +324,174 @@ where
     }
 }
 
+#[cfg(feature = "dot_parser")]
+#[macro_use]
+pub mod dot_parser {
```
ABorgna, line 329: nitpick: It'd be easier to read if this module was reflected in the file hierarchy (in a `src/dot/dot_parser.rs`).

ABorgna (approved):
Nice!

PR #657, opened 2024-08-14
Add an option to format Dot graphs with Graph attributes

s1gtrap:
I needed a dot graph with a transparent background, and you can add attributes for nodes/vertices with `Dot::with_attr_getters`, so I was surprised to learn graph attributes weren't supported. Obviously there's only a single graph being formatted at a time, so adding a getter is a bit overkill. I made a `petgraph::dot::Config` variant with a key/value `String` do, and figured it might as well be submitted upstream.


src/dot.rs
```diff
@@ -111,6 +112,8 @@ pub enum Config {
     NodeNoLabel,
     /// Do not print the graph/digraph string.
     GraphContentOnly,
+    /// For aing any attributes, mostly intended for: https://graphviz.org/docs/graph/
```
XVilka, line 115: What `aing` means?

XVilka (changes requested):

src/dot.rs
s1gtrap: Oops! I probably meant for it to say 'adding'. Fixed.

XVilka:
@s1gtrap could you please rebase once more to fix the CI?

s1gtrap:
> @s1gtrap could you please rebase once more to fix the CI?

Rebased on d6d0ecdcdfad9fa16eef98ff166933f364155175.

XVilka (approved):

PR #658, opened 2024-08-19
Add graph6 format encoder and decoder for undirected graphs

BryanCruz:
### Add [graph6 format](https://users.cecs.anu.edu.au/~bdm/data/formats.txt) encoder and decoder for undirected graphs.

Format for storing undirected graphs in a compact manner, using only printable ASCII characters.

Files in this format have text type and contain one line per graph.

`graph6` is suitable for small graphs, or large dense graphs.

encoded graph examples:
* petersen: `IheA@GUAo`
* 30 vertices and 124 edges: ```GBYAcCC?JYOoq?_BgG?G@Kcig?C[cDacRBBS`DGGOMO@[A?isvUK?QOofI_SOSG_O?ahXGD??```

Format reference: https://users.cecs.anu.edu.au/~bdm/data/formats.txt

Other implementations and usages:
* networkX: https://networkx.org/documentation/stable/reference/readwrite/sparsegraph6.html
* Wolfram: https://reference.wolfram.com/language/ref/format/Graph6.html.en
* SageMath: https://doc.sagemath.org/html/en/reference/graphs/sage/graphs/graph.html#sage.graphs.graph.Graph.graph6_string
* nauty tool: https://pallini.di.uniroma1.it/
  * `graph6` can be used as an input and output format


XVilka:
Could you please put more details where this format is used? Because it's unclear from the description. Worth adding this information in the commit message as well.

BryanCruz:
hi @XVilka, thanks for the follow up on this

updated both the commit message and the PR description

XVilka (approved):

Cargo.toml
```diff
@@ -35,7 +35,7 @@ debug = true
 
 [dependencies]
 fixedbitset = { version = "0.4.0", default-features = false }
-indexmap = "2.0"
+indexmap = "~2.5.0"
```
cuviper, line 38: Please don't do this! It prevents users with more current compilers from using the latest petgraph and indexmap together, even though they would otherwise work fine.

A better way is to downgrade it just for your MSRV CI, either manually with `cargo update --precise` or using Rust 1.84's new resolver (in beta) before running the test with your actual MSRV.

* https://doc.rust-lang.org/beta/cargo/reference/resolver.html#rust-version
* to be announced: https://github.com/rust-lang/blog.rust-lang.org/pull/1451

madsmtm: I also just hit this, since I have another dependency that specifies `indexmap = "2.7"`.

PR #659, opened 2024-08-19
add random graph generator

BryanCruz:
Generates a random undirected graph with given `order` and with `p` probability of each possible edge existing.


bluss:
This overlaps with existing graph generation in https://github.com/petgraph/petgraph/blob/master/src/quickcheck.rs

PR #663, opened 2024-08-26
Update remove_node doc comment in graphmap.rs

MeGaGiGaGon:
The doc comment for `GraphMap::remove_node` confused me initially, making me think it was a fallible method, so this makes it match `GraphMap::remove_single_edge`.

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -282,7 +282,7 @@ where
         n
     }
 
-    /// Return `true` if node `n` was removed.
+    /// Return `true` if it did exist.
     ///
     /// Computes in **O(V)** time, due to the removal of edges with other nodes.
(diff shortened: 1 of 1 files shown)


src/graphmap.rs
```diff
@@ -282,7 +282,7 @@ where
         n
     }
 
-    /// Return `true` if node `n` was removed.
+    /// Return `true` if it did exist.
```
starovoid, line 285: Add description line "Remove node `n` from the graph".

In total, the doc comment should be:
```markdown
/// Remove node `n` from the graph.
///
/// Return `true` if it did exist.
```

starovoid (changes requested):

src/graphmap.rs
MeGaGiGaGon: Updated

PR #670, opened 2024-09-17
Fix #586

### LINKED ISSUE #586: immediately_dominated_by function called on root of graph returns root itself
Hello petgraph maintainers!

While using this amazing crate we stumbled upon 1 small bug. The documentation of the [immediately_dominated_by](https://docs.rs/petgraph/latest/petgraph/algo/dominators/struct.Dominators.html#method.immediately_dominated_by) states:

> Iterate over all nodes immediately dominated by the given node (not including the given node itself).

But when we call this function on a root of a graph the function returns the root itself. See the example.

If you have any questions please let me know.

Thank you 👋 

### Test Case

use petgraph::{algo::dominators, stable_graph::StableDiGraph};
fn main() {
    let mut graph: StableDiGraph<&str, ()> = StableDiGraph::new();
    let root = graph.add_node("ROOT");
    let a = graph.add_node("A");
    let b = graph.add_node("B");
    graph.extend_with_edges([(root, a), (root, b)]);
    let dominators = dominators::simple_fast(&graph, root);
    let dominated_by_root: Vec<_> = dominators
        .immediately_dominated_by(root)
        .map(|n| graph.node_weight(n).unwrap())
        .collect();
    println!("{:?}", dominated_by_root);
}

### Steps To Reproduce

 * Create a new binary project with petgraph as dependency
 * Paste the code in main.rs
 * let it run with cargo run

### Actual Results

["ROOT", "A", "B"]

### Expected Results

["A", "B"]

asoffer:
This makes it so the root of the dominator tree is not returned during
iteration over `dom_tree.immediately_dominated_by(root)`.

### DIFF
--- a/src/algo/dominators.rs
+++ b/src/algo/dominators.rs
@@ -133,7 +133,7 @@ where
 
     fn next(&mut self) -> Option<Self::Item> {
         for next in self.iter.by_ref() {
-            if next.1 == &self.node {
+            if next.1 == &self.node && next.1 != next.0 {
                 return Some(*next.0);
             }
         }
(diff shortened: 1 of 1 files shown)


ABorgna (approved):
Thanks!
I just added a comment to document the filtering.

PR #675, opened 2024-10-24
Dynamic Toposort Support

lmondada:
Hi there,

I would like to add support for directed acyclic graphs (DAGs) in petgraph. I propose adding a wrapper type `Acyclic<G>` that behaves like the underlying graph `G`, but checks for acyclicity whenever an edge addition is attempted. This functionality is provided through an implementation of the `Build` trait, as well as the `try_add_edge` method.

The idea is similar to the standalone crate [daggy](https://github.com/mitchmindtree/daggy) except that i) this PR implements a more efficient algorithm for this purpose and ii) I am unsure whether daggy is still maintained.

## Algorithm
I looked at the literature to find the recommended way for doing dynamic acyclicity checks (i.e. not run toposort everytime an edge is added). I have settled (for now) on the PK algorithm from [1]. This seems to strike a nice balance between simplicity and performance.

The idea is fairly straightforward: keep a Vec of toposorted nodes of the graph. When an edge is added that does not respect the current ordering, find all nodes in the order between the offending edge endpoints and reorder them appropriately.

[1] "A Dynamic Topological Sort Algorithm for Directed Acyclic Graphs" by D. Pierce and P. Kelly, JEA, 2004.

## Status of this PR
This is still work in progress, and you might be able to help me here (especially points 1-2).
- [x] I have attempted to implement the `Into{NodeIdentifiers, Edges, etc}` for all reference types, giving me a bound of the type
```rust
impl<'a, G: GraphBase> IntoNodeReferences for &'a Acyclic<G>
where
    &'a G: IntoNodeReferences { ... }
```
However, this creates an infinite recursion in the compiler, with the error:
```
error[E0275]: overflow evaluating the requirement `&visit::filter::NodeFiltered<_, _>: visit::IntoNodeIdentifiers`
```
I wonder if you have had this issue before when implementing other types. Do you have any suggestions on how I should solve this? Unlike `Reversed` and `NodeFiltered` (and others), the generic type `<G>` must be the graph itself and not a reference, as it must be mutable.

- [x] Benchmarking shows that the implemention is currently slow, dominated by memory allocations for the DFS traversals. Given that only a small region of the graph is expected to be traversed at each run, it is wasteful to reallocate a map the size of the graph at each call. The current code structure makes this a bit hard to change. One solution I can see is adding a `WithVisitMap(G)` wrapper type that can be used to specify a custom `VisitMap` type for `G`.
- [x]  Add quickcheck tests.


lmondada:
My new commit (incoming) solves the memory bottleneck, and I am getting decent results in benchmarking. The first line is the result using the dynamic toposort, the second is a baseline where a new toposort is re-computed from scratch every time an edge is added:

```
test acyclic_bench           ... bench:     418,393.75 ns/iter (+/- 7,771.51)
test toposort_baseline_bench ... bench:   1,383,431.25 ns/iter (+/- 109,105.02)
```

Unfortunately, I had to add a `G: Visitable` bound on the `Acyclic<G>` datatype, and so afaik I cannot use the `delegate_impl` macros. The result is some boilerplate, please tell me if you know of a way to get rid of it! I also had to increase the visibility of `dfs_visitor` to `pub(crate)` -- do you have a better solution in mind?

lmondada:
Points 2-3 from above are resolved. I am not sure how to solve the first point -- I'm hoping you have a suggestion, so I'll mark this as ready for review :)

lmondada:
I have resolved the last open point mentioned above (point 1) by only implementing the `IntoXYZ` traits for `Graph` and `StableGraph` using a macro. This should cover most simple use cases. The macro doesn't work for `Csr` or `DiGraphMap` without some changes, so I am leaving this to future work.

I have also changed the Dfs maps to `FixedBitSet` (instead of the VisitMap provided by the type) because that allows me to reset only the bits that have been changed. This is sensible because we should expect most edge additions to only traverse small regions of the total graph, and thus doing full resets is overkill. This has led to big speedups in my testing.

Looking forward to your review!

ABorgna (approved):
Nice!

The macro for trait implementations is a bit annoying, but I think it shouldn't be that much of an issue here.

Issue #666, opened 2024-09-08
Is there interest for an implementation of Paton's "An Algorithm for Finding a Fundamental Set of Cycles of a Graph"?


torfsen:
For a personal project I've implemented [Paton's "An Algorithm for Finding a Fundamental Set of Cycles of a Graph"](https://dl.acm.org/doi/abs/10.1145/363219.363232), which is, as far as I understand, still one of the go-to solutions for that problem. My implementation is based on `petgraph` without any additional dependencies, and I'd be happy to have it included in `petgraph` itself if there is any interest for that.

The code in its current form can be found here: https://github.com/torfsen/petgraph-paton. So far it's undocumented and has no tests (but it's been working without issues for me). Of course that's something I would work on if there is any interest, along with the other things mentioned in the ["Implementing New Algorithms"](https://github.com/petgraph/petgraph/blob/master/CONTRIBUTING.rst#implementing-new-algorithms) section of the contributing docs.

PR #676, opened 2024-11-03
Cycle Basis Algorithm

chocchillson:
A [cycle basis](https://en.wikipedia.org/wiki/Cycle_basis) algorithm based on algorithm [CACM 491]( https://dl.acm.org/doi/pdf/10.1145/363219.363232), and adapted from the [networkx implementation](https://networkx.org/documentation/stable/_modules/networkx/algorithms/cycles.html#cycle_basis).

This algorithm is for undirected graphs, and counts self-loops as cycle basis. I have adapted it to handle graphs with parallel edges (by ignoring subsequent edges beyond the first between nodes).

I have included a quickcheck that check against the cyclometric number, and a bench test.

I'm still new to rust, so any suggestions on how to make this more idiomatic (and performant) gratefully received. Also suggestions on any traits or existing petgraph features that could be better utilised.

### DIFF
--- a/benches/cycle_basis.rs
+++ b/benches/cycle_basis.rs
@@ -1,0 +1,31 @@
+#![feature(test)]
+
+extern crate petgraph;
+extern crate test;
+
+use petgraph::prelude::*;
+use std::cmp::{max, min};
+use test::Bencher;
+
+use petgraph::algo::cycle_basis;
+
+#[bench]
+fn cycle_basis_bench(bench: &mut Bencher) {
+    static NODE_COUNT: usize = 100;
+    let mut g = Graph::new_undirected();
+    let nodes: Vec<NodeIndex<_>> = (0..NODE_COUNT).map(|i| g.add_node(i)).collect();
+    for i in 0..NODE_COUNT {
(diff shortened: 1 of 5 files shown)


RaoulLuque:
Hey, wanted to ask if you are still interested in merging the PR, if so I'd be willing to review your code :) Just asking because I wouldn't want to dive into the topic/code for nothing 😊

chocchillson:
Hi, yes I am, if you think it's a useful addition.

RaoulLuque:
Also I have the following notes:
- An issue has been opened before this PR asking whether Petgraph would be interested in this algorithm. Maybe check our their linked implementation and compare for notes? https://github.com/petgraph/petgraph/issues/666
- There seems to be a distinction to be made with regard to a Fundamental Cycle Basis and general Cycle Basis. Please specify this in your Code. I suppose you are also computing a fundamental cycle basis since you linked to the same paper referenced in the issue above.
- Please give some more reference for the cyclomatic number as described in the quickcheck, since I am unable to find a precise definition with a quick websearch.

PR #678, opened 2024-11-09
Potential fix for `index_twice_mut` to work with stacked borrows

### LINKED ISSUE #582: Soundness issue in index_twice_mut
While conducting internal unsafe reviews @cramertj found a soundness issue in `index_twice_mut`:

https://github.com/petgraph/petgraph/blob/c78b5b13a816f1f0daf37f3964db0d33c6cbf38d/src/graph_impl/mod.rs#L1156-L1162

This is unsound because it uses a reference to `&mut *self_mut` after creating the first indexed-value reference, invalidating it. This should be changed to use a by-pointer indexing method.

Note that this issue also exists in 

https://github.com/petgraph/petgraph/blob/c78b5b13a816f1f0daf37f3964db0d33c6cbf38d/src/graph_impl/stable_graph/mod.rs#L775-L781

Imberflur:
Fixes #582.

This is done via a new utility method that uses pointer arithmetic instead of Vec indexing to avoid materializing a reference to the whole slice.

I assume that `GraphIndex` is meant to be only implemented in this crate since it has `doc(hidden)` methods. Thus, more strictly enforcing this by sealing the trait can be considered a non-breaking change. This is done because we need to trust the values returned by `GraphIndex` for `twice_index_mut` to be sound.

 I'm assuming this will all be replaced in https://github.com/petgraph/petgraph/pull/567 (as noted by https://github.com/petgraph/petgraph/issues/582#issuecomment-1699803181). And it seems like with a breaking change there is room for rearranging traits to make this cleaner or some other solution.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -1152,9 +1152,19 @@ where
         T: GraphIndex,
         U: GraphIndex,
     {
+        // We can trust these values since GraphIndex is sealed.
         assert!(T::is_node_index() != U::is_node_index() || i.index() != j.index());
 
         // Allow two mutable indexes here -- they are nonoverlapping
+        //
+        // We know this will either use the IndexMut impl for NodeIndex or EdgeIndex since
+        // GraphIndex is only implemented for those two types. Both of those IndexMut
+        // implementations avoid stacked borrows violations by using
+        // util::index_mut_no_sb_invalidation which avoids materializing a reference to the whole
+        // slice.
+        //
+        // This approach is taken to avoid breaking changes. This is expected to be removed or
+        // replaced with a cleaner solution in https://github.com/petgraph/petgraph/pull/567.
         unsafe {
             let self_mut = self as *mut _;
             (
@@ -1870,7 +1880,7 @@ where
     Ix: IndexType,
 {
     fn index_mut(&mut self, index: NodeIndex<Ix>) -> &mut N {
-        &mut self.nodes[index.index()].weight
+        &mut crate::util::index_mut_no_sb_invalidation(&mut self.nodes, index.index()).weight
     }
 }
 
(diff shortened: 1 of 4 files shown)


XVilka (changes requested):
1. Please rebase to check the CI again
2. This solution looks like overengineered to me.

@bluss @ABorgna do you have any opinion on that?

Imberflur:
> or mark the PR with a `!` following the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

Is this the PR title? Description?

> This solution looks like overengineered to me.

Happy to iterate on this.

RaoulLuque:
Looks good to me : ) Tested locally, and the added test does fail with Miri with the old version and passes with your changes. I think we could think about not sealing this trait for now, as that would technically still be a breaking change, even if probably nobody will have implemented the trait since it's methods are `doc(hidden)`.

Thanks :sauropod: 

PS: Will check, why CI is not running ^^

RaoulLuque:
CI is now able to run with the newest CI related PRs being merged to `master` and `0.8` ^^

Imberflur:
> I think we could think about not sealing this trait for now, as that would technically still be a breaking change, even if probably nobody will have implemented the trait since it's methods are doc(hidden).

I went ahead and did this! At least this should be possible to fully address in the next version of the graphs. FYI if anyone has implemented this, then they most certainly have the same UB that is fixed here, in their associated `IndexMut` impl.

RaoulLuque:
I see, thanks :) Yeah, that makes sense with the UB. Maybe there is some sort of hint we could give as a warning, which is not considered a breaking change ? :thinking:

Imberflur:
Perhaps a note could be added to the docs of `GraphIndex`?

PR #590, opened 2023-10-07
feat: Add algorithm to find bridge edges

pragmaticTNT:
Derived from #473 (written by @starovoid).

Here's a list of changes that we (@gmorenz and myself) made:
- Used iteration instead of recursion; with the original version we hit a stack overflow. 
- Tried to make variable names self documenting (e.g. `fup` to `earliest-backedge`)
- Change the output type from `Vec<G::NodeId, G::NodeId)>` to `impl Iterator<Item = G::EdgeRef>`

The original code also had an algorithm to find articulation points (node variant of bridge edges). To keep the pull request small we only added the bridge algorithm. We would be happy to include the other algorithm in another pull request. As a result we named the module `bridges` instead of `bridges_and_ap`, we think this is probably a better (more succinct) name anyways.

PR #681, opened 2024-11-24
Add articulation points implementation

daehiff:
### Motivation

This pull request adds a function to compute articulation points in a graph using Tarjan's algorithm, implemented with an iterative approach. Articulation points are nodes whose removal increases the number of connected components in the graph. 
Currently, `petgraph` lacks an implementation for this functionality. This makes it impossible to for instance determine if a vertex of a graph is cutting with petgraph.

*Note*: This approach is not a direct replication of existing implementations (e.g., [[here](https://www.geeksforgeeks.org/tarjan-algorithm-find-strongly-connected-components/)](https://www.geeksforgeeks.org/tarjan-algorithm-find-strongly-connected-components/)), as it avoids recursion entirely.

### Runtime

Tarjan's algorithm is an efficient solution to this problem, with a time complexity of \(O(V + E)\), making it well-suited for large graphs.

### DIFF
--- a/src/algo/articulation_points.rs
+++ b/src/algo/articulation_points.rs
@@ -1,0 +1,78 @@
+use std::cmp::min;
+use std::collections::{HashMap, HashSet};
+use std::hash::Hash;
+
+use crate::visit::{EdgeRef, IntoEdges, IntoNodeReferences, NodeIndexable, NodeRef};
+
+pub fn articulation_points<G>(g: G) -> HashSet<G::NodeId>
+where
+    G: IntoNodeReferences + IntoEdges + NodeIndexable,
+    G::NodeWeight: Clone,
+    G::EdgeWeight: Clone + PartialOrd,
+    G::NodeId: Eq + Hash,
+{
+    let mut visited = HashSet::with_capacity(g.node_references().size_hint().0);
+    let mut parent = HashMap::with_capacity(g.node_references().size_hint().0);
+    let mut low = HashMap::with_capacity(g.node_references().size_hint().0);
+    let mut disc = HashMap::with_capacity(g.node_references().size_hint().0);
+    let mut ap = HashSet::with_capacity(g.node_references().size_hint().0);
+    let mut time = 0;
+
+    for node in g.node_references() {
+        let node_id = g.to_index(node.id());
+        if !visited.contains(&node_id) {
+            dfs(
+                &g,
+                node_id,
+                &mut visited,
+                &mut parent,
+                &mut low,
+                &mut disc,
+                &mut ap,
+                &mut time,
+            );
+        }
+    }
+
+    ap.into_iter().map(|id| g.from_index(id)).collect()
+}
+
+fn dfs<G>(
+    g: &G,
+    u: usize,
+    visited: &mut HashSet<usize>,
+    parent: &mut HashMap<usize, usize>,
+    low: &mut HashMap<usize, usize>,
+    disc: &mut HashMap<usize, usize>,
+    ap: &mut HashSet<usize>,
+    time: &mut usize,
+) where
+    G: IntoEdges + NodeIndexable,
+{
+    visited.insert(u);
+    *time += 1;
+    disc.insert(u, *time);
+    low.insert(u, *time);
+    let mut children = 0;
+
+    for edge in g.edges(g.from_index(u)) {
+        let v = g.to_index(edge.target());
+        if !visited.contains(&v) {
+            children += 1;
+            parent.insert(v, u);
+            dfs(g, v, visited, parent, low, disc, ap, time);
+
+            low.insert(u, min(low[&u], low[&v]));
+
+            if parent.contains_key(&u) && low[&v] >= disc[&u] {
+                ap.insert(u);
+            }
+        } else if v != parent.get(&u).cloned().unwrap_or(usize::MAX) {
+            low.insert(u, min(low[&u], disc[&v]));
+        }
+    }
+
+    if parent.get(&u).is_none() && children > 1 {
+        ap.insert(u);
+    }
+}
--- a/src/algo/articulation_points.rs
+++ b/src/algo/articulation_points.rs
@@ -1,27 +1,65 @@
+use crate::visit;
+use crate::visit::{EdgeRef, IntoEdges, IntoNodeReferences, NodeIndexable, NodeRef};
 use std::cmp::min;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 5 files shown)


src/algo/articulation_points.rs
```diff
@@ -0,0 +1,78 @@
...
+                &mut ap,
+                &mut time,
+            );
+        }
+    }
+
+    ap.into_iter().map(|id| g.from_index(id)).collect()
+}
+
+fn dfs<G>(
```
daehiff, line 40: Think there is a DFS function in petgraph

starovoid:
Hi, great job! 

But I think it will be better the implementation using stack instead of recursion. Just like it was with the bridge edges search algorithm in PR #590.

Practice has shown that a recursive implementation can face the problem of stack overflow.

daehiff:
Hey, thanks for the quick reply!

That is a good suggestion! I'll update the PR. 
Do you have any other requirements (testing concepts ect.), we should be aware of?

starovoid:
Yes, I have two thoughts

# Picky testing
I think it would be a good idea to write some randomized tests.

In general, the algorithm may look like this:

1. Creating a random graph on a random number of vertices;
2. Find the articulation points;
3. We remove the found vertices from the graph one by one and check that each time the number of connectivity components increases;

Later, when version `0.7.0` is released, I plan to open a PR with the addition of a randomized test for `bridges` implementation too.

# Add bench
`petgraph` has a good tradition of adding performance tests for new algorithms. 
I think it would be logical to add a benchmark identical to #590, since the algorithms are similar.

daehiff:
Alright, sorry the late reply.

I have added:

1. A benchmark similar to bridges
2. A few more tests and some other graph structures
3. A quickcheck test for random graphs

XVilka:
@daehiff please update the PR description to describe the reasons for this implementation, etc.

daehiff:
@XVilka all checks have been passed and there has been some review comments.

Is there a timeline or a ways how we can get in touch? It is kind of hard todo a contribution without knowing contribution guidelines ect.

I asked @ABorgna via discord, but no response on that end.

XVilka:
> @XVilka all checks have been passed and there has been some review comments.
> 
> Is there a timeline or a ways how we can get in touch? It is kind of hard todo a contribution without knowing contribution guidelines ect.
> 
> I asked @ABorgna via discord, but no response on that end.

I forgot we have a DIscord. Just joined it.

src/algo/articulation_points.rs
```diff
@@ -0,0 +1,146 @@
...
+
+/// \[Generic\] Find articulation points in a graph using [Tarjan's algorithm](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm).
+///
+/// Compute the articulation points of a graph (Nodes, which would increase the number of connected components in the graph.
+///
+/// # Arguments
+/// * `graph`: A directed graph
+///
+/// # Returns
+/// * `HashSet`: Hashset of the node ids which correspond to the articulation points of the grpah.
```
ABorgna, line 15:
```suggestion
/// * `HashSet`: Hashset of the node ids which correspond to the articulation points of the graph.
```

tests/articulation_points.rs
```diff
@@ -0,0 +1,183 @@
+use petgraph::{
+    algo::articulation_points::articulation_points,
+    csr::IndexType,
```
ABorgna, line 3:
```suggestion
```

tests/articulation_points.rs
```diff
@@ -0,0 +1,183 @@
...
+    graph::{NodeIndex, UnGraph},
+};
+
+use std::collections::HashSet;
+
+#[test]
+fn art_single_node() {
+    let mut gr = UnGraph::<&str, ()>::new_undirected();
+    let a = gr.add_node("A");
```
ABorgna, line 12:
```suggestion
    let _a = gr.add_node("A");
```

src/algo/articulation_points.rs
```diff
@@ -0,0 +1,146 @@
...
+    ap: &mut HashSet<usize>,
+    time: &mut usize,
+) where
+    G: IntoEdges + NodeIndexable,
+{
+    let mut stack: Vec<RecursionStep> = vec![RecursionStep::BaseStep(target_node)];
+    let mut children_count: HashMap<usize, usize> = HashMap::new();
+
+    while let Some(recursionStep) = stack.pop() {
+        match recursionStep {
```
ABorgna, line 103:
```suggestion
    while let Some(recursion_step) = stack.pop() {
        match recursion_step {
```

src/algo/articulation_points.rs
```diff
@@ -0,0 +1,146 @@
...
+    ProcessChildStep(usize, usize),
+    NoBackEdgeConditionCheck(usize, usize),
+    RootMoreThanTwoChildrenCheck(usize),
+}
+
+/// Helper that performs the required DFS in an iterative manner.
+fn _dfs<G>(
+    g: &G,
+    target_node: usize,
+    visited: &mut Vec<bool>,
```
ABorgna, line 90: `Vec<bool>` uses one byte per element, since it needs to support arbitrary references.

Use a [`FixedBitSet`](https://docs.rs/fixedbitset/latest/fixedbitset/struct.FixedBitSet.html) instead.

src/algo/articulation_points.rs
```diff
@@ -0,0 +1,146 @@
...
+    let mut ap = HashSet::with_capacity(graph_size);
+    let mut time = 0;
+
+    for node in g.node_references() {
+        let node_id = g.to_index(node.id());
+        if !visited[node_id] {
+            _dfs(
+                &g,
+                node_id,
+                &mut visited,
```
ABorgna, line 65: nitpick: This could be an auxiliary state struct.

ABorgna (commented):
Nice! I mostly only have nitpicks.

src/algo/articulation_points.rs
daehiff: Added a struct `ArticulationPointTracker`, I am kindof unhappy with the naming tho.

src/algo/articulation_points.rs
daehiff: Added a struct `ArticulationPointTracker`, I am kindof unhappy with the naming tho.

ABorgna: I'd normally call it "Context", but it's private so it doesn't really matter.

ABorgna (approved):

PR #682, opened 2024-11-24
feat: Add Kou's algorithm for finding a MST

daehiff:
# Motivation

This PR implements a generic Steiner Tree algorithm for undirected graphs based on Kou's algorithm. The algorithm leverages Dijkstra's shortest path, Floyd-Warshall path reconstruction, and minimum spanning tree computation to determine the Steiner tree for a set of terminal nodes. 
The implementation follows the implementation in the [networkx library](https://networkx.org/documentation/stable/_modules/networkx/algorithms/shortest_paths/dense.html#floyd_warshall). A detailed description can be found in [A fast algorithm for Steiner trees](https://link.springer.com/article/10.1007/BF00288961).

`petgraph` has no algorithm to approximate a minimal steiner tree. The algorithm [melhorn](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.approximation.steinertree.steiner_tree.html) in the networkx documentation, seems to be only different by the complexity class and apperently is an extension. So this could be implemented at a later point.

# Runtime
 $\mathcal{O}(|S| |V|^2)$, where \(|S|\) is the terminal set size, \(|V|\) and \(|E|\) are graph nodes and edges, respectively.

# Further Notes

Floyd-warshall implementation is adjusted to also allow path reconstruction (see [here](https://en.wikipedia.org/wiki/Floyd–Warshall_algorithm)) this is required in the function `subgraph_edges_from_metric_closure`


XVilka (dismissed):
Please fix this:
```

---- src/algo/floyd_warshall.rs - algo::floyd_warshall::floyd_warshall_path (line 127) stdout ----
error[E0432]: unresolved import `petgraph::algo::floyd_warshall_path`
 --> src/algo/floyd_warshall.rs:130:5
  |
6 | use petgraph::algo::floyd_warshall_path;
  |     ^^^^^^^^^^^^^^^^-------------------
  |     |               |
  |     |               help: a similar name exists in the module: `floyd_warshall`
  |     no `floyd_warshall_path` in `algo`

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0432`.
Couldn't compile the test.

failures:
    src/algo/floyd_warshall.rs - algo::floyd_warshall::floyd_warshall_path (line 127)
```

daehiff:
I always get the following issue when trying to run the pipeline: `1 workflow awaiting approval `. Any possible way around this?

XVilka:
> I always get the following issue when trying to run the pipeline: `1 workflow awaiting approval `. Any possible way around this?
> 
> At least `cargo test --verbose --no-default-features` went trough in the pipeline. So this seems then to be a merge-commit/pipeline issue.

It is a restriction by GitHub to avoid malicious CI runs from external contributors. Once the first PR of you will get merged, it wont happen anymore on any subsequent PRs.

XVilka:
@daehiff it still happens

daehiff:
> @daehiff it still happens

Ohh, sorry about that.

src/algo/dijkstra.rs
ABorgna, line 1: Leftover diff?

src/algo/floyd_warshall.rs
```diff
@@ -10,7 +10,7 @@ use crate::visit::{
 #[allow(clippy::type_complexity, clippy::needless_range_loop)]
 /// \[Generic\] [Floyd–Warshall algorithm](https://en.wikipedia.org/wiki/Floyd%E2%80%93Warshall_algorithm) is an algorithm for all pairs shortest path problem
 ///
-/// Compute shortest paths in a weighted graph with positive or negative edge weights (but with no negative cycles)
+/// Compute distance of shortest paths in a weighted graph with positive or negative edge weights (but with no negative cycles)
```
ABorgna, line 13: "Length" makes more sense here.
|
```suggestion
/// Compute the length of each shortest path in a weighted graph with positive or negative edge weights (but with no negative cycles).
```

src/algo/floyd_warshall.rs
ABorgna, line 1: Line 83 emits a warning now:

```
warning: variable does not need to be mutable
  --> src/algo/floyd_warshall.rs:83:5
   |
83 |     mut edge_cost: F,
   |     ----^^^^^^^^^
   |     |
   |     help: remove this `mut`
   |
   = note: `#[warn(unused_mut)]` on by default
```

ABorgna (commented):
Looks great!

I ran the benchmarks to see if the extra code made the old `floyd_warshal` impl slower, but it actually seems to be sightly faster now :)

You mentioned you'll add a quickcheck test, I'll wait for that before merging.

ABorgna (approved):

Issue #683, opened 2024-11-25
Add methods to add new sets to UnionFind


programmerjake:
### Summary

`UnionFind` is missing the ability to add new sets (increasing the internal vectors' length) after creating the `UnionFind`.

### Motivation

It's part of the standard disjoint-set data-structure operations. Also, without it it makes it very annoying when trying to transform some problem into a disjoint-set data-structure, where you incrementally add sets and union existing sets as you go along, since currently the API limitations requires you to convert that incremental algorithm into a 2-pass algorithm so you can calculate the length to pass into `UnionFind::new`, and then the second pass actually uses the `UnionFind` data-structure.

### Details

* What code needs to change? New traits? Which modules?

Add a new method to `UnionFind` to add a new set and return the new set's index.

* Is this a new graph algorithm? Include links to papers, Wikipedia, or other
  implementations of this algorithm that exist.

https://en.wikipedia.org/wiki/Disjoint-set_data_structure#Making_new_sets

* Are you willing to implement this yourself? Mentor someone else and help them
  implement it?

I can implement this, it's pretty trivial...

PR #684, opened 2024-11-25
Add `UnionFind::new_set`

programmerjake:
Adds `UnionFind::new_set` to add a new set to an existing `UnionFind`, which is helpful when you want to incrementally use the `UnionFind` algorithm and don't know how many elements you have ahead of time.

Fixes: #683

Also adds additional utility methods since `UnionFind` doesn't have fixed length anymore:
* `UnionFind::new_empty()` -- equivalent to `Vec::new()` for when you need a `const` empty `UnionFind`.
* `UnionFind::is_empty()`/`UnionFind::len()` -- returns the number of elements, like `Vec::is_empty()`/`Vec::len()`
* `impl Default for UnionFind` -- makes an empty `UnionFind`

I also switched `self.parent.len()` calls to use `self.len()` since that's more canonical.

I also added a test checking that `UnionFind::new_set()` works correctly.

### DIFF
--- a/src/unionfind.rs
+++ b/src/unionfind.rs
@@ -26,6 +26,15 @@ pub struct UnionFind<K> {
     rank: Vec<u8>,
 }
 
+impl<K> Default for UnionFind<K> {
+    fn default() -> Self {
+        Self {
+            parent: Vec::new(),
+            rank: Vec::new(),
+        }
+    }
+}
+
 #[inline]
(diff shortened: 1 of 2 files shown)


programmerjake:
it's been a while and there have been multiple releases since I made this PR...can it be merged or do I need to make some changes?

ABorgna (approved):
This is great, thank you!

PR #686, opened 2024-11-28
feat: Add Shortest Path Faster Algorithm Implementation

starovoid:
SPFA (Shortest Path Faster Algorithm) is the improved variation of Bellman-Ford.

### Time Complexity: 
- Average Time Complexity: `O(|E|) `, but this bound has not been approved yet.
- Worstcase Time Complexity: `O(|V|*|E|)`

### Space complexity
- Space complexity: `O(V)`

### [Wikipedia](https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm#CITEREFMoore1959) says:
A variation of the Bellman–Ford algorithm described by [Moore (1959)](https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm#CITEREFMoore1959), reduces the number of relaxation steps that need to be performed within each iteration of the algorithm. If a vertex v has a distance value that has not changed since the last time the edges out of v were relaxed, then there is no need to relax the edges out of v a second time. In this way, as the number of vertices with correct distance values grows, the number whose outgoing edges that need to be relaxed in each iteration shrinks, leading to a constant-factor savings in time for [dense graphs](https://en.wikipedia.org/wiki/Dense_graph). This variation can be implemented by keeping a collection of vertices whose outgoing edges need to be relaxed, removing a vertex from this collection when its edges are relaxed, and adding to the collection any vertex whose distance value is changed by a relaxation step. In China, this algorithm was popularized by Fanding Duan, who rediscovered it in 1994, as the "shortest path faster algorithm".[[6]](https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm#cite_note-6)

### About implementation
I used a stack instead of a queue to improve performance. In this regard, there is a slight discrepancy with the classic description of the "SPFA" algorithm, but this option was also tested on random tests against Bellman-Ford [here](https://github.com/starovoid/graphalgs/blob/main/src/shortest_path/spfa.rs).

### DIFF
--- a/benches/spfa.rs
+++ b/benches/spfa.rs
@@ -1,0 +1,35 @@
+#![feature(test)]
+
+extern crate petgraph;
+extern crate test;
+
+use petgraph::prelude::*;
+use std::cmp::{max, min};
+use test::Bencher;
+
+use petgraph::algo::spfa;
+
+#[bench]
+fn spfa_bench(bench: &mut Bencher) {
+    static NODE_COUNT: usize = 100;
+    let mut g = Graph::new();
+    let nodes: Vec<NodeIndex<_>> = (0..NODE_COUNT).map(|i| g.add_node(i)).collect();
+    for i in 0..NODE_COUNT {
+        let n1 = nodes[i];
+        let neighbour_count = i % 8 + 3;
+        let j_from = max(0, i as i32 - neighbour_count as i32 / 2) as usize;
+        let j_to = min(NODE_COUNT, j_from + neighbour_count);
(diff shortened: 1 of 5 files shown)


starovoid:
Made the rebase to start a new branch based on `spfa` soon. I plan to add an implementation of [Johnson's algorithm](https://en.wikipedia.org/wiki/Johnson%27s_algorithm) based on `spfa` in order to get maximum performance on sparse graphs.

PR #692, opened 2024-12-01
ci: Detect breaking PRs with `semver-checks`

ABorgna:
Runs [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) on each PR comparing it against the its base, and posting a comment when breaking API changes are detected.

The action fails on detected changes unless we flag the PR as breaking following the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format. In that case we still post a comment with the results.

Note that there exists a [`cargo-semver-checks-actions`](https://github.com/obi1kenobi/cargo-semver-checks-action) by obi1kenobi already but it's targeted towards comparing against the latest published crate instead of checking individual PRs, so we cannot use it here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,10 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: '*'
+  merge_group:
+    types: [ checks_requested ]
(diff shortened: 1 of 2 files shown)


XVilka (approved):
Nice idea!

PR #696, opened 2024-12-09
Edges undirected (refresh #352)

starovoid:
I have rebased the changes proposed in PR #352, and indeed I see a slight decrease in performance.

Now I'm going to look for a way to get back the previous edge iteration performance, but at the moment I need ideas.

### DIFF
--- a/src/graph_impl/stable_graph/mod.rs
+++ b/src/graph_impl/stable_graph/mod.rs
@@ -176,6 +176,17 @@ impl<N, E> StableGraph<N, E, Directed> {
     }
 }
 
+impl<N, E> StableGraph<N, E, Undirected> {
+    /// Create a new `StableGraph` with undirected edges.
+    ///
+    /// This is a convenience method. See `StableGraph::with_capacity`
+    /// or `StableGraph::default` for a constructor that is generic in all the
+    /// type parameters of `StableGraph`.
+    pub fn new_undirected() -> Self {
+        Self::with_capacity(0, 0)
+    }
+}
+
 impl<N, E, Ty, Ix> StableGraph<N, E, Ty, Ix>
 where
     Ty: EdgeType,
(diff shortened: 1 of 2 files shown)


starovoid:
I have tested several implementations, and I think that without changing the signature of `Edges` or unsafe code, it is not possible to return past performance, since it is necessary to dynamically check which iterator to use on each call of `Edges::next`.
Now I plan to make the `Edges` back a structure, and I think further decisions should be made by the maintainers.

It may be worth adding the `EdgesUndirect` struct in addition to `Edges` or changing the signatures to specify an internal iterator (`Single` or `Chain`), avoiding checks during iteration.

src/graph_impl/stable_graph/mod.rs
```diff
@@ -271,7 +282,6 @@ where
             self.g.add_node(Some(weight))
         }
     }
-
```
XVilka, line 274: Unnecessary change?

src/graph_impl/stable_graph/mod.rs
```diff
@@ -732,6 +691,46 @@ where
             ty: PhantomData,
         }
     }
+    /// Lookup if there is an edge from `a` to `b`.
```
XVilka, line 694: Missing newline

XVilka (changes requested):

starovoid:
> I am not sure it would be a good change in 0.6.x series. Something like that is a better fit for 0.7.x

Hi, I think that in `0.7.x` these changes will lose their relevance altogether, due to the new graph design in general. Or they will have to be significantly redesigned again.

XVilka:
> > I am not sure it would be a good change in 0.6.x series. Something like that is a better fit for 0.7.x
> 
> Hi, I think that in `0.7.x` these changes will lose their relevance altogether, due to the new graph design in general. Or they will have to be significantly redesigned again.

Yes, it's quite possible. Lets keep this open until that.

PR #701, opened 2024-12-24
Almost satisfy clippy

starovoid:
This PR satisfies almost all the warnings from Clippy.

In some places, Clippy warnings were indeed fixed, in others `#[allow(...)]` were used.

#### 1. Fix dead code warnings
1. Remove unused method of private trait `WithDummy`
2. Allow unused fields in some structs that implement `Debug` (it may be breaking to just remove these fields).

#### 2. Remove unecessary lifetime annotations
Most cases looks like:

```rust
// Before:
impl<'a, E, Ix: IndexType> visit::EdgeRef for EdgeReference<'a, E, Ix> 

// After
impl<E, Ix: IndexType> visit::EdgeRef for EdgeReference<'_, E, Ix>
```

#### 6. Allow clippy::needless_range_loop in benches

### DIFF
--- a/benches/acyclic.rs
+++ b/benches/acyclic.rs
@@ -11,6 +11,7 @@ use test::Bencher;
 
 /// Dynamic toposort using Acyclic<G>
 #[bench]
+#[allow(clippy::needless_range_loop)]
 fn acyclic_bench(bench: &mut Bencher) {
     static NODE_COUNT: usize = 100;
     let mut g = Acyclic::<DiGraph<usize, ()>>::new();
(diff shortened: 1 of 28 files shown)


starovoid:
**Why "Almost"?**

I haven't decided what to do with the latest warning yet:

```
warning: type parameter `Ty` goes unused in function definition
   --> benches/common/factories.rs:180:16
    |
180 | fn parse_graph<Ty, G>(s: &str) -> G
    |                ^^
    |
    = help: consider removing the parameter
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#extra_unused_type_parameters
    = note: `#[warn(clippy::extra_unused_type_parameters)]` on by default
```

I think the right solution would be to just ignore it using `#[allow(...)]`, but perhaps you can suggest a better solution.

XVilka (approved):
Thanks for those! Checked everything and slight modernization is good.

PR #709, opened 2024-12-31
Feat: WFC-C Algorithm for generating vertex coloring

AnarchistHoneybun:
## Details
Computes a valid [vertex coloring](https://en.wikipedia.org/wiki/Graph_coloring) for an undirected graph using the Wave Function Collapse algorithm. Returns a `HashMap` that maps node IDs to their assigned colors.

Algorithm Source: https://arxiv.org/pdf/2108.09329

## Requirements

- [ ] quickcheck property test
- [ ] benchmark test
- [ ] algorithm documentation
- [ ] complexity details
- [x] reading material links
- [ ] make algo work for generic graphs

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -19,6 +19,7 @@ pub mod min_spanning_tree;
 pub mod page_rank;
 pub mod simple_paths;
 pub mod tred;
+pub mod wfc_color;
 
 use std::num::NonZeroUsize;
 
--- a/src/algo/wfc_color.rs
+++ b/src/algo/wfc_color.rs
@@ -1,0 +1,147 @@
+use crate::visit::{IntoNeighbors, NodeCount, NodeIndexable, Visitable};
+use std::hash::Hash;
+use std::collections::HashMap;
+
+/// [Generic] Wave Function Collapse graph coloring algorithm.
+///
+/// Compute a valid vertex coloring for an undirected graph using the Wave Function
+/// Collapse algorithm. The algorithm assigns colors (represented as usize values)
+/// to vertices such that no adjacent vertices share the same color.
+///
+/// The graph should be `Visitable` and implement `IntoNeighbors`. The implementation
+/// uses 1-based color numbering.
+///
+/// Returns a `HashMap` that maps node IDs to their assigned colors.
+///
+/// # Example
+/// ```rust
+/// use petgraph::Graph;
+/// use petgraph::Undirected;
+/// use std::collections::HashMap;
+/// use petgraph::algo::wfc_color::wfc_color;
+///
+/// let mut graph = Graph::<(), (), Undirected>::new_undirected();
+/// let a = graph.add_node(());
+/// let b = graph.add_node(());
+/// let c = graph.add_node(());
+///
+/// graph.extend_with_edges(&[
+///     (a, b),
+///     (b, c),
+///     (c, a),
+/// ]);
+///
+/// let coloring = wfc_color(&graph);
+/// assert_ne!(coloring[&a], coloring[&b]); // Adjacent vertices have different colors
+/// ```
+pub fn wfc_color<G>(graph: G) -> HashMap<G::NodeId, usize>
+where
+    G: IntoNeighbors + NodeCount + NodeIndexable + Visitable,
+    G::NodeId: Eq + Hash + Copy,
+{
+    // Helper function for constraint propagation
+    fn propagate<G>(
+        graph: G,
+        start: G::NodeId,
+        colors: &mut HashMap<G::NodeId, usize>,
+        domains: &mut HashMap<G::NodeId, Vec<usize>>,
+    ) where
+        G: IntoNeighbors,
+        G::NodeId: Eq + Hash + Copy,
+    {
+        let mut stack = vec![start];
+        while let Some(u) = stack.pop() {
+            if let Some(&color) = colors.get(&u) {
+                for neighbor in graph.neighbors(u) {
+                    if !colors.contains_key(&neighbor) {
+                        if let Some(domain) = domains.get_mut(&neighbor) {
+                            if domain.contains(&color) {
+                                domain.retain(|&c| c != color);
+                                if domain.len() == 1 {
+                                    colors.insert(neighbor, domain[0]);
+                                    stack.push(neighbor);
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    // Helper function to calculate entropy of a node
+    fn entropy<N>(node: N, colors: &HashMap<N, usize>, domains: &HashMap<N, Vec<usize>>) -> usize
+    where
+        N: Eq + Hash + Copy,
+    {
+        if colors.contains_key(&node) {
+            0
+        } else {
+            domains.get(&node).map_or(0, |d| d.len())
+        }
+    }
+
+    // Initialize state
+    let max_degree = (0..graph.node_bound())
+        .map(|i| graph.from_index(i))
+        .map(|n| graph.neighbors(n).count())
+        .max()
+        .unwrap_or(0);
+
+    let mut max_colors = max_degree + 1;
+    let mut colors = HashMap::new();
+    let mut domains = HashMap::new();
+
+    loop {
+        // Reset for this attempt
+        colors.clear();
+        domains = (0..graph.node_bound())
+            .map(|i| graph.from_index(i))
+            .map(|n| (n, (1..=max_colors).collect()))
+            .collect();
+
+        // Color highest degree vertex first
+        let start = (0..graph.node_bound())
+            .map(|i| graph.from_index(i))
+            .max_by_key(|&n| graph.neighbors(n).count())
+            .unwrap();
+
+        colors.insert(start, 1);
+        propagate(&graph, start, &mut colors, &mut domains);
+
+        // Main coloring loop
+        while colors.len() < graph.node_count() {
+            // Find uncolored vertex with lowest non-zero entropy
+            let next = (0..graph.node_bound())
+                .map(|i| graph.from_index(i))
+                .filter(|v| !colors.contains_key(v))
+                .min_by_key(|&v| {
+                    let e = entropy(v, &colors, &domains);
+                    if e == 0 {
+                        usize::MAX
+                    } else {
+                        e
+                    }
+                });
+
+            match next {
+                Some(v) => {
+                    if let Some(domain) = domains.get(&v) {
+                        if !domain.is_empty() {
+                            colors.insert(v, domain[0]);
+                            propagate(&graph, v, &mut colors, &mut domains);
+                        }
+                    }
+                }
+                None => {
+                    // If we can't proceed, increase color count and restart
+                    max_colors += 1;
+                    continue;
+                }
+            }
+        }
+        break;
+    }
+
+    colors
+}


AnarchistHoneybun:
@starovoid when the guide says "make the algo work for generic graphs", is what I have so far satisfactory or does something more needs to be done?

starovoid:
> @starovoid when the guide says "make the algo work for generic graphs", is what I have so far satisfactory or does something more needs to be done?

All right. This just implies the use of traits instead of specific types.
I would only suggest checking what happens if you run the algorithm on a directed graph? Will it be interpreted as undirected, or will an error occur?

AnarchistHoneybun:
Running it on a sample directed graph produced wrong results. Like the graph is taken in fine and it gives a result, but my verification function marked it as invalid and I tested it by generating a dotgraph format output and visualizing too. What expected behaviour do we have in mind? 
based on [this reddit comment](https://www.reddit.com/r/learnmath/comments/1hpsb5h/comment/m4jwank/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) if I'm understanding it correct, we can change the directed graph into an undirected graph first, and then run our algo over it?

starovoid:
> Running it on a sample directed graph produced wrong results. Like the graph is taken in fine and it gives a result, but my verification function marked it as invalid and I tested it by generating a dotgraph format output and visualizing too. What expected behaviour do we have in mind? 
> based on [this reddit comment](https://www.reddit.com/r/learnmath/comments/1hpsb5h/comment/m4jwank/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) if I'm understanding it correct, we can change the directed graph into an undirected graph first, and then run our algo over it? 

I see two common options:
1. Return  `Result` from function, where the error will cover all invalid cases.
2. Modify the viewing of graph edges so that the graph is perceived as undirected. But won't it reduce performance for good input graphs?

I would prefer the second option if it does not reduce performance. Otherwise, the first one. In any case, the user will most likely want to pass an undirected graph to the function.
And, of course, both behaviors should be documented.

AnarchistHoneybun:
I think we should just go with the first approach. Performance is already an issue. The python implementation that I'm referencing completes all benchmark trials in the microsecond territory, whereas this one is barely contained inside one second.
The python approach has a lot of the graph parsing etc moved out of the coloring function, and I attempted the same. That gives us good improvements but again, we want to operate on the graph level and not have the user do the parsing and give us arrays of nodes and connections or whatever.

Is there a way to restrict the function to undirected graphs only?

starovoid:
> Is there a way to restrict the function to undirected graphs only?

You can use [`GraphProp`](https://docs.rs/petgraph/latest/petgraph/visit/trait.GraphProp.html) trait.

AnarchistHoneybun:
@starovoid lmk if something needs changing etc.

starovoid:
> @starovoid lmk if something needs changing etc.

Now we have to wait for the maintainers' review. But I'll look at the implementation in my free time too.

AnarchistHoneybun:
definitely think a few allocations could be avoided here and there but anytime i try and do that the performance sinks. At this point a fix will probably come to me in a dream.

starovoid:
@RaoulLuque   Please look, I plan to include it in the `0.8.3` release.

RaoulLuque:
Hey,
First off, thanks for the contribution :) The algorithm looks really good, especially because it is so fast. The only thing I am unsure about is why the paper has so little citations/does not seem to be published 🤔 The performance looks really good in any case though :)

This Weekend I'm not at home, so I won't be able to look at the code in detail until next week :) But I already have a question:
The paper also states this but I was wondering why one would have to restart the algorithm when an uncolored vertex is collapsed and has entropy 0. Couldn't we just increase the number of colors (M) and assign that new color to the vertex and then continue? I would imagine restarting the algorithm (with one more color) would result in exactly the same assignment or what am I missing?
This could potentially be a worthwhile optimization 🤔

AnarchistHoneybun:
I'll look into it

RaoulLuque:
Also I could also do this in the review (I don't mind either way :) ), but maybe if you want you could change the docstring to match the new unified docstrings for Algorithms as suggested here:
https://github.com/petgraph/petgraph/pull/815
Note that the initial suggestion at the top of the PR has slightly changed in the course of the discussions in the PR. Maybe just refer to the new state of DSatur on the master branch 🌞

AnarchistHoneybun:
I've sent an email to the authors to see if they've already pursued the restart vs continuing with M+1 colors issue, hopefully they'll have some insight.

starovoid:
Hi! CI fails for two reasons:
1. bad formatting (use `cargo fmt`);
2. using `std` that violates `no_std` support, try to fix imports with `core` or `alloc` crates.

AnarchistHoneybun:
I'll fix both soon. Can you check if the quickcheck file is fine though, I keep seeing errors throughout it in vscode that disappear when pressing Ctrl+S, and reappear after a while

RaoulLuque:
I think the linter/rust-analyzer is confused depending on whether one uses the quickcheck! macro. So maybe try using the other method of creating quickchecks like so:

```rust
#[cfg(feature = "stable_graph")]
#[test]
fn steiner_tree_spans_terminals() {
    fn prop(g: UnGraph<(), u32>) -> bool {
        if g.node_count() <= 1 {
            return true; // We naturally don't support steiner trees with zero or one node
        }

        // Run the steiner tree algorithm on connected components, to test it on both
        // connected and disconnected graphs.
        let mut connected_components = Vec::new();
        let mut visited = g.visit_map();

        for node in g.node_indices() {
            if !visited.is_visited(&node) {
                let mut component = HashSet::new();

                let mut dfs = Dfs::new(&g, node);
                while let Some(nx) = dfs.next(&g) {
                    visited.visit(nx);
                    component.insert(nx);
                }

                connected_components.push(component);
            }
        }

        for component in connected_components {
            if component.len() < 2 {
                continue; // We naturally don't support steiner trees with zero or one node
            } else {
                let g = g.filter_map(
                    |node_index, _| {
                        if component.contains(&node_index) {
                            Some(())
                        } else {
                            None
                        }
                    },
                    |edge_index, edge_weight| {
                        let edge = g.edge_endpoints(edge_index).unwrap();
                        if component.contains(&(edge.0)) && component.contains(&(edge.1)) {
                            Some(*edge_weight)
                        } else {
                            None
                        }
                    },
                );

                let terminals = g.node_indices().take(5).collect::<Vec<_>>();
                let m_steiner_tree = steiner_tree(&g, &terminals);

                let steiner_tree_nodes: Vec<NodeIndex> = m_steiner_tree.node_indices().collect();

                let spans_terminals = terminals.iter().all(|&t| steiner_tree_nodes.contains(&t));

                if !spans_terminals {
                    return false; // The steiner tree does not span all terminals
                }
            }
        }

        true
    }

    quickcheck::quickcheck(prop as fn(Graph<(), u32, Undirected>) -> bool);
}
```

Also, there seem to be some changes which look like there newly introduced, although they already exist in the main branch (quickchecks for steiner / johnson). Was there some issue with rebasing or why do they show up as changes? Could you please also fix that, I think it would be good if the git blame of those does not get switched up if there is no reason :thinking:

RaoulLuque:
> I've sent an email to the authors to see if they've already pursued the restart vs continuing with M+1 colors issue, hopefully they'll have some insight.

Okay, I have thought about it for a while and now understand why the restarting is necessary.

Suppose the algorithm is applied to the complete graph on 3 vertices (K_3), then initially all vertices have an entropy of 2. Then, the first vertex is assigned a color and this is propagated to the two other ones. Since they now have entropy 1 in the propagation step, they immediately get assigned the color. But this already constitutes an invalid coloring since all vertices are pairwise neighbors and two vertices now have the same color. So basically, this can happen due to the BFS in the propagation step.

Now I wonder, whether this might not happen if instead one used a DFS in the propagation step, that is, if that would speed up the algorithm since no restarts would be necessary. However, that is a different  question and definitely shouldn't be your concern.

Sry for bringing it up initially and in any case, thanks for writing them an E-Mail :)

AnarchistHoneybun:
I'm working on the mentioned tasks. as for the git blame issues caused by the botched merge, I have to admit it seems a bit beyond my git know-how. if you could guide me on it, or have some sources i can read etc to learn about it that'd be great

RaoulLuque:
Ah, I see, no worries :) I'm not an expert either, maybe @starovoid has a better idea. I would suggest the following:

Basically you have 2/3 merge commits in your PR which merge the changes from master into the branch which makes them seem like your changes. Instead you could rebase them. You can read up on the difference of the two online :) Most guides will be better than my explanation could be 😅

Now to use rebase instead of merge:
You could first backup the current branch as is locally (i.e create a copy of the folder). 
Then revert the current state of your fork to before the first merge commit. Then you could either rebase first or cherry pick the other commits from your PR which are not merge commits. Cherry pick is again a git command/term which you can lookup online :)

This way you can basically cherry pick (literally) together all your commits which are not merge commits, then rebase (or rebase before cherry picking) and then your changes/commits will go on top of what is already on the master branch.  You would then have to do git --force push to force these commits onto the current branch and override the current ones :)

I hope this helps, otherwise feel free to reach out. I also feel like LLMs can be quite helpful when dealing with these git things sometimes ^^

Just backup your current state before, so that you don't loose your changes :)

AnarchistHoneybun:
i'll look that up. although if we're not totally married to the idea of keeping this specific PR open I could scrap it and open another one. that way would probably be a lot faster, at the cost of not keeping all discussions on this algo in one place

RaoulLuque:
As you wish :) In that case you could also copy the changes over to a clean slate branch which is up to date and then force push to the remote/origin of this branch :) LLMs should also be able to help you with the latter 😁 Then you would combine the not having to deal with the commits and stuff and we would still have this PR 👍

AnarchistHoneybun:
i think I've tackled everything mentioned? i'll investigate the bfs v dfs thing mentioned, but apart from that it all looks good

RaoulLuque:
Yeah, thanks very much for all the work, rebasing everything and adressing the comments :) Coincidentally I've got time to look over the things right now, so I'll see until where I can get :relaxed:

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
+use crate::algo::vec;
+use crate::algo::Vec;
+use crate::alloc::collections::vec_deque::VecDeque;
+use crate::alloc::string::String;
+use crate::alloc::string::ToString;
+use crate::visit::{GraphProp, IntoNeighbors, NodeCount, NodeIndexable, Visitable};
+use core::hash::Hash;
+use hashbrown::HashMap;
+
+/// \[Generic\] [Wave Function Collapse algorithm][1] to properly color a non-weighted undirected graph.
```
RaoulLuque, line 32: That's probably something I don't quite understand, but why does the graph need to be non-weighted?

RaoulLuque, line 32:
```suggestion
/// \[Generic\] [Wave Function Collapse Coloring algorithm][1] to properly color a non-weighted undirected graph.
```
The double coloring is maybe a bit redundant, but this way we would give it the same name as the authors from the paper :) It don't mind either way :+1:

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
...
+        .max()
+        .unwrap_or(0);
+    let colors = max_degree + 1;
+
+    // Create and run WFC state
+    let mut wfc_state = WfcState::new(node_count, colors, connections);
+    let result = wfc_state.run()?;
+
+    // Convert result to hashmap
+    let mut color_map = HashMap::new();
```
starovoid:
```suggestion
    let mut color_map = HashMap::with_capacity(node_count);
```

This should slightly improve performance.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    Ok(color_map)
+}
+
+#[derive(Debug)]
+struct WfcState {
+    nodes: usize,
+    colors: usize,
+    connections: Vec<Vec<bool>>,
+    available_colors: Vec<Vec<bool>>,
```
starovoid: Try using [`FixedBitSet`](https://docs.rs/fixedbitset/latest/fixedbitset/struct.FixedBitSet.html) to save memory if it doesn't hurt performance. Maybe it is also possible to use the adjacency matrix obtained from the [`GetAdjacencyMatrix`](https://docs.rs/petgraph/latest/petgraph/visit/trait.GetAdjacencyMatrix.html#associatedtype.AdjMatrix) trait.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        Ok(())
+    }
+
+    fn run(&mut self) -> Result<Vec<isize>, String> {
+        while !self.finished {
+            self.restart_flag = false;
+            self.find_lowest_entropy();
+
+            if let Some(index) = self.min_index {
```
RaoulLuque: Why is `min_index` a field in the wfc state struct? Why do we not have it as a local variable here and return it with `find_lowest_entropy` ?
It looks to me like there is no other usage of min_index except for here, so we could also just have a local variable for that?

Same with `restart_flag` if I'm not mistaken :thinking:

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
...
+/// assert_ne!(coloring[&b], coloring[&c]); // Adjacent vertices have different colors
+/// assert_ne!(coloring[&c], coloring[&a]); // Adjacent vertices have different colors
+/// ```
+pub fn wfc_coloring<G>(graph: G) -> Result<HashMap<G::NodeId, usize>, String>
+where
+    G: IntoNeighbors + NodeCount + NodeIndexable + Visitable + GraphProp,
+    G::NodeId: Eq + Hash + Copy,
+{
+    if graph.is_directed() {
+        return Err("Graph must be undirected".into());
```
starovoid: The `String` error is difficult to handle on the user's side. Please make the `Err` variant a structure or enum.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    fn collapse(&mut self, index: usize) -> Result<(), String> {
+        if self.finished {
+            return Ok(());
+        }
+        if self.entropy[index] == 0 {
+            return Err("Impossible pattern".to_string());
+        }
+
+        self.entropy[index] = usize::MAX;
```
RaoulLuque: To me it would seem preferable (and possibly more Rust idiomatic?) to use a `Vec<Option<usize>> for `self.entropy`. Then one could use `None` instead of `usize::MAX` to indicate that a vertex has already gotten a color assigned to it.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
...
+        }
+    }
+
+    fn collapse(&mut self, index: usize) -> Result<(), String> {
+        if self.finished {
+            return Ok(());
+        }
+        if self.entropy[index] == 0 {
+            return Err("Impossible pattern".to_string());
+        }
```
RaoulLuque:
```suggestion
```
You use Errors/Results at several places in the code. I am really not that aqcuianted with Rust errors, so I might be wrong about some things. However, for this one for example, I would say that it should not occur, right?
That is, we can just omit this if check and increase performance. Since, this function is only called, if `find_lowest_entropy` actually found a node.
This might also be applicable to other Errors that are raised.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,247 @@
...
+
+        self.entropy[index] = usize::MAX;
+        self.affected_nodes.push_back(index);
+
+        let color_index = self.available_colors[index]
+            .iter()
+            .position(|&x| x)
```
RaoulLuque:
```suggestion
            .next()
```
If you just want an element in the iterator/to check whether one exists, you could use none instead of position. Not sure if there is a difference if the predicate in position is basically true, but seems more idiomatic :)

RaoulLuque (commented):
Again, thanks for the good work implementing the algorithm :)

The general structure is really nice and easy to understand :sun_with_face: 
Also, using a state struct seems like a good approach :) I didn't have enough time to look at everything, but this is the first part I guess and the rest will follow in the course of the next week :+1: 

I commented some small things and one thing about the errors/results you return in the method. I am not sure how we should actually handle those.

The one for directed graphs makes sense in my opinion, but in my opinion at least one of the other errors should actually never occur if the functions are only used in the context of the wfc-c algorithm :thinking:

src/algo/coloring/wfc.rs
RaoulLuque: Wouldn't it make sense here to have a "global" petgraph error type for errors like this one? Where an undirected graph was supposed to be provided and it is actually directed / other way around ?

starovoid: In this case, it is not entirely clear how to separate general errors from those specific to a particular algorithms/types/etc.

AnarchistHoneybun: I'm assuming we're to replace all string errors with some form of this enum thing, right?

src/algo/coloring/wfc.rs
AnarchistHoneybun: I don't think it necessarily has to be. I wrote up this implementation a while back (when I was mostly figuring out how traits etc worked) so a lot of the stuff directly referenced the dsatur implementation. + the paper makes no mention of weights so I thought to make it explicit. afaik this only goes as far as the docstring, so we can amend it

src/algo/coloring/wfc.rs
starovoid: I suggest something like:
```rust
pub enum WfcColoringError {
    Error1, // Replace with meaningful naming
    Error2,
    ....
}
```

starovoid:
Using `AdjMatrix` unexpectedly significantly reduced performance, I rolled back this change in the local branch and the previous performance returned. I will ask you to return the old implementation of `connections'. I'm sorry for that bad idea :(

tests/coloring.rs
```diff
@@ -57,3 +57,95 @@ fn dsatur_coloring_bipartite() {
...
+fn wfc_coloring_random_graph() {
+    let mut graph: Graph<(), (), Undirected> = Graph::new_undirected();
+    let a = graph.add_node(());
+    let b = graph.add_node(());
+    let c = graph.add_node(());
+    let d = graph.add_node(());
+    let e = graph.add_node(());
+    let f = graph.add_node(());
+
+    graph.extend_with_edges(&[
```
starovoid:
```suggestion
    graph.extend_with_edges([
```

tests/coloring.rs
```diff
@@ -57,3 +57,95 @@ fn dsatur_coloring_bipartite() {
...
+    let a = graph.add_node(());
+    let b = graph.add_node(());
+    let c = graph.add_node(());
+    let d = graph.add_node(());
+    let e = graph.add_node(());
+    let f = graph.add_node(());
+    let g = graph.add_node(());
+    let h = graph.add_node(());
+
+    graph.extend_with_edges(&[
```
starovoid:
```suggestion
    graph.extend_with_edges([
```

tests/coloring.rs
```diff
@@ -57,3 +57,95 @@ fn dsatur_coloring_bipartite() {
...
+
+#[test]
+fn wfc_coloring_directed() {
+    let mut graph: Graph<(), (), petgraph::Directed> = Graph::new();
+    let a = graph.add_node(());
+    let b = graph.add_node(());
+    let c = graph.add_node(());
+    let d = graph.add_node(());
+
+    graph.extend_with_edges(&[(a, b), (b, c), (c, d), (d, a), (a, c)]);
```
starovoid:
```suggestion
    graph.extend_with_edges([(a, b), (b, c), (c, d), (d, a), (a, c)]);
```

starovoid (changes requested):
A few annoying `clippy` lints.

AnarchistHoneybun:
can't believe both of us decided to do it in the same minute lol

src/algo/coloring/wfc.rs
RaoulLuque: > In this case, it is not entirely clear how to separate general errors from those specific to a particular algorithms/types/etc.

Yeah, sry for the very late reply, but after thinking about it some more, I agree with you :) Maybe this could be something one could think about in a separate issue/PR someday ^^? But also maybe not. I'll open another one if I still think it's worthwhile after some time

starovoid: Perhaps we could export the general error under the "thiserror" or "anyhow" features, but this seems superfluous to me.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+            Some(index) => EntropyResult::Found(index),
+            None => EntropyResult::Finished,
+        }
+    }
+
+    fn collapse(&mut self, index: usize) -> Result<(), WfcColoringError> {
+        if self.finished {
+            return Ok(());
+        }
+
```
RaoulLuque:
```suggestion
```
Couldn't this check be omitted? It seems to be redundant, since we only call `collapse` if `find_lowest_entropy` actually found a node in which case self.finished should be false.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+        self.available_colors[index].clear();
+        self.available_colors[index].set(color_index, true);
+        self.output[index] = color_index as isize;
+
+        Ok(())
+    }
+
+    fn propagate(&mut self) -> Result<(), WfcColoringError> {
+        let mut visited = vec![false; self.nodes];
```
RaoulLuque: This could be replaced by a [`VisitMap`](https://docs.rs/petgraph/latest/petgraph/visit/trait.VisitMap.html), which should be more efficient.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+                        self.entropy[node_index] = Some(current_entropy - 1);
+
+                        if self.entropy[node_index] == Some(0) {
+                            return Err(WfcColoringError::NoValidConfiguration);
+                        }
+                        if self.entropy[node_index] == Some(1) && !visited[node_index] {
+                            visited[node_index] = true;
+                            self.affected_nodes.push_back(node_index);
+                        }
+                    }
```
RaoulLuque:
```suggestion
                      let current_entropy) = self.entropy[node_index].expect("Entropy should be some according to if block condition");
                      self.entropy[node_index] = Some(current_entropy - 1);
                      
                      if self.entropy[node_index] == Some(1) && !visited[node_index] {
                          visited[node_index] = true;
                          self.affected_nodes.push_back(node_index);
                      }
```
You could remove this `if let` Block since this will always be entered by the `self.entropy[node_index].is_some()` condition. Instead you could use an `.expect()` on `self.entropy[node_index]` to obtain the current entropy.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+
+        while let Some(index) = self.affected_nodes.pop_front() {
+            let color_index = self.available_colors[index]
+                .ones()
+                .next()
+                .ok_or(WfcColoringError::NoValidConfiguration)?;
+
+            #[allow(clippy::needless_range_loop)]
+            for node_index in 0..self.nodes {
+                if self.connections[index * self.nodes + node_index]
```
RaoulLuque, line 249: I guess this is going to be rolled back to the none adjacency matrix version? This right here might actually be the cause of the performance drop. We could use the graph adjacency here and iterate over the neighbours of the nodes possibly to speed this up? Might result in the adj matrix being faster after all. Not sure tho. By me, we could also just rollback to before adj matrix, just an idea ^^

RaoulLuque (commented):
Okay, first off, once again thanks for the PR and all the effort so far :)

This is only part of my review since while I was reviewing the code and algo I noticed another seemingly important thing. The algorithm as described in the original paper seems to be wrong. This can be checked with the complete graph on 3 vertices which will receive an invalid coloring.

It will never enter the main while loop in Algorithm 4 in the paper and instead terminate after the first `propagate(u)` step. In this step, the two vertices which are not u will receive the same color, since both of them have the initial entropy of 2.

However, this implementation seems to mitigate that issue by not assigning colors in the propagate step. Doing so would need additional checks I guess. I would however suggest to add a note to the Docstring of the alg addressing this issue in the original paper?

Also, part of my review was trying to get rid of all the errors which are returned by the algo. Because I think most of them cannot/should not actually occur. The only error that should be able to occur is the `WfcColoringError::DirectedGraph` I think.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    fn propagate(&mut self) -> Result<(), WfcColoringError> {
+        let mut visited = vec![false; self.nodes];
+
+        while let Some(index) = self.affected_nodes.pop_front() {
+            let color_index = self.available_colors[index]
+                .ones()
+                .next()
+                .ok_or(WfcColoringError::NoValidConfiguration)?;
+
```
RaoulLuque, line 243:
```suggestion

self.output[index] = color_index as isize;
self.entropy[index] = None;

```
We assign the color here to speed up the entire algorithm without breaking anything (hopefully - feel free to double check this), since we do this and then adjust of the neighbors immediately (I hope that explanation makes sense).

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+                    && self.available_colors[node_index].contains(color_index)
+                {
+                    self.available_colors[node_index].set(color_index, false);
+
+                    if let Some(current_entropy) = self.entropy[node_index] {
+                        self.entropy[node_index] = Some(current_entropy - 1);
+
+                        if self.entropy[node_index] == Some(0) {
+                            return Err(WfcColoringError::NoValidConfiguration);
+                        }
```
RaoulLuque: If the entropy of a node is 0 we should restart the algorithm instead of throwing an error. This should just be recoverable by incrementing the number of possible colors :thinking:

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+            #[allow(clippy::needless_range_loop)]
+            for node_index in 0..self.nodes {
+                if self.connections[index * self.nodes + node_index]
+                    && self.entropy[node_index].is_some()
+                    && self.available_colors[node_index].contains(color_index)
+                {
+                    self.available_colors[node_index].set(color_index, false);
+
+                    if let Some(current_entropy) = self.entropy[node_index] {
+                        self.entropy[node_index] = Some(current_entropy - 1);
```
RaoulLuque:
```suggestion
                        self.entropy[node_index] = Some(current_entropy.saturating_sub(1));
```
Make sure not to underflow here

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    fn propagate(&mut self) -> Result<(), WfcColoringError> {
+        let mut visited = vec![false; self.nodes];
+
+        while let Some(index) = self.affected_nodes.pop_front() {
+            let color_index = self.available_colors[index]
+                .ones()
+                .next()
+                .ok_or(WfcColoringError::NoValidConfiguration)?;
```
RaoulLuque, line 242:
```suggestion
            let color_index = self.available_colors[index]
                .ones()
                .next()
                .expect("A color should be available, since otherwise entropy should be 0 and we would have restarted the algorithm.");
```
Since we always check if the entropy is 0, we should be able to unwrap here without any problems.

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+#[derive(Debug)]
+struct WfcState {
+    nodes: usize,
+    colors: usize,
+    connections: FixedBitSet,
+    available_colors: Vec<FixedBitSet>,
+    entropy: Vec<Option<usize>>,
+    output: Vec<isize>,
```
RaoulLuque, line 152: Just noticed this but why is this isize and not usize or any other unsigned integer type?

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,287 @@
...
+            return Ok(());
+        }
+
+        self.entropy[index] = None;
+        self.affected_nodes.push_back(index);
+
+        let color_index = self.available_colors[index]
+            .ones()
+            .next()
+            .ok_or(WfcColoringError::NoValidConfiguration)?;
```
RaoulLuque:
```suggestion
        let color_index = self.available_colors[index]
            .ones()
            .next()
            .expect("A color should be available, since otherwise entropy should be 0 and we would have restarted the algorithm.");
```
This should be able to be unwrapped with the same reasoning as in `propagate`.

tests/quickcheck.rs
```diff
@@ -1706,3 +1706,18 @@ quickcheck! {
...
+    fn wfc_coloring_quickcheck(g: Graph<(), (), Undirected>) -> bool {
+        match wfc_coloring(&g) {
+            Ok(coloring) => {
+                assert!(
+                    is_proper_coloring(&g, &coloring),
+                    "wfc_coloring returned a non proper coloring"
+                );
+                true
+            },
+            Err(_) => true // WFC can legitimately fail on some graphs, unlike DSatur
```
RaoulLuque:
```suggestion
            // The algorithm should not fail on undirected graphs
            Err(_) => false
```

AnarchistHoneybun:
I'm away from my system for a while, so i can get to these changes in the evening (ist time), at the earliest. hope that's cool

RaoulLuque:
Yeah, no worries at all :) Thanks for letting us know 🌞

src/algo/coloring/wfc.rs
AnarchistHoneybun: iirc we had a similar loop pre-adj matrix version, and I think that change was already rolled back as well. even so, will check again

AnarchistHoneybun:
all of the stylistic and similar changes have been done. algorithm level changes I'm still tinkering with as some stuff is breaking, but it'll be done soon too

RaoulLuque:
Alright, thanks for the heads up ! :sun_with_face: 

Let me know if there is anything you're unsure with :+1: 

Also, FYI as stated above, you can easily accept suggestions here in GitHub. That way the reviewers are also credited for their contributions :)

AnarchistHoneybun:
oh I didn't know that, I thought it'd be the same as making the change locally and pushing it and I prefer that since making non local changes and pulling sometimes ends with them getting out of sync etc, but I'll keep that in mind

src/algo/coloring/wfc.rs
AnarchistHoneybun: we have sentinel values for unassigned colors, at the start. though we could make it more idiomatic by using Option<usize>, i suppose. thoughts on that?

src/algo/coloring/wfc.rs
AnarchistHoneybun: opted for a bitset similar to how the other boolean arrays were replaced, if the performance gains are unsatisfactory i'll move to visitmap next

AnarchistHoneybun:
most of the stuff's done, i think the only one that remains is the isize->result(usize) for output and the bitset instead of visitmap, which I'll need some adjudication on before deciding on one or the other.
that said, the restart instead of fail change in propagate makes me a little uneasy, and I think we should add more tests (though what kind of tests is a bit beyond me) to cover our bases

RaoulLuque:
Alright, thanks again for addressing the things from my review :) 

I won't have time to look over the newest version today, but I think I'll get to it sometime on the weekend :sun_with_face: 

Regarding the tests:
I agree with you that there is some unease with unwrapping things instead of returning an error to the user. However, I find that way, way more useful. Since, returning an error would somehow suggest that user made an error, but if instead the algorithm is faulty that only leads to confusion. So, if the library panics, we know that there is an error in the algorithm.

With regard to that, if we remove all "return error sources" except for the "Wrong Graph Type" one, quickcheck should take care of most of the checking for us :relaxed: 

From glancing over the current version I saw that propagate still returns a result. So maybe try removing that error as well and instead use `.expect()` and explain in the expect why this should be safe :)

Also, regarding the way to restart, I would suggest to move the restarting logic into the main loop :+1:  In my opinion, this would make the logic of the algorithm clearer. Instead of restarting the entire algorithm somewhere in some method (actually two in this case), it is immediately visible when looking at the main loop and the methods don't have any drastic "side effects" if that makes sense.

To address is, we could return a Restart from the `find_lowest_entropy` and then in the main loop, in the respective match case, call `restart_wfc` (instead of calling `restart_wfc` in the `find_lowest_entropy`and then returning).Similarly, propagate could return an enum of sorts to signal to the main loop, that it needs a restart. Checking for that, the main loop could also call `restart_wfc` in the appropriate case.

src/algo/coloring/wfc.rs
RaoulLuque: Ah, I see. Well there is an argument to be made there I guess, but I suppose we would assume that the algorithm correctly assigns each vertex a color, right? In that case, we just assign 0 first to all vertices and then we know the algorithm only terminates once every vertex has been assigned a color (entropy = None for all vertices)

src/algo/coloring/wfc.rs
```diff
@@ -0,0 +1,288 @@
...
+pub fn wfc_coloring<G>(graph: G) -> Result<HashMap<G::NodeId, usize>, WfcColoringError>
+where
+    G: IntoNeighbors + NodeCount + NodeIndexable + Visitable + GraphProp,
+    G::NodeId: Eq + Hash + Copy,
+{
+    if graph.is_directed() {
+        return Err(WfcColoringError::DirectedGraph);
+    }
+
+    let node_count = graph.node_count();
```
RaoulLuque:
```suggestion
    let node_count = graph.node_bound();
```
I just noticed this, but you can not assume that you can create a vertex with all vertices by using the node_count. Instead you need to use `node_bound`, since the indices used for the nodes might exceed node_count - 1 (in case of StableGraphs).

RaoulLuque:
The above is not actually the entire review instead I just ignored the things I addressed in the [comment above](https://github.com/petgraph/petgraph/pull/709#issuecomment-2992008279) the review. I just noticed that while glancing over another time ^^

Do you think you'll be able to incorporate the ideas I am suggesting in the [comment](https://github.com/petgraph/petgraph/pull/709#issuecomment-2992008279) or should I put them in a review :)?

RaoulLuque:
*just saw the commit, thanks :+1: One more thing tho  :sweat_smile: Didn't think about that earlier: Maybe also rename the `node_count` variable to `node_bound` or smth to make it clearer since it now does not track the actually node count anymore :)

AnarchistHoneybun:
uhhh the larger change botched the tests, did a git reset to bring it back to an older state, lemme see what's wrong with that

starovoid:
Hello @AnarchistHoneybun, there is a conflict with the file `coloring.rs `. Please fix it, I plan to merge this PR.

But before concluding the PR discussion, I would like to understand why the current implementation is so much inferior to DSatur in performance, while in the paper the WFC-C algorithm is many times faster.

```text
test dsatur_coloring_bench ... bench:   7,207,114.70 ns/iter (+/- 131,760.09)
test wfc_coloring_bench    ... bench: 146,609,748.80 ns/iter (+/- 7,193,834.55)
```

AnarchistHoneybun:
I'll look into the perf issue

RaoulLuque:
> Hello @AnarchistHoneybun, there is a conflict with the file `coloring.rs `. Please fix it, I plan to merge this PR.
> 
> But before concluding the PR discussion, I would like to understand why the current implementation is so much inferior to DSatur in performance, while in the paper the WFC-C algorithm is many times faster.
> 
> ```
> test dsatur_coloring_bench ... bench:   7,207,114.70 ns/iter (+/- 131,760.09)
> test wfc_coloring_bench    ... bench: 146,609,748.80 ns/iter (+/- 7,193,834.55)
> ```

I suspect that the difference in runtime has something to do with the flaw I pointed out in the paper, see here:
https://github.com/petgraph/petgraph/pull/709#pullrequestreview-2932826006

Possibly, the authors were not aware of this and ran the algorithm as it is described in the paper - computing wrong colorings, but with very good performance. I am afraid that balancing the correctness and performance of this algorithm might not work out, note sure tho. Would be very sad, as a lot of work has already been put into this

starovoid:
> Possibly, the authors were not aware of this and ran the algorithm as it is described in the paper - computing wrong colorings, but with very good performance. I am afraid that balancing the correctness and performance of this algorithm might not work out, note sure tho. Would be very sad, as a lot of work has already been put into this

The algorithm is initially positioned as a heuristic, so an error is acceptable only if it does not tend to 100% :)
I found these accuracy comparison in the article:
> As for DSatur, WFC-C is 9.22% more optimal and in the case of latin square, WFC-C was 38.03% more optimal.

Is it known what accuracy the current (slow) implementation has?

--- 
Let it be a draft until we choose a compromise option. I would be happy to accept this algorithm in one form or another, as long as it beats the existing DSatur in at least one parameter.

RaoulLuque:
> > Possibly, the authors were not aware of this and ran the algorithm as it is described in the paper - computing wrong colorings, but with very good performance. I am afraid that balancing the correctness and performance of this algorithm might not work out, note sure tho. Would be very sad, as a lot of work has already been put into this
> 
> The algorithm is initially positioned as a heuristic, so an error is acceptable only if it does not tend to 100% :) I found these accuracy comparison in the article:
> 
> > As for DSatur, WFC-C is 9.22% more optimal and in the case of latin square, WFC-C was 38.03% more optimal.
> 
> Is it known what accuracy the current (slow) implementation has?
> 
> Let it be a draft until we choose a compromise option. I would be happy to accept this algorithm in one form or another, as long as it beats the existing DSatur in at least one parameter.

Actually, I am not talking about the accuracy / comptetitiveness of the algorithm. I was referring to the fact that the heuristic sometimes returns a coloring which is infeasible, which does not happen to DSatur afaik. I agree that a heuristic mustn't always return an optimal result but I would've thought that at it least returns a feasible solution (if one exists and the algorithm terminates). But I guess that does not necessarily hold either :thinking:

In that case, it is probably worth it to go back to the original design as the authors intended and just return an Option and return None if the algorithm does not find a feasible solution? In that case however, it would be necessary to implement a feasibility check of sorts, which would decrease the runtime. Or one would add a note saying that the algorithm may not always produce valid colorings, but personally, I wouldn't like that.

LightenedLimited:
Hi folks! Just stumbled across this, I'm one of the authors of the original algorithm. I realized I'm a bit late here, but wanted to give some insights into both the paper and the algorithm itself :)
* **Why doesn't it always return correct colorings?** One design decision made for the algorithm itself was trading a guarantee of accuracy for a massive gain in average computational time and slightly superior results. By itself, there's many such use cases where this is a viable tradeoff (e.g, game generation, social media graph coloring, etc.). The publicly available paper doesn't contain the most up-to-date analysis on the algorithm unfortunately (there is a private version, which I am happy to discuss). The original algorithm was observed from one such use case. One of my colleagues did modify this algorithm slightly to have a guarantee of correctness, I'm happy to ask him if it's of interest. 
* **What the heck is up with the performance?** You are indeed meant to assigning colors in the propagate, and this should drastically speed up the performance.

PR #714, opened 2025-01-01
ci: downgrade hashbrown rather than limiting indexmap

cuviper:
Pinning `indexmap = "~2.5.0"` for MSRV prevents users with more current compilers from using the latest petgraph and indexmap together, even though they would otherwise work fine. Instead, dependencies can just be downgraded as needed for CI testing, and users who need the older compiler support can also manage this in their own lock files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,11 @@ jobs:
           profile: minimal
           toolchain: ${{ matrix.rust }}
           override: true
+      - name: MSRV downgrade
+        if: matrix.rust == '1.64.0'
+        run: |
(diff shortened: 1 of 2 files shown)


XVilka (approved):
Thanks for the fix!

cuviper:
Thanks for merging! Is there any plan for a point release so this can reach crates.io?

XVilka:
@cuviper yes, we could. But I am afk for a couple weeks. @indietyp or @ABorgna  could probably help with this meanwhile.

indietyp:
Sure I can create a release tomorrow

PR #716, opened 2025-01-02
Bugfix: Allow use of 2.7.0 Indexmap for people running newer versions of rust

pvichivanives:
This lets indexmap V2.7 work with people running newer versions of rust. I didn't bump the MSRV but instead added a lockfile-change to make 1.64 work, let me know if you would rather bump msrv.

Talked to the maintainer of indexmap here: https://github.com/indexmap-rs/indexmap/issues/366 and this seems to be a good way to support both old and new version users.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,11 @@ jobs:
           profile: minimal
           toolchain: ${{ matrix.rust }}
           override: true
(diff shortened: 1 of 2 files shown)


cuviper:
I've done basically the same in #714.

pvichivanives:
Thanks, somehow missed it!

PR #718, opened 2025-01-04
feat: Add methods with recoverable errors for `Graph` and `StableGraph`

starovoid:
## Methods with recoverable errors
This PR provides `try_add_node`, `try_add_edge` and `try_update_edge` for `Graph` and `StableGraph`.

The `GraphError` type has been added to represent possible errors:
```rust
/// The error type for fallible `Graph` & `StableGraph` operations.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum GraphError {
    /// The Graph is at the maximum number of nodes for its index.
    NodeIxLimit,

    /// The Graph is at the maximum number of edges for its index.
    EdgeIxLimit,

    /// The node with the specified index is missing from the graph.
    NodeMissed(usize),

    /// Node indices out of bounds.
    NodeOutBounds,
}
```
To represent potential errors in `MatrixGraph` and other storages some other Errors will be needed due to the significant differences in provided APIs.

## Rewrite fallible methods with its `try` counterparts
To avoid code duplication, I rewrote `add_node`, `add_edge` and `update_edge` for both `Graph` and `StableGraph` using `try` counterparts. 
This changed the set of possible panics, but I tried to keep the custom panic messages:
```rust
panic!(
    "StableGraph::add_edge: node index {} is not a node in the graph",
    i
);

panic!("Graph::add_edge: node indices out of bounds")
```

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -558,15 +558,7 @@ where
     /// **Panics** if the Graph is at the maximum number of nodes for its index
     /// type (N/A if usize).
     pub fn add_node(&mut self, weight: N) -> NodeIndex<Ix> {
-        let node = Node {
-            weight,
-            next: [EdgeIndex::end(), EdgeIndex::end()],
-        };
-        let node_idx = NodeIndex::new(self.nodes.len());
-        // check for max capacity, except if we use usize
-        assert!(<Ix as IndexType>::max().index() == !0 || NodeIndex::end() != node_idx);
-        self.nodes.push(node);
-        node_idx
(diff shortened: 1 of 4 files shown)


notxvilka (approved):
In some future release we should make this behavior the default in `add_edge()`, not `try_add_edge()` in my opinion.

starovoid:
> In some future release we should make this behavior the default in `add_edge()`, not `try_add_edge()` in my opinion.

I think this will mostly resolve itself when switching to the `Storage` API #563.

PR #719, opened 2025-01-04
Implement `Csr::try_add_edge`

starovoid:
The `Csr::try_add_edge` function is implemented, which checks the given indexes and return possible error of type 
```rust
/// The error type for fallible operations with `Csr`.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum CsrError {
    /// Both vertex indexes go outside the graph.
    IndicesOutBounds(usize, usize),
}
```

It is currently impractical to create checked counterparts of other `Csr` methods, as this will require significant code duplication. It's easier to make breaking changes in one of the further major releases, moving away from panic to `Error` and `Option` in the entire `Csr` API.

### DIFF
--- a/src/csr.rs
+++ b/src/csr.rs
@@ -309,6 +309,7 @@ where
 
     /// Try to add an edge from `a` to `b` to the `Csr`, with its associated
     /// data weight.
+    ///
     /// Return `true` if the edge was added
     ///
     /// If you add all edges in row-major order, the time complexity
--- a/src/csr.rs
+++ b/src/csr.rs
@@ -1,6 +1,7 @@
 //! Compressed Sparse Row (CSR) is a sparse adjacency matrix graph.
 
 use std::cmp::{max, Ordering};
+use std::fmt;
 use std::iter::{Enumerate, Zip};
 use std::marker::PhantomData;
 use std::ops::{Index, IndexMut, Range};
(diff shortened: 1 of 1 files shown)


ABorgna (commented):
Nice.

I realize we don't do it anywhere else, but should we derive `std::error::Error` for `CsrError`. (And similarly in your other PRs).

ABorgna:
> It is currently impractical to create checked counterparts of other Csr methods, as this will require significant code duplication. It's easier to make breaking changes in one of the further major releases, moving away from panic to Error and Option in the entire Csr API.

Agreed. Mind adding a new issue, so we can track it in the [milestone](https://github.com/petgraph/petgraph/milestone/6)?

starovoid:
> I realize we don't do it anywhere else, but should we derive `std::error::Error` for `CsrError`. (And similarly in your other PRs).

I think we shouldn't do this, since `CsrError` is already described using `Display` and `Debug`, for a combination of which the trait `std::error::Error` is implemented by default.

ABorgna:
I don't think it is?
```rust
let err: &dyn std::error::Error = &CsrError::IndicesOutBounds(0, 0);
```
```
the trait bound `CsrError: StdError` is not satisfied
```

(btw, the tuple-struct error could use some named fields instead :) )

starovoid:
> Agreed. Mind adding a new issue, so we can track it in the [milestone](https://github.com/petgraph/petgraph/milestone/6)?

Good idea, opened #724

starovoid:
> I don't think it is?
> 
> ```rust
> let err: &dyn std::error::Error = &CsrError::IndicesOutBounds(0, 0);
> ```
> 
> ```
> the trait bound `CsrError: StdError` is not satisfied
> ```

Oh, I'll add such blancked implementation:
```rust
impl std::error::Error for CsrError {}
```

---


> (btw, the tuple-struct error could use some named fields instead :) )

In this particular error, both tuple fields are equivalent, and the naming options `node_ix_1` and `node_ix_2` seem too verbose. I am glad for suggestions if there is a better naming option for the fields.

ABorgna (approved):

PR #730, opened 2025-02-06
Add checked `UnionFind` methods

starovoid:
PR provides the following `UnionFind` methods, that are free from panics:

```rust
pub fn try_find(&self, mut x: K) -> Option<K>

pub fn try_find_mut(&mut self, x: K) -> Option<K>

pub fn try_equiv(&self, x: K, y: K) -> Result<bool, K>

pub fn try_union(&mut self, x: K, y: K) -> Result<bool, K>
```

Some of old methods were rewritten using the `try` twins with the addition of `.unwrap` or `.expect`, for example:
```rust
pub fn find(&self, x: K) -> K {
    self.try_find(x).expect("The index is out of bounds")
}
```
If you consider these changes to be critical & breaking, I will return back the old `assert`ions.

I'm open to discussion!

### DIFF
--- a/src/unionfind.rs
+++ b/src/unionfind.rs
@@ -95,19 +95,25 @@ where
     ///
     /// **Panics** if `x` is out of bounds.
     pub fn find(&self, x: K) -> K {
-        assert!(x.index() < self.len());
-        unsafe {
-            let mut x = x;
-            loop {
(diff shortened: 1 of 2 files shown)


ABorgna (approved):
Thanks!

ABorgna:
Error messages are not stable, changing them is not a semver break.

Issue #723, opened 2025-01-27
Adjust find_negative_cycle Implementation


paulatsydney:
Hi guys, appreciate if you have any idea how to implement this:

In the source code of bellman_ford.rs -> fn find_negative_cycle, I want to adjust two parts (of course I copied it out to my own crate):
1. If the sum value is less than -0.1 (rather than 0), treat it as a negative cycle
I found the below source code, want to change the IF condition, idea is to change it to "if distance[ix(i)] + w + 0.1 < distance[ix(j)]", but the compiling error doesn't allow me to add f64 to FloatMeasure, how can I achieve it?
-----------------------------------------------------------------
// Step 2: Check for negative weight cycle
    'outer: for i in g.node_identifiers() {
        for edge in g.edges(i) {
            let j = edge.target();
            let w = *edge.weight();
            if distance[ix(i)] + w < distance[ix(j)] {
                // Step 3: negative cycle found
-----------------------------------------------------------------
2. Although source node is passed into this function, but in some case, it will return the negative cycle that doesn't contain the source node. How to avoid that?

Thank you so much even you just provide some thoughts.

starovoid:
Hi, in PR #733, I add two methods for `FloatMeasure` trait, that now you can write something like this:
```rust
if distance[ix(i)] + w + G::EdgeWeight::from_f32(0.1) < distance[ix(j)]
```

But these changes won't be accepted quickly, so I advise you not to wait and just clone them into your own code.

starovoid:
Hello @paulatsydney , changes necessary for this task have been accepted in #733.

PR #733, opened 2025-02-07
feat!: Add `from_f32/64` methods for `Float`, `Unit`, and `Bounded` measures

starovoid:
Add two required methods for `FloatMeasure`, `BoundedMeasure` and `UnitMeasure` traits:
```rust
/// A floating-point measure.
pub trait FloatMeasure: Measure + Copy {
    fn zero() -> Self;
    fn infinite() -> Self;

    // NEW:
    fn from_f32(val: f32) -> Self;
    fn from_f64(val: f64) -> Self;
}
```

### Motivation
This will add flexibility in programming operations with edge weights and make possible the behavior requested in issue #723.

---
BREAKING CHANGE:

This PR adds new required methods to the public trait

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -734,6 +734,8 @@ impl<M> Measure for M where M: Debug + PartialOrd + Add<M, Output = M> + Default
 pub trait FloatMeasure: Measure + Copy {
     fn zero() -> Self;
     fn infinite() -> Self;
+    fn from_f32(val: f32) -> Self;
+    fn from_f64(val: f64) -> Self;
 }
 
 impl FloatMeasure for f32 {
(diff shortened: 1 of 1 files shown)


starovoid:
I think it would be nice to add the same methods also for `UnitMeasure`, `PositiveMeasure` and maybe for `BoundedMeasure` bedore release `0.8.0`. It may also be worth keeping only the `from_f64` method, but I still prefer to have two options for better integration with algebra libraries (such as `nalgebra`) in the future.
The `Measure` trait looks too general and in my opinion does not have to be a real number semantically, so we should not implement `from_f64` and `from_f32` for it.

But whatever my reasoning, I would like to see the opinion of colleagues and the maintainers on this matter. Nevertheless, measure types are one of the key abstractions of `petgraph`.

PR #738, opened 2025-02-16
Connectivity Algorithms

BryanCruz:
# Add Connectivity Algorithms

Algorithms:
* Find graph **articulation points** (also know as **cut vertices** or **separating vertices**)
* Find graph **bridges** (also know as **Cut Edges** or **Cut Arcs**)
* Find graph **biconnected components** (also know as **2-connected component**)
* Find graph **2-edge-connected components**

References: 
* [Articulation Points and Biconnected Components](https://en.wikipedia.org/wiki/Biconnected_component)
* [Bridges](https://en.wikipedia.org/wiki/Bridge_(graph_theory))
* [Biconnected Graphs](https://en.wikipedia.org/wiki/Biconnected_graph)
* [k-edge-connected Graphs](https://en.wikipedia.org/wiki/K-edge-connected_graph)


RaoulLuque:
Regarding the bridge finding algorithm, there is [another PR](https://github.com/petgraph/petgraph/pull/590), which implemented such an algorithm.

I compared the performance of this implementation to the one suggested in the other PR, and the other implementation seems to be 2-7 times faster.

The benchmark setup can be found [here](https://gist.github.com/RaoulLuque/fd8afa1ad2715f7e07c0eb4964f64ed8). I just added a `benches/bridges.rs` file and reused most of the benchmarks you used, while also adding one that the other PR used.

The results from this implementation are:
```
bridges_bench           ... bench:     517,470.31 ns/iter (+/- 61,815.38)
bridges_search_2000n    ... bench:  44,879,972.10 ns/iter (+/- 3,265,361.30)
bridges_search_full     ... bench:       8,305.99 ns/iter (+/- 443.49)
bridges_search_petersen ... bench:       4,739.99 ns/iter (+/- 212.79)
bridges_search_praust   ... bench:      11,064.40 ns/iter (+/- 677.18)
```

While the results of the implementation from the other PR are:
```
bridges_bench           ... bench:      76,177.97 ns/iter (+/- 6,825.55)
bridges_search_2000n    ... bench:  19,864,410.00 ns/iter (+/- 5,565,975.53)
bridges_search_full     ... bench:       1,948.67 ns/iter (+/- 60.93)
bridges_search_petersen ... bench:         964.86 ns/iter (+/- 100.76)
bridges_search_praust   ... bench:       2,047.10 ns/iter (+/- 176.95)
```

Note that in order to be able to run the benchmarks with the other PR, I had to rebase. The setup of the same benches/bridges.rs file can be found [here](https://gist.github.com/RaoulLuque/7ce116940e9c43689e5181fd10dc059e).

PR #739, opened 2025-02-17
Add Dinic's Maximum Flow Algorithm

BryanCruz:
# Add Dinic's Maximum Flow Algorithm

References:
* https://en.wikipedia.org/wiki/Dinic%27s_algorithm
* https://cp-algorithms.com/graph/dinic.html

### DIFF
--- a/src/algo/maximum_flow/dinics.rs
+++ b/src/algo/maximum_flow/dinics.rs
@@ -1,0 +1,243 @@
+use std::{collections::VecDeque, ops::Sub};
+
+use crate::{
+    algo::{EdgeRef, PositiveMeasure},
+    prelude::Direction,
+    visit::{
+        Data, EdgeCount, EdgeIndexable, IntoEdges, IntoEdgesDirected, NodeCount, NodeIndexable,
+        VisitMap, Visitable,
+    },
+};
+
+/// Dinic's (or Dinitz's) algorithm.
+///
+/// Computes the [maximum flow][ff] of a weighted directed graph.
+///
+/// Returns the maximum flow and also the computed edge flows.
+///
+/// [ff]: https://en.wikipedia.org/wiki/Dinic%27s_algorithm
+pub fn dinics<N>(
+    network: N,
+    source: N::NodeId,
+    sink: N::NodeId,
+) -> (N::EdgeWeight, Vec<N::EdgeWeight>)
+where
+    N: NodeCount + EdgeCount + IntoEdgesDirected + EdgeIndexable + NodeIndexable + Visitable,
+    N::EdgeWeight: Sub<Output = N::EdgeWeight> + PositiveMeasure,
+{
+    let mut max_flow = N::EdgeWeight::zero();
+    let mut flows = vec![N::EdgeWeight::zero(); network.edge_count()];
+    let mut level_graph = vec![0; network.node_count()];
+    while build_level_graph(&network, source, sink, &mut level_graph, &flows) {
+        let flow_increase = find_blocking_flow(network, source, sink, &level_graph, &mut flows);
+        max_flow = max_flow + flow_increase;
+        // Resets level graph for next iteration
+        level_graph = vec![0; network.node_count()];
+    }
+    (max_flow, flows)
+}
+
+/// Makes a BFS to label network vertices with levels representing
+/// their distance to the source vertex.
+///
+/// Returns a boolean indicating if sink vertex is reachable.
+fn build_level_graph<N>(
+    network: N,
+    source: N::NodeId,
+    sink: N::NodeId,
+    level_graph: &mut [usize],
+    flows: &[N::EdgeWeight],
+) -> bool
+where
+    N: IntoEdgesDirected + NodeIndexable + EdgeIndexable + Visitable,
+    N::EdgeWeight: Sub<Output = N::EdgeWeight> + PositiveMeasure,
+{
+    let mut visited = network.visit_map();
+    let mut queue = VecDeque::new();
+    visited.visit(source);
+    queue.push_back(source);
+
+    level_graph[NodeIndexable::to_index(&network, source)] = 1;
+    while let Some(vertex) = queue.pop_front() {
+        let vertex_level = level_graph[NodeIndexable::to_index(&network, vertex)];
+        let out_edges = network.edges_directed(vertex, Direction::Outgoing);
+        let in_edges = network.edges_directed(vertex, Direction::Incoming);
+        for edge in out_edges.chain(in_edges) {
+            let next = other_endpoint(&network, edge, vertex);
+            let edge_index = EdgeIndexable::to_index(&network, edge.id());
+            let residual_cap = residual_capacity(&network, edge, next, flows[edge_index]);
+            if !visited.is_visited(&next) && (residual_cap > N::EdgeWeight::zero()) {
+                visited.visit(next);
+                level_graph[NodeIndexable::to_index(&network, next)] = vertex_level + 1;
+                queue.push_back(next);
+            }
+        }
+    }
+
+    let sink_level = level_graph[NodeIndexable::to_index(&network, sink)];
+    sink_level > 0
+}
+
+/// Find blocking flow for current level graph by repeatingly finding
+/// augmenting paths in it using DFS.
+///
+/// Attach computed flows to given `flows` parameter and returns the
+/// flow increase of current level graph.
+fn find_blocking_flow<N>(
+    network: N,
+    source: N::NodeId,
+    sink: N::NodeId,
+    level_graph: &[usize],
+    flows: &mut [N::EdgeWeight],
+) -> N::EdgeWeight
+where
+    N: NodeCount + IntoEdgesDirected + NodeIndexable + EdgeIndexable + Visitable,
+    N::EdgeWeight: Sub<Output = N::EdgeWeight> + PositiveMeasure,
+{
+    let mut flow_increase = N::EdgeWeight::zero();
+    let mut edge_to = vec![None; network.node_count()];
+    while find_augmenting_path(&network, source, sink, &level_graph, &flows, &mut edge_to) {
+        let mut path_flow = N::EdgeWeight::max();
+
+        // Find the bottleneck capacity of the path
+        let mut vertex = sink;
+        while let Some(edge) = edge_to[NodeIndexable::to_index(&network, vertex)] {
+            let edge_index = EdgeIndexable::to_index(&network, edge.id());
+            let residual_capacity = residual_capacity(&network, edge, vertex, flows[edge_index]);
+            path_flow = min::<N>(path_flow, residual_capacity);
+            vertex = other_endpoint(&network, edge, vertex);
+        }
+
+        // Update the flow of each edge along the discovered path
+        let mut vertex = sink;
+        while let Some(edge) = edge_to[NodeIndexable::to_index(&network, vertex)] {
+            let edge_index = EdgeIndexable::to_index(&network, edge.id());
+            flows[edge_index] =
+                adjusted_residual_flow(&network, edge, vertex, flows[edge_index], path_flow);
+            vertex = other_endpoint(&network, edge, vertex);
+        }
+        flow_increase = flow_increase + path_flow;
+    }
+    flow_increase
+}
+
+/// Makes a DFS to build an augmenting path to destination vertex using
+/// previously found vertex levels.
+///
+/// Returns a boolean indicating if an augmenting path to destination was found.
+fn find_augmenting_path<N>(
+    network: N,
+    source: N::NodeId,
+    destination: N::NodeId,
+    level_graph: &[usize],
+    flows: &[N::EdgeWeight],
+    edge_to: &mut [Option<N::EdgeRef>],
+) -> bool
+where
+    N: NodeCount + IntoEdgesDirected + NodeIndexable + EdgeIndexable + Visitable,
+    N::EdgeWeight: Sub<Output = N::EdgeWeight> + PositiveMeasure,
+{
+    let mut visited = network.visit_map();
+    let mut queue = VecDeque::new();
+    visited.visit(source);
+    queue.push_back(source);
+
+    while let Some(vertex) = queue.pop_front() {
+        let vertex_index = NodeIndexable::to_index(&network, vertex);
+        let out_edges = network.edges_directed(vertex, Direction::Outgoing);
+        let in_edges = network.edges_directed(vertex, Direction::Incoming);
+        for edge in out_edges.chain(in_edges) {
+            let next_vertex = other_endpoint(&network, edge, vertex);
+            let next_vertex_index = NodeIndexable::to_index(&network, next_vertex);
+            let edge_index: usize = EdgeIndexable::to_index(&network, edge.id());
+            let residual_cap = residual_capacity(&network, edge, next_vertex, flows[edge_index]);
+            if level_graph[next_vertex_index] == level_graph[vertex_index] + 1
+                && !visited.is_visited(&next_vertex)
+                && (residual_cap > N::EdgeWeight::zero())
+            {
+                visited.visit(next_vertex);
+                edge_to[next_vertex_index] = Some(edge);
+                if destination == next_vertex {
+                    return true;
+                }
+                queue.push_front(next_vertex);
+            }
+        }
+    }
+    false
+}
+
+/// Returns the adjusted residual flow for given edge and flow increase.
+fn adjusted_residual_flow<N>(
+    network: N,
+    edge: N::EdgeRef,
+    target_vertex: N::NodeId,
+    flow: N::EdgeWeight,
+    flow_increase: N::EdgeWeight,
+) -> N::EdgeWeight
+where
+    N: NodeIndexable + IntoEdges,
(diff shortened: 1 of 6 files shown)


src/algo/maximum_flow/dinics.rs
```diff
@@ -0,0 +1,244 @@
...
+
+/// Dinic's (or Dinitz's) algorithm.
+///
+/// Computes the [maximum flow][ff] of a weighted directed graph.
+///
+/// Returns the maximum flow and also the computed edge flows.
+///
+/// [ff]: https://en.wikipedia.org/wiki/Dinic%27s_algorithm
+pub fn dinics<N>(
+    network: N,
```
RaoulLuque: I am not sure about the choice of name of N as generic here, since N is already used as the name of the generic parameter for the node labels in the Graph struct. I imagine you've just copied the naming over from Ford Fulkerson, but I would prefer using G and am interested in what you think or whether I am missing something?

If you agree, we could also change the generic parameter naming in Ford Fulkerson while we are at it (or in another small PR) :)

RaoulLuque:
As mentioned in [PR 740, by @starovoid ]https://github.com/petgraph/petgraph/pull/740#issuecomment-2888556128), the algorithm does not seem to be improving upon the existing implementation when running the benchmarks. Indeed it seems to be performing worse.

This may be because the graphs from the benchmark have the same number of nodes as edges. However, in such a setting, Ford Fulkerson and Dinics should actually perform similarly. The results on my PC are however as follows:

```
dinics_bench            ... bench:      28,915.92 ns/iter (+/- 2,343.77)
ford_fulkerson_bench            ... bench:      17,511.19 ns/iter (+/- 1,822.12)
```

Which shows the decrease in performance. I also added a similar benchmark for both an increased number of edges, where the Dinic performs similarly poor. The bench setup is as follows:

```rust
#[bench]
fn dinics_bench_many_edges(bench: &mut Bencher) {
    static NODE_COUNT: usize = 1_001;
    let mut g: Graph<usize, usize> = Graph::new();
    let nodes: Vec<NodeIndex<_>> = (0..NODE_COUNT).map(|i| g.add_node(i)).collect();
    for j in [1, 2, 4, 5, 10, 20, 25, 50] {
        for i in 0..(NODE_COUNT - 1) / j  {
            g.add_edge(nodes[i], nodes[(i + 1) * j], 1);
        }
    }
    bench.iter(|| {
        let _flow = dinics(
            &g,
            NodeIndex::from(0),
            NodeIndex::from(g.node_count() as u32 - 1),
        );
    });
}
```

And the results are:
```
dinics_bench_many_edges ... bench:     355,747.67 ns/iter (+/- 14,674.08)
ford_fulkerson_bench_many_edges ... bench:     205,071.71 ns/iter (+/- 3,839.07)
```

BryanCruz:
Hi @RaoulLuque , thanks for pointing out the benchmark and for providing a larger bench test.

I reimplemented some parts of the algorithm and got it to work a lot better. 

The Dinics and Ford Fulkerson benches that were being compared contained graph instances that favored a Ford Fulkerson method, so I added some others that favors Dinics. I added the benchmark results in the PR description.  

I'll also check the Maximum Bipartite Matching PR soon.

RaoulLuque:
Hi, thanks for reviewing the code and improving it :) Those results look really promising !

Ah, I didn't know that they were more favorable for Ford Fulkerson, I'll look into that interesting. Thanks for pointing that out 😁

I'll have a look at the new code prob next week if that's okay, because I currently don't have a lot of time unfortunately :)

(If someone else wants to have a look, that is also great of course 😁)

RaoulLuque:
(one thing I can already say tho is that a docstring in the [algo docs format](https://github.com/petgraph/petgraph/blob/master/assets/guide/algodocs.md) should be added :) )

BryanCruz:
Made a fix in the algorithm execution, and added a Min S-T Cut implementation.

Will check the algo docs format.

BryanCruz:
Added docs in [algo docs format](https://github.com/petgraph/petgraph/blob/master/assets/guide/algodocs.md)

src/algo/maximum_flow/dinics.rs
```diff
@@ -0,0 +1,335 @@
...
+    destination: N::NodeId,
+) -> (N::EdgeWeight, Vec<N::EdgeWeight>)
+where
+    N: NodeCount + EdgeCount + IntoEdgesDirected + EdgeIndexable + NodeIndexable + Visitable,
+    N::EdgeWeight: Sub<Output = N::EdgeWeight> + PositiveMeasure,
+{
+    let mut max_flow = N::EdgeWeight::zero();
+    let mut flows = vec![N::EdgeWeight::zero(); network.edge_count()];
+    let mut visited = network.visit_map();
+    let mut level_edges = vec![Default::default(); network.node_count()];
```
RaoulLuque:
```suggestion
    let mut flows = vec![N::EdgeWeight::zero(); network.edge_bound()];
    let mut visited = network.visit_map();
    let mut level_edges = vec![Default::default(); network.node_bound()];
```
Please use edge and node bound instead of count, as the indices of the nodes and edges might not be 'gapless', see https://docs.rs/petgraph/latest/petgraph/#differences-of-stable-and-non-stable-graphs-in-index-management

src/algo/maximum_flow/dinics.rs
```diff
@@ -0,0 +1,335 @@
...
+    destination: N::NodeId,
+    flows: &[N::EdgeWeight],
+    level_edges: &mut [Vec<N::EdgeRef>],
+) -> Vec<usize>
+where
+    N: NodeCount + IntoEdgesDirected + NodeIndexable + EdgeIndexable,
+    N::EdgeWeight: Sub<Output = N::EdgeWeight> + PositiveMeasure,
+{
+    let mut level_graph = vec![0; network.node_count()];
+    let mut bfs_queue = VecDeque::with_capacity(network.node_count());
```
RaoulLuque:
```suggestion
    let mut level_graph = vec![0; network.node_bound()];
    let mut bfs_queue = VecDeque::with_capacity(network.node_count());
```
See previous comment. Note that `node_count()` is okay for the bfs_queue :+1:

src/algo/maximum_flow/dinics.rs
```diff
@@ -0,0 +1,335 @@
...
+where
+    N: Data,
+    N::EdgeWeight: PartialOrd,
+{
+    if a < b {
+        a
+    } else {
+        b
+    }
+}
```
RaoulLuque: This is very niche, but not sure if this is actually correct. Since you only demand the `EdgeWeight` is `PartialOrd`, a and b might not be comparable (see https://en.wikipedia.org/wiki/Partially_ordered_set) in which case b is considered the minimum? Probably this is very niche, but therefore you would probably want to impose the `Ord` trait on the EdgeWeight instead I think

src/algo/maximum_flow/dinics.rs
```diff
@@ -0,0 +1,335 @@
...
+    flows: &mut [N::EdgeWeight],
+    level_edges: &mut [Vec<N::EdgeRef>],
+    visited: &mut N::Map,
+) -> N::EdgeWeight
+where
+    N: NodeCount + IntoEdges + NodeIndexable + EdgeIndexable + Visitable,
+    N::EdgeWeight: Sub<Output = N::EdgeWeight> + PositiveMeasure,
+{
+    let mut flow_increase = N::EdgeWeight::zero();
+    let mut edge_to = vec![None; network.node_count()];
```
RaoulLuque:
```suggestion
    let mut edge_to = vec![None; network.node_bound()];
```
See previous comments

src/algo/maximum_flow/dinics.rs
```diff
@@ -0,0 +1,335 @@
...
+///
+/// # Complexity
+/// * Time complexity:
+///   * In general: **O(|V|²|E|)**
+///   * In networks with only unit capacities: **O(min{|V|²ᐟ³, |E|¹ᐟ²} |E|)**
+/// * Auxiliary space: **O(|V| + |E|)**.
+///
+/// where **|V|** is the number of nodes and **|E|** is the number of edges.
+///
+/// [ff]: https://en.wikipedia.org/wiki/Dinic%27s_algorithm
```
RaoulLuque:
```suggestion
/// Implements [Dinic's (or Dinitz's) algorithm][dinics], which builds successive
/// level graphs using breadth-first search and finds blocking flows within
/// them through depth-first searches.
///
/// # Arguments
/// * `network` — A directed graph with positive edge weights, namely "flow capacities".
/// * `source` — The source node where flow originates.
/// * `destination` — The destination node where flow terminates.
///
/// # Returns
/// Returns a tuple of two values:
/// * `N::EdgeWeight`: computed maximum flow;
/// * `Vec<N::EdgeWeight>`: the flow of each edge. The vector is indexed by the graph's edge indices.
///
/// # Complexity
/// * Time complexity:
///   * In general: **O(|V|²|E|)**
///   * In networks with only unit capacities: **O(min{|V|²ᐟ³, |E|¹ᐟ²} |E|)**
/// * Auxiliary space: **O(|V| + |E|)**.
///
/// where **|V|** is the number of nodes and **|E|** is the number of edges.
///
/// [dinics]: https://en.wikipedia.org/wiki/Dinic%27s_algorithm
```

RaoulLuque (commented):
Hey, thanks for reworking some things from the initial implementation and adapting the docs to match our algo docs :sauropod: 

I would suggest to move the minimum S-T Cut to another PR :+1: 

Other than that, I suggested a few changes, but in general, LGTM :)

I was wondering with the benchmarks, whether you looked at some of the graphs from the common.rs and took into consideration using one of those?

src/algo/maximum_flow/dinics.rs
BryanCruz: > I was wondering with the benchmarks, whether you looked at some of the graphs from the common.rs and took into consideration using one of those?

I took a look at the `factories.rs`, but didn't found any weighted graph.

The tests for Maximum Flow should be a bit more specific than generic graphs because of the network nature the problem instance should have. Also, we need Source and Destination nodes

So its a lot easier to create specific instances for benchmarks than adapt some existing one 🤔 

> I am not sure about the choice of name of N as generic here, since N is already used as the name of the generic parameter for the node labels in the Graph struct. I imagine you've just copied the naming over from Ford Fulkerson, but I would prefer using G and am interested in what you think or whether I am missing something?

I don't have a strong preference here. I think `N` also makes sense because we're dealing with **N**etworks, but I can change to `G` in order to not conflict with node labels default generic name.

> If you agree, we could also change the generic parameter naming in Ford Fulkerson while we are at it (or in another small PR) :)

I can change here, no problem

BryanCruz: > I would suggest to move the minimum S-T Cut to another PR 👍

#847

src/algo/maximum_flow/dinics.rs
BryanCruz: done

also added a test to check the correct behavior in stable graphs

src/algo/maximum_flow/dinics.rs
BryanCruz: instead of adding the `Ord` trait requirement, I added `panic` when neither `a < b` or `a >= b` are `true`

I think its worth to keep the `PartialOrd` to handle cases like floats as edge weights

RaoulLuque: Ah, that is a good point. Allowing floats is definitely desirable. I am not sure though whether panicking is desirable.

Maybe someone else has a good idea for handling this, but maybe we should do if b < a { b } else { a } and then always call it with the existing value as a and the new value as b. This way, we won't update, if the values are not updatable?

I think I'll be fine both ways, just want to make sure we thought about all the possibilities :)

src/algo/maximum_flow/dinics.rs
RaoulLuque: That's great, thanks :sun_with_face: 

This is just a general not-quite-satisfiedness of myself with the current testing, but probably in the future there should be some framework for us to test that algorithms that are supposed to work in different classes, actually do :) But that's just a side note and does not have to bother you ^^

src/algo/maximum_flow/dinics.rs
RaoulLuque: Thanks for separating into two PRs :+1: I think I'd prefer using `G` as the generic parameter, just for consistency sake :)

RaoulLuque:
Sry for the additional comment, but since you seem to be quite knowledgable when it comes to the different use-cases of Dinics/Ford Fulkerson (Edmonds Karp), I think it would be a good idea to document when to use which function. We could use the `maximum_flow/mod.rs` for this and add a comment there. In the `ford_fulkerson` and `dinics` docstrings we could then refer to the `mod.rs` docs for when to use this or other maximum flow functions :)

(You can use `cargo doc --open` to compile and open the docs in your browser to see how they would look like and if links etc. work ^^)

src/algo/maximum_flow/dinics.rs
BryanCruz: I'm in favor of `panic!` because if the algorithm doesn't work for a certain graph instance, then the reason is very clear for the user.

Without raising any errors and failing silently, if values are not comparable for some reason, the algorithm could give an invalid answer to the user and they may not know why it's wrong or not even notice.

If someone realy needs the algorithm to work in graph instances with values that are not comparable, they can make a few tweaks and create another function like `dinics_unsafe`. Since I don't know if such case will ever exist, I wouldn't implement it now.

What do you think ?

BryanCruz:
> I think it would be a good idea to document when to use which function. We could use the `maximum_flow/mod.rs` for this and add a comment there. In the `ford_fulkerson` and `dinics` docstrings we could then refer to the `mod.rs` docs for when to use this or other maximum flow functions :)

good idea, added in latest commit

src/algo/maximum_flow/dinics.rs
RaoulLuque: I see, yeah. You're arguments do make sense :+1: As I said I am torn, or rather fine with both solutions :) In any case, we should document the behavior clearly in the docstring :+1: ( If u didn't do that already 😁)

RaoulLuque:
Okay, so if you adjust the docstring of dinics to document the possibility to panic as stated in https://github.com/petgraph/petgraph/pull/739#discussion_r2217908348, I think this Pr would be ready to merge, thanks a lot again :star2: 

Just one more comment I have, just below. Note that you can also accept suggestions right here on GitHub, this way the reviewers will be credited in Git for their suggestions :)

src/algo/maximum_flow/ford_fulkerson.rs
```diff
@@ -0,0 +1,215 @@
...
+        flow + delta
+    } else {
+        let end_point = NodeIndexable::to_index(&network, vertex);
+        panic!("Illegal endpoint {}", end_point);
+    }
+}
+
+/// [Ford-Fulkerson][ff] algorithm in the [Edmonds-Karp][ek] variation.
+///
+/// Computes the [maximum flow] from `source` to `destination` in a weighted directed graph.
```
RaoulLuque:
```suggestion
/// [Ford-Fulkerson][ff] algorithm in the [Edmonds-Karp][ek] variation.
/// Computes the [maximum flow] from `source` to `destination` in a weighted directed graph.
```
This makes it such that on the module docs for `maximum_flow`, the description of Ford Fulkerson also shows the text about computing a max flow :)

src/algo/maximum_flow/dinics.rs
BryanCruz: adjusted docstrings

src/algo/maximum_flow/dinics.rs
```diff
@@ -11,11 +11,15 @@ use crate::{
...
 /// Compute the maximum flow from `source` to `destination` in a directed graph.
-///
 /// Implements [Dinic's (or Dinitz's) algorithm][dinics], which builds successive
 /// level graphs using breadth-first search and finds blocking flows within
 /// them through depth-first searches.
 ///
+/// For simplicity, the algorithm requires `N::EdgeWeight` to implement
+/// only [PartialOrd] trait, and not [Ord], but will panic if it tries to
+/// compare two elements that aren't comparable (i.e., given two edges `a`
+/// and `b`, neither `a >= b` or `a < b`).
```
RaoulLuque:
```suggestion
/// compare two elements that aren't comparable (i.e., given two edge weights `a`
/// and `b`, where neither `a >= b` nor `a < b`).
```

src/algo/maximum_flow/dinics.rs
```diff
@@ -310,6 +314,8 @@ where
 }
 
 /// Returns the minimum value between given `a` and `b`.
+/// Will panic if it tries to compare two elements that aren't comparable
+/// (i.e., given two elements `a` and `b`, neither `a >= b` or `a < b`).
```
RaoulLuque:
```suggestion
/// (i.e., given two elements `a` and `b`, neither `a >= b` nor `a < b`).
```

RaoulLuque:
Looks great, thanks for all your patience with my review :relaxed:  Just these two small language nitpicks, after which I will merge the PR :+1:

BryanCruz:
> Looks great, thanks for all your patience with my review ☺️ Just these two small language nitpicks, after which I will merge the PR 👍

thanks ☺️ , commited the latest suggestions

PR #740, opened 2025-02-17
Add Maximum Bipartite Matching Algorithm

BryanCruz:
# Add Maximum Bipartite Matching Algorithm

Solves the Maximum Matching Problem for a Bipartite graph by reducing it to a Maximum Flow problem and using existent algorithm for Maximum Flow.

References:
* https://en.wikipedia.org/wiki/Matching_(graph_theory)
* https://en.wikipedia.org/wiki/Maximum_flow_problem

## Creating the Max Flow Instance
In the algorithm I manually copy all nodes and edges from original graph to create a network flow instance. 
I believe there are faster ways to do this and optimize the algorithm, but I don't know how to do so, so I'm accepting some help on this.

## Further Optimization
This algorithm can be further optimized when [Dinic's Algorithm PR](https://github.com/petgraph/petgraph/pull/739) is accepted, because Dinic's is much faster than Ford Fulkerson method for [finding maximum flows in unit networks](https://cp-algorithms.com/graph/dinic.html#unit-networks).

I executed the bench tests with Ford Fulkerson and with the Dinic's algorithm, and this were the results:
```
using ford_fulkerson to find max flow:
    test maximum_bipartite_matching_100  ... bench:      32,223.41 ns/iter (+/- 5,470.65)
    test maximum_bipartite_matching_1000 ... bench:  27,260,136.10 ns/iter (+/- 5,475,181.00)

using dinics to find max flow:
    test maximum_bipartite_matching_100  ... bench:      14,483.79 ns/iter (+/- 512.79)
    test maximum_bipartite_matching_1000 ... bench:   1,867,409.90 ns/iter (+/- 105,994.25)
```

### DIFF
--- a/benches/matching.rs
+++ b/benches/matching.rs
@@ -3,13 +3,14 @@
 extern crate petgraph;
 extern crate test;
 
+use petgraph::visit::NodeIndexable;
 use test::Bencher;
 
 #[allow(dead_code)]
 mod common;
 use common::*;
 
-use petgraph::algo::{greedy_matching, maximum_matching};
+use petgraph::algo::{greedy_matching, maximum_bipartite_matching, maximum_matching};
 use petgraph::graph::UnGraph;
 
 fn huge() -> UnGraph<(), ()> {
@@ -27,6 +28,30 @@ fn huge() -> UnGraph<(), ()> {
 
     // 999 nodes, 83500 edges
     UnGraph::from_edges(&edges)
+}
+
+fn generate_bipartite(node_count: u32) -> (UnGraph<(), ()>, Vec<u32>, Vec<u32>) {
+    let mut edges = Vec::new();
+
+    let mut partition_1 = Vec::new();
+    let mut partition_2 = Vec::new();
+    for i in 0..node_count {
+        for j in i..node_count {
+            if i % 6 == 0 && j % 2 == 1 {
+                edges.push((i, j));
+            }
+        }
+    }
+
+    for i in (0..node_count).step_by(6) {
+        partition_1.push(i);
+    }
+
+    for i in (1..node_count).step_by(2) {
+        partition_2.push(i);
+    }
+
+    (UnGraph::from_edges(&edges), partition_1, partition_2)
 }
 
 #[bench]
@@ -75,4 +100,32 @@ fn maximum_matching_bigger(bench: &mut Bencher) {
 fn maximum_matching_huge(bench: &mut Bencher) {
     let g = huge();
     bench.iter(|| maximum_matching(&g));
+}
+
+#[bench]
+fn maximum_bipartite_matching_100(bench: &mut Bencher) {
+    let (g, partition_1, partition_2) = generate_bipartite(100);
+    let partition_1_ids = partition_1
+        .iter()
+        .map(|&id| NodeIndexable::from_index(&g, id as usize))
+        .collect();
+    let partition_2_ids = partition_2
+        .iter()
+        .map(|&id| NodeIndexable::from_index(&g, id as usize))
+        .collect();
+    bench.iter(|| maximum_bipartite_matching(&g, &partition_1_ids, &partition_2_ids));
+}
+
+#[bench]
+fn maximum_bipartite_matching_1000(bench: &mut Bencher) {
+    let (g, partition_1, partition_2) = generate_bipartite(1_000);
+    let partition_1_ids = partition_1
+        .iter()
+        .map(|&id| NodeIndexable::from_index(&g, id as usize))
+        .collect();
+    let partition_2_ids = partition_2
+        .iter()
+        .map(|&id| NodeIndexable::from_index(&g, id as usize))
+        .collect();
+    bench.iter(|| maximum_bipartite_matching(&g, &partition_1_ids, &partition_2_ids));
 }
--- a/src/algo/matching.rs
+++ b/src/algo/matching.rs
@@ -2,9 +2,11 @@ use std::collections::VecDeque;
 use std::hash::Hash;
 
 use crate::visit::{
-    EdgeRef, GraphBase, IntoEdges, IntoNeighbors, IntoNodeIdentifiers, NodeCount, NodeIndexable,
-    VisitMap, Visitable,
+    EdgeCount, EdgeIndexable, EdgeRef, GraphBase, IntoEdges, IntoNeighbors, IntoNodeIdentifiers,
+    IntoNodeReferences, NodeCount, NodeIndexable, VisitMap, Visitable,
 };
+
+use crate::{algo::ford_fulkerson, graph::NodeIndex, Directed, Graph};
 
 /// Computed
 /// [*matching*](https://en.wikipedia.org/wiki/Matching_(graph_theory)#Definitions)
@@ -590,5 +592,109 @@ fn augment_path<G>(
         augment_path(graph, target, source, mate, label);
     } else {
         panic!("Unexpected label when augmenting path");
+    }
+}
+
+/// Compute the [*maximum matching*][1]
+/// for a bipartite graph by reducing it to a [maximum flow][2] problem,
+/// and then using Ford Fulkerson method to solve maximum flow problem.
+///
+/// [1]: https://en.wikipedia.org/wiki/Matching_(graph_theory)
+/// [2]: https://en.wikipedia.org/wiki/Maximum_flow_problem
+///
+/// The input graph is treated as if undirected.
+pub fn maximum_bipartite_matching<G>(
+    graph: G,
+    partition_a: &Vec<G::NodeId>,
+    partition_b: &Vec<G::NodeId>,
+) -> Matching<G>
+where
+    G: NodeIndexable + EdgeIndexable + NodeCount + EdgeCount + IntoNodeReferences + IntoEdges,
+{
+    let (network, source, sink) =
+        maximum_bipartite_matching_instance(&graph, partition_a, partition_b);
+
+    let (_, flow) = ford_fulkerson(&network, source, sink);
+    let mut mate = vec![None; graph.node_count()];
+    let mut n_edges = 0;
+
+    for edge in graph.edge_references() {
+        if flow[EdgeIndexable::to_index(&graph, edge.id())] == 1 {
+            let (source, target) =
+                source_and_target_from_partitions::<G>(edge, partition_a, partition_b);
+            mate[NodeIndexable::to_index(&graph, source)] = Some(target);
+            mate[NodeIndexable::to_index(&graph, target)] = Some(source);
+            n_edges += 1;
+        }
+    }
+
+    Matching::new(graph, mate, n_edges)
+}
+
+/// Create a network from given graph.
+/// Created Nodes and Edges indices are compatible
+/// with the ones from original graph.
+fn maximum_bipartite_matching_instance<G>(
+    graph: &G,
+    partition_a: &Vec<G::NodeId>,
+    partition_b: &Vec<G::NodeId>,
+) -> (Graph<(), usize, Directed>, NodeIndex, NodeIndex)
+where
+    G: NodeIndexable + EdgeIndexable + NodeCount + EdgeCount + IntoNodeReferences + IntoEdges,
+{
+    let mut network = Graph::with_capacity(
+        graph.node_count() + 2,
+        graph.edge_count() + partition_a.len() + partition_b.len(),
+    );
+
+    // Add nodes from original graph
+    for _ in 0..graph.node_count() {
+        network.add_node(());
+    }
+
+    // Add edges from original graph, directed from partition_a to partition_b
+    for edge in graph.edge_references() {
+        let (source, target) =
+            source_and_target_from_partitions::<G>(edge, partition_a, partition_b);
+        let source_index = NodeIndexable::to_index(&graph, source);
+        let target_index = NodeIndexable::to_index(&graph, target);
+        network.add_edge(
+            NodeIndexable::from_index(&network, source_index),
+            NodeIndexable::from_index(&network, target_index),
+            1,
+        );
+    }
+
+    // Add source node
+    let source = network.add_node(());
+    for &node in partition_a {
+        let node_index = NodeIndexable::to_index(&graph, node);
+        network.add_edge(source, NodeIndex::new(node_index), 1);
+    }
+
+    // Add sink node
+    let sink = network.add_node(());
+    for &node in partition_b {
+        let node_index = NodeIndexable::to_index(&graph, node);
+        network.add_edge(NodeIndex::new(node_index), sink, 1);
+    }
+
+    (network, source, sink)
+}
+
+fn source_and_target_from_partitions<G>(
+    edge: G::EdgeRef,
+    partition_a: &Vec<G::NodeId>,
+    partition_b: &Vec<G::NodeId>,
+) -> (G::NodeId, G::NodeId)
+where
+    G: IntoEdges,
+{
+    if partition_a.contains(&edge.source()) {
+        (edge.source(), edge.target())
+    } else if partition_b.contains(&edge.source()) {
+        (edge.target(), edge.source())
+    } else {
+        panic!("Partitions are inconsistent.");
     }
 }
--- a/tests/matching.rs
+++ b/tests/matching.rs
@@ -1,8 +1,9 @@
 use std::collections::HashSet;
 use std::hash::Hash;
 
-use petgraph::algo::{greedy_matching, maximum_matching};
+use petgraph::algo::{greedy_matching, maximum_bipartite_matching, maximum_matching};
 use petgraph::prelude::*;
+use petgraph::visit::EdgeIndexable;
 
 macro_rules! assert_one_of {
     ($actual:expr, [$($expected:expr),+]) => {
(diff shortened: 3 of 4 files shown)


tests/matching.rs
```diff
@@ -140,3 +140,68 @@ fn is_perfect_in_stable_graph() {
...
+#[test]
+fn maximum_bipartite_empty() {
+    let g: UnGraph<(), ()> = UnGraph::default();
+    let m = maximum_bipartite_matching(&g, &[], &[]);
+    assert_eq!(collect(m.edges()), set![]);
+    assert_eq!(collect(m.nodes()), set![]);
+}
+
+#[test]
+fn maximum_bipartite_k2() {
```
RaoulLuque, line 153: Maybe additionally/instead (of this one) add disjoint/odd_path/star test cases like with the other matching tests?

benches/matching.rs
```diff
@@ -29,6 +30,30 @@ fn huge() -> UnGraph<(), ()> {
     UnGraph::from_edges(&edges)
 }
 
+fn generate_bipartite(node_count: u32) -> (UnGraph<(), ()>, Vec<u32>, Vec<u32>) {
```
RaoulLuque, line 33: Maybe add as a docstring the number of nodes (possibly with partition size) and vertices in the resulting graph and rename the function parameter to something else than node_count since it does not actually reflect the final node count of the bipartite graph

RaoulLuque (commented):
Hey, first off, thanks for the PR :) I left some comments but other than that, it looks good to me 👍 Maybe check if a rebase is necessary?

I'm not a maintainer, so I don't have the final say for a merge, but will try to make someone else have a look at it too !

src/algo/matching.rs
```diff
@@ -592,3 +594,107 @@ fn augment_path<G>(
...
 }
+
+/// Compute the [*maximum matching*][1]
+/// for a bipartite graph by reducing it to a [maximum flow][2] problem,
+/// and then using Ford Fulkerson method to solve maximum flow problem.
+///
+/// [1]: https://en.wikipedia.org/wiki/Matching_(graph_theory)
+/// [2]: https://en.wikipedia.org/wiki/Maximum_flow_problem
+///
+/// The input graph is treated as if undirected.
```
RaoulLuque, line 605: I would suggest adding a docstring example for users to better understand usage of the function. See e.g. Ford fulkerson

RaoulLuque:
> It's interesting to see benchmarks against the existing [`maximum_matching`](https://docs.rs/petgraph/latest/petgraph/algo/matching/fn.maximum_matching.html) algorithm.

Here are the benchmarks from my machine. The generic implementation is the existing one and the specific one is the one from this PR. The numbers indicate a variable which indicates 3/2 times the node count (1000 -> 666 nodes in the graph):
```
maximum_bipartite_matching_1000_generic_implementation  ... bench:  62,301,335.60 ns/iter (+/- 2,157,471.47)
maximum_bipartite_matching_1000_specific_implementation ... bench:  34,427,523.40 ns/iter (+/- 1,588,382.20)
maximum_bipartite_matching_100_generic_implementation   ... bench:      43,549.43 ns/iter (+/- 2,659.71)
maximum_bipartite_matching_100_specific_implementation  ... bench:      44,129.50 ns/iter (+/- 3,134.17)
```
As can be seen, the difference seems to be noticeable on bigger graphs but negligible on smaller graphs.

However, the difference increases when instead of Ford-Fulkerson, Dinics from [this PR](https://github.com/petgraph/petgraph/pull/739) is used (I was unable to find out, why the other results change so drastically as well, since the only thing I change is the Max-Flow algorithm used in the maximum bipartite matching algorithm - I reran both benchmarks multiple times with the same results):
```
test maximum_bipartite_matching_1000_generic_implementation  ... bench:  74,867,185.50 ns/iter (+/- 10,313,362.71)
test maximum_bipartite_matching_1000_specific_implementation ... bench:   2,635,163.48 ns/iter (+/- 144,904.74)
test maximum_bipartite_matching_100_generic_implementation   ... bench:      19,793.50 ns/iter (+/- 6,764.38)
test maximum_bipartite_matching_100_specific_implementation  ... bench:      18,730.08 ns/iter (+/- 3,924.35)
```

starovoid:
> However, the difference increases when instead of Ford-Fulkerson, Dinics from [this PR](https://github.com/petgraph/petgraph/pull/739) is used (I was unable to find out, why the other results change so drastically as well, since the only thing I change is the Max-Flow algorithm used in the maximum bipartite matching algorithm

The results are similar to the truth, but also impressive! Dinic's algorithm is theoretically indeed faster that current max-flow implementation without shortest path searching, and this benchmark confirms this. But to be sure, I would like to take a look at the benchmark code.

Thank you for the good work done, I would like to include the benchmark code in the repository after the PR merge.

RaoulLuque:
Regarding:

> In the algorithm I manually copy all nodes and edges from original graph to create a network flow instance.
I believe there are faster ways to do this and optimize the algorithm, but I don't know how to do so, so I'm accepting some help on this.

I tried out some things, but was also unable to find a good approach. I thought about copying the entire graph and deleting all edges, but that added new problems, where the newly created graph was of type G which one would have to impose more trait bounds on. So maybe, @starovoid, you have an idea, and otherwise we just leave it at that.

RaoulLuque:
On the same note I wondered whether the docstring you write for one of your helper functions is actually correct:
```rust
/// Create a network from given graph.
/// Created Nodes and Edges indices are compatible
/// with the ones from original graph.
fn maximum_bipartite_matching_instance<G>( ...
```

Since you just iterate over the node count in the given graph and create a node for each, the guarantee would only hold, if the graph has "packed" node indices. In particular, this does not hold for StableGraphs. However, I was not able to feed a StableGraph into the function yet, since I was somehow not able to create a StableGraph type which satisfies the IntoEdges trait bound.

RaoulLuque:
> The results are similar to the truth, but also impressive! Dinic's algorithm is theoretically indeed faster that current max-flow implementation without shortest path searching, and this benchmark confirms this. But to be sure, I would like to take a look at the benchmark code.
> 
> Thank you for the good work done, I would like to include the benchmark code in the repository after the PR merge.

I see. What do you mean exactly by take a look at the benchmark code? The code of the benchmark is included in this PR and the [linked Dinics PR](https://github.com/petgraph/petgraph/pull/739) in [benches/matching.rs](https://github.com/petgraph/petgraph/blob/374679a077501aea5c4d2833ad9b30ccf0426ad0/benches/matching.rs).

I just copied the benchmark and used the generic maximum matching algorithm instead of the bipartite specific one:
```rust
#[bench]
fn maximum_bipartite_matching_1000_generic_implementation(bench: &mut Bencher) {
    let (g, partition_a, partition_b) = generate_bipartite(1_000);
    let partition_a_ids: Vec<_> = partition_a
        .iter()
        .map(|&id| NodeIndexable::from_index(&g, id as usize))
        .collect();
    let partition_b_ids: Vec<_> = partition_b
        .iter()
        .map(|&id| NodeIndexable::from_index(&g, id as usize))
        .collect();
    bench.iter(|| maximum_matching(&g));
}
```

Note that I also did not remove the partition a and b collection code since I was not sure if it counts towards the benchmark time. Probably not because it is not part of the bench closure, but I wanted to make sure

src/algo/matching.rs
```diff
@@ -592,3 +594,107 @@ fn augment_path<G>(
...
+            n_edges += 1;
+        }
+    }
+
+    Matching::new(graph, mate, n_edges)
+}
+
+/// Create a network from given graph.
+/// Created Nodes and Edges indices are compatible
+/// with the ones from original graph.
```
RaoulLuque, line 636: On the same note of creating the max flow instance: I wondered whether the docstring you write here is actually correct. That is, whether the indices are actually compatible with the ones from the "original" graph.

Since you just iterate over the node count in the given graph and create a node for each, the guarantee would only hold, if the graph has "packed" node indices. In particular, this does not hold for StableGraphs. However, I was not able to feed a StableGraph into the function yet, since I was somehow not able to create a StableGraph type which satisfies the IntoEdges trait bound.

starovoid:
> So maybe, @starovoid, you have an idea, and otherwise we just leave it at that.

The test from #739 on my PC showed the opposite result in the max-flow task, so the first step is to dive into `dinic`, write more benchmarks and understand this phenomenon. 

First attemt
```
test dinics_bench ... bench:                    16,392.67 ns/iter (+/- 230.41)
test ford_fulkerson_bench ... bench:       9,859.98 ns/iter (+/- 152.55)
```

Second attempt
```
test dinics_bench ... bench:      15,856.29 ns/iter (+/- 157.41)
test ford_fulkerson_bench ... bench:       9,486.44 ns/iter (+/- 211.69)
```
---


If the `dinic` implementation turns out to be really faster, the first task will be to adopt it. I will pre-schedule both of these algorithms for version `0.8.4`. 

If it turns out to solve all the issues quickly, then I will include it in version `0.8.3`, but at the moment there are many other PRs that are hungry for attention. For example, we need to choose between #590 and `bridges` from #738.

RaoulLuque:
> The test from #739 on my PC showed the opposite result in the max-flow task, so the first step is to dive into `dinic`, write more benchmarks and understand this phenomenon.

Okay, interesting. I did not actually run those benchmarks yet, good catch. I will then briefly look into those as well.

src/algo/matching.rs
RaoulLuque: I just confirmed that this may lead to issues with StableGraph. I tested this by adding the following to algo/matching.rs:

```rust
#[cfg(test)]
mod tests {
    use std::*;
    use std::prelude::rust_2024::Vec;
    use crate::prelude::*;
    use crate::algo::matching::maximum_bipartite_matching_instance;
    use crate::prelude::{StableGraph, StableUnGraph};

    #[test]
    fn foo() {

        println!("Test");

        let mut g: StableGraph<(), (), _> = StableUnGraph::default();

        let a_1 = g.add_node(());
        let a_2 = g.add_node(());
        let b_1 = g.add_node(());
        let b_2 = g.add_node(());

        // g.extend_with_edges([
        //     (a_2, b_1),
        //     (a_2, b_1),
        // ]);
        println!("Before removal: {:?}", g.node_indices().collect::<Vec<_>>());
        // [NodeIndex(0), NodeIndex(1), NodeIndex(2), NodeIndex(3)]  ← no gap

        // 2. Remove node a_1
        g.remove_node(a_1);
        println!("After removal:  {:?}", g.node_indices().collect::<Vec<_>>());
        // [NodeIndex(1), NodeIndex(2), NodeIndex(3)]  ← gap at index 0


        let g_i = maximum_bipartite_matching_instance(
            &&g,
            &[a_2],
            &[b_1, b_2],
        );

        println!("Created Bipartite Graph:  {:?}", g_i.0.node_indices().collect::<Vec<_>>());
        // [NodeIndex(0), NodeIndex(1), NodeIndex(2), NodeIndex(3), NodeIndex(4)]  ← NodeIndex 0 is back (this is the old NodeIndex(1) vertex
        println!("Created Bipartite Graph Edges: {:?}", g_i.0.edge_references().map(|e| (e.source(), e.target())).collect::<Vec<_>>());
        // From the edges we can see that the newly created source and sink correspond the NodeIndices 3 and 4 respectively.
    }
}
```

And running this test with:
```
cargo test --lib algo::matching::tests::foo -- --nocapture
```

One can now see that the indexes of the nodes are shifted. In the original (Stable) graph, the indices of the actual nodes are 1,2,3 and in the constructed bipartite graph, their corresponding node indices are 0,1,2.

BryanCruz:
> > So maybe, @starovoid, you have an idea, and otherwise we just leave it at that.
> 
> The test from #739 on my PC showed the opposite result in the max-flow task, so the first step is to dive into `dinic`, write more benchmarks and understand this phenomenon.
> 
> First attemt
> 
> ```
> test dinics_bench ... bench:                    16,392.67 ns/iter (+/- 230.41)
> test ford_fulkerson_bench ... bench:       9,859.98 ns/iter (+/- 152.55)
> ```
> 
> Second attempt
> 
> ```
> test dinics_bench ... bench:      15,856.29 ns/iter (+/- 157.41)
> test ford_fulkerson_bench ... bench:       9,486.44 ns/iter (+/- 211.69)
> ```
> 
> If the `dinic` implementation turns out to be really faster, the first task will be to adopt it. I will pre-schedule both of these algorithms for version `0.8.4`.
> 
> If it turns out to solve all the issues quickly, then I will include it in version `0.8.3`, but at the moment there are many other PRs that are hungry for attention. For example, we need to choose between #590 and `bridges` from #738.

@starovoid about the Dinic's implementation and why the benchmarks in it were slower than Ford-Fulkersons, I added some commits and comments in #739 . Also, updated its description.

BryanCruz:
> However, the difference increases when instead of Ford-Fulkerson, Dinics from https://github.com/petgraph/petgraph/pull/739 is used (I was unable to find out, why the other results change so drastically as well, since the only thing I change is the Max-Flow algorithm used in the maximum bipartite matching algorithm - I reran both benchmarks multiple times with the same results):

@RaoulLuque see Analysis > Special Cases from [Dinics' Wikipedia article](https://en.wikipedia.org/wiki/Dinic%27s_algorithm):

image

src/algo/matching.rs
BryanCruz: thanks for noticing this problem

I added a mapping from the original graph indices to the "cloned" graph indices, should resolve this problem

tests/matching.rs
BryanCruz: added tests for disjoint and star

odd_path doesn't fit to a bipartite instance

benches/matching.rs
BryanCruz: added docs and renamed node_count

BryanCruz:
added benchmarks for current `maximum_matching` with the same bipartite instances used in new algorithm

src/algo/matching.rs
BryanCruz: added docs in petgraph's recommended format

src/algo/matching.rs
BryanCruz: added a test with stable graph to assert it's working even with removed nodes

RaoulLuque:
Thanks for the great work :star2: I'll have a look latest in the course of next week (does not mean someone else shouldn't / can't have a look in the meantime ^^) :)

In the meantime you could check if the comments on the other PR about `node_bound` instead of `node_count` apply here somewhere as well :+1:

RaoulLuque:
Hey, just wanted to let you know that I did not forget about reviewing, but currently I have exam phase at Uni and it was a bit more busy than I expected lately so I was not able to review your PR until now :o The phase will end on 24th however, and after that I should have enough time to review the PR so we can get the improvements and your work merged :)

benches/matching.rs
```diff
@@ -29,6 +30,47 @@ fn huge() -> UnGraph<(), ()> {
     UnGraph::from_edges(&edges)
 }
 
+/// Generate a bipartite graph instance.
+///
+/// This function produces an undirected bipartite graph with two partitions:
+/// * `partition_1`: nodes with indices divisible by 6 (i.e., `0, 6, 12, ...`);
+/// * `partition_2`: nodes with odd indices (i.e., `1, 3, 5, ...`).
```
RaoulLuque, line 37: I think the docstring for the helper function is a great idea ! :heart_hands: However, technically I think the resulting graph has `node_count_upper_bound` many nodes. More precisely, the even indices not divisible by 6 will just be nodes without any edges adjacent to them. Not 100% sure, but this should be what happens when one calls `from_edges`. So maybe specify that as well. That is, that the resulting graph will have some node without any edges :+1: 
I would however suggest an alternative approach to generate the graph (with only the vertices with indices divisible by 6 or odd), by only actually adding the vertices that are needed (with indices divisible by 6 or odd) and returning a HashMap with your graph which maps the indices to the NodeIndexes in the graph.
(I am suggesting the latter since we are using these graphs to benchmark and thus they probably shouldn't contain unnecessary nodes since these might bias towards one algorithm or another if one algorithm does not handle "stale" nodes well).

benches/matching.rs
```diff
@@ -29,6 +30,47 @@ fn huge() -> UnGraph<(), ()> {
...
+///
+/// An edge is created between a node `a` in `partition_1` and a node `b` in `partition_b`
+/// if `a` index is lower than `j` index.
```
RaoulLuque, line 40:
```suggestion
/// An edge is created between a node with index `i` in `partition_1` and a node with index `j` in `partition_2`
/// if `i` is smaller than or equal to `j`.
```

src/algo/matching.rs
```diff
@@ -618,3 +621,164 @@ fn augment_path<G>(
...
+    let destination = network.add_node(());
+    for &node in partition_b {
+        let node_index = NodeIndexable::to_index(&graph, node);
+        network.add_edge(*index_map.get(&node_index).unwrap(), destination, 1);
+    }
+
+    (network, source, destination)
+}
+
+fn source_and_target_from_partitions<G>(
```
RaoulLuque, line 769:
```suggestion

/// Returns (endpoint_a, endpoint_b) such that endpoint_a is in `partition_a` and endpoint_b in `partition_b`.
///
/// Panics if the source node is in neither partition.
fn source_and_target_from_partitions<G>(
```

tests/matching.rs
RaoulLuque: I see, nice, thanks :) Why does odd_path not fit? The graph should be bipartite I think 😁

RaoulLuque (commented):
Alright, got to the review a bit earlier than expected I guess :) Again, thanks a lot for your great work 🌞  Had only a few nitpicks 👍

tests/matching.rs
BryanCruz: You're right, my bad. Added test case for odd_path too

Issue #737, opened 2025-02-13
Construct `Measure`s from `f32` & `f64`


starovoid:
### Abstract
Add the `from_f32` and `from_f64` constructors to the `FloatMeasure`, `UnitMeasure` & `BoundedMeasure` traits:
```rust
fn from_f32(val: f32) -> Self;
fn from_f64(val: f64) -> Self;
```

### Motivation
This will make it possible to construct assoiceated *weight* types inside generic algorithms, for example, by allowing to write something like:
```rust
if distance[ix(i)] + w + G::EdgeWeight::from_f32(0.1) < distance[ix(j)]
```
Actually, such feature request was received in issue #723.

### Details
- To get integer types from `floats`, I suggest using the standard `as` cast in implementations.
- I suggest implementing exactly both methods for `f32` and `f64`, so as not to lose flexibility in choosing between accuracy and memory consumption.
- The `Measure` trait looks too general and in my opinion does not have to be a **Real number** semantically, so we should not implement `from_f64` and `from_f32` for it.
- These changes are definitely breaking and can only be accepted in a major release (`0.8.0` or further).

### Progress
#733
- [ ] `FlaotMeasure`
- [ ] `UnitMeasure`
- [x] `BoundedMeasure`

PR #728, opened 2025-02-04
feat!: Add support for specifying rankdir on dot plots.

volodymyrd:
Extends Config for rankdir support - direction of [graph layout](https://graphviz.org/docs/attrs/rankdir/).

BREAKING CHANGE:

Add new variant of public enum `dot::Config`.

PR #662, opened 2024-08-23
feat: Add Bron-Kerbosch algorithm for maximal cliques

qoqosz:
This PR resurrects https://github.com/petgraph/petgraph/pull/72 by applying @jrraymond's changes to the recent (`0.6.5`) `master` branch with only minor code style changes. This also addresses a recent issue https://github.com/petgraph/petgraph/issues/620.

The added algorithm enumerates maximal cliques in a graph by using Bron-Kerbosch algorithm with pivoting.

Specifically, my added changes include:
- extracting the algorithm to its separate file
- adding a documentation with an example
- moving the reference implementation to the tests file
- modifying test cases so they do not take too long to run

I did not apply the suggestion to use `FixedBitSet` instead of `HashMap` from the original PR as I don't know how to approach it for the generic graph.

PR #753, opened 2025-03-21
feat!: Release `0.8.0`

starovoid:
A sufficient number of proposed changes have accumulated to combine them and publish a new major release numbered `0.8.0`.

BREAKING CHANGE:

This will require the user to provide extra type parameter in some APIs (Read more in #747).

## List of changes
- [x] #747
    The main innovation of the current release, the long-awaited feature that has become very relevant due to the transition of dependent projects to support `no_std`.
- [x] #662 
- [x] #611
- [x] #728
- [x] #686
- [x] #737 
- [x] #720 
- [x] #718 

## Note
There are still a large number of PRs that we want to adopt in the near future, so we should expect at least a release of `0.8.1` soon after the completion of `0.8.0`.

Thank you all for participating!


ABorgna (approved):

ABorgna:
Merged the changelog updates from #735, and moved the 0.8.0 release notes to the new markdown file

PR #759, opened 2025-04-01
fix!: Correct typing for steiner_tree

daehiff:
# The issue
`steiner_tree` is not usable with other index types then `i32`, hence make the indextype a variable.

Sorry for the previous oversight

BREAKING CHANGE: Will require an additional index type on the `steiner_tree` function


ABorgna (approved):
Good catch. Thanks!

PR #767, opened 2025-04-07
chore: Trim the `hashbrown` dependencies

cuviper:
This disables `hashbrown`'s default features and only enables its `"default-hasher"`. In particular, this trims the `allocator-api2` dependency that is otherwise unused.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ debug = true
 [dependencies]
 fixedbitset = { version = "0.5.7", default-features = false }
 indexmap = { version = "2.5.0", default-features = false }
-hashbrown = { version = "^0.15.0", default-features = false, features = ["default-hasher"] }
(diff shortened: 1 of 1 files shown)


cuviper:
We could also add back the `"inline-more"` feature, if you think that's important for performance.

Cargo.toml
```diff
@@ -43,7 +43,7 @@ debug = true
 [dependencies]
 fixedbitset = { version = "0.5.7", default-features = false }
 indexmap = { version = "2.5.0", default-features = false }
-hashbrown = { version = "^0.15.0" }
+hashbrown = { version = "^0.15.0", default-features = false, features = ["default-hasher"] }
```
starovoid:
```suggestion
hashbrown = { version = "^0.15.0", default-features = false, features = ["default-hasher", "inline-more"] }
```

Thanks for the PR! I think keeping a potential performance improvement would be the right decision.

PR #770, opened 2025-04-15
docs: Specify that Acyclic::try_udpate_edge may add an edge

### LINKED ISSUE #769: `Acyclic::try_add_edge` shouldn't err if the edge is already present
The docs for `try_add_edge` say:

> ... (returns) [`AcyclicEdgeError`] if the edge would create a cycle, a self-loop or if the edge addition failed in the underlying graph.

I'm using `GraphMap` as the underlying graph, where you can add duplicated edges, I don't think any of these conditions are true but I'm getting an error.

Minimal repro:

```rs
fn main() {
    let mut graph = Acyclic::<GraphMap<usize, (), Directed>>::new();
    graph.add_node(0);
    graph.add_node(1);
    graph.try_add_edge(0, 1, ()).unwrap();
    graph.try_add_edge(0, 1, ()).unwrap(); // `Result::unwrap()` on an `Err` value: InvalidEdge
}
```

lmondada:
Closes #769

### DIFF
--- a/src/acyclic.rs
+++ b/src/acyclic.rs
@@ -165,10 +165,14 @@ where
     /// would create a cycle, a self-loop or if the edge addition failed in
     /// the underlying graph.
     ///
-    /// In cases where edge addition cannot fail in the underlying graph (e.g.
-    /// when multi-edges are allowed, as in [`DiGraph`] and [`StableDiGraph`]),
-    /// this will return an error if and only if [`Self::is_valid_edge`]
-    /// returns `false`.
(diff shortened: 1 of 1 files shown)


ABorgna:
We had [an issue](https://github.com/petgraph/petgraph/pull/771) with ci caches. Mind rebasing on that?

lmondada:
Ready for review!

I've expanded the docs to point out the risk of confusion from name clashes between `Build::add_edge` and a graph type's own graph construction methods. I've also created #770 to keep track of this issue, although I do not see an easy fix for it.

ABorgna (approved):

PR #776, opened 2025-04-23
feat: Implement `DataMap` for `GraphMap` graphs

skius:
The `DataMap` implementation is the only trait missing to make `GraphMap` graphs usable with `isomorphism` functions.

The `Vec<usize>` return type of isomorphism matchings is slightly confusing for `GraphMap`, but I think it's fine to include this PR by itself. In the future a specific "MatchResolver" or similar could always be added for `GraphMap`s that turns the `Vec<usize>` (which is using the order of `GraphMap::nodes`) into a `G0::N ==> G1::N` mapping.

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -1367,11 +1367,11 @@ where
     Ty: EdgeType,
     S: BuildHasher,
 {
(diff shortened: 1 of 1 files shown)


starovoid (approved):

starovoid:
A good addition, it will be included in the `0.8.2` release.

PR #783, opened 2025-05-02
Make `Csr::from_sorted_edges` generic over edge type

### LINKED ISSUE #588: Can we remove `Directed` trait constraint?
After check the source code, I think `from_sorted_edges` also apply to undirected csr.

Source:
https://github.com/petgraph/petgraph/blame/2cf4969eefc8001a5135904fce2f8976e57d033d/src/csr.rs#L137
```rust
impl<N, E, Ix> Csr<N, E, Directed, Ix>
where
    Ix: IndexType,
{
    /// Create a new `Csr` from a sorted sequence of edges
    ///
    /// Edges **must** be sorted and unique, where the sort order is the default
    /// order for the pair *(u, v)* in Rust (*u* has priority).
    ///
    /// Computes in **O(|E| + |V|)** time.
    /// # Example
    /// ```rust
    /// use petgraph::csr::Csr;
    /// use petgraph::prelude::*;
    ///
    /// let graph = Csr::<(),()>::from_sorted_edges(&[
    ///                     (0, 1), (0, 2),
    ///                     (1, 0), (1, 2), (1, 3),
    ///                     (2, 0),
    ///                     (3, 1),
    /// ]);
    /// ```
    pub fn from_sorted_edges<Edge>(edges: &[Edge]) -> Result<Self, EdgesNotSorted>
    where
        Edge: Clone + IntoWeightedEdge<E, NodeId = NodeIndex<Ix>>,
        N: Default,
    {
        let max_node_id = match edges
            .iter()
            .map(|edge| {
                let (x, y, _) = edge.clone().into_weighted_edge();
                max(x.index(), y.index())
            })
            .max()
        {
            None => return Ok(Self::with_nodes(0)),
            Some(x) => x,
        };
        let mut self_ = Self::with_nodes(max_node_id + 1);
        let mut iter = edges.iter().cloned().peekable();
        {
            let mut rows = self_.row.iter_mut();

            let mut rstart = 0;
            let mut last_target;
            'outer: for (node, r) in (&mut rows).enumerate() {
                *r = rstart;
                last_target = None;
                'inner: loop {
                    if let Some(edge) = iter.peek() {
                        let (n, m, weight) = edge.clone().into_weighted_edge();
                        // check that the edges are in increasing sequence
                        if node > n.index() {
                            return Err(EdgesNotSorted {
                                first_error: (n.index(), m.index()),
                            });
                        }
                        /*
                        debug_assert!(node <= n.index(),
                                      concat!("edges are not sorted, ",
                                              "failed assertion source {:?} <= {:?} ",
                                              "for edge {:?}"),
                                      node, n, (n, m));
                                      */
                        if n.index() != node {
                            break 'inner;
                        }
                        // check that the edges are in increasing sequence
                        /*
                        debug_assert!(last_target.map_or(true, |x| m > x),
                                      "edges are not sorted, failed assertion {:?} < {:?}",
                                      last_target, m);
                                      */
                        if !last_target.map_or(true, |x| m > x) {
                            return Err(EdgesNotSorted {
                                first_error: (n.index(), m.index()),
                            });
                        }
                        last_target = Some(m);
                        self_.column.push(m);
                        self_.edges.push(weight);
                        rstart += 1;
                    } else {
                        break 'outer;
                    }
                    iter.next();
                }
            }
            for r in rows {
                *r = rstart;
            }
        }

        Ok(self_)
    }
}
```

uulm-janbaudisch:
I am not sure if there was a reason for the constraint, but it seems to work fine for undirected graphs.

Fixes #588.

### DIFF
--- a/src/csr.rs
+++ b/src/csr.rs
@@ -162,8 +162,9 @@ pub struct EdgesNotSorted {
     first_error: (usize, usize),
 }
 
-impl<N, E, Ix> Csr<N, E, Directed, Ix>
+impl<N, E, Ty, Ix> Csr<N, E, Ty, Ix>
 where
+    Ty: EdgeType,
     Ix: IndexType,
 {
     /// Create a new `Csr` from a sorted sequence of edges


starovoid:
I think this was done some times ago due to the requirements of unambiguity of the edge order. 
Thanks for the reminder.

uulm-janbaudisch:
What I noticed when using CSRs for undirected graphs is that edges have to be present in both directions for `from_sorted_edges` to work correctly. This might be important for users. Maybe a note in the docs would be enough?

starovoid:
> Maybe a note in the docs would be enough?

Let's do that.
And to ensure that everything works correctly and avoid future regressions, please write a test for constructing a `Csr` for an undirected graph.

RaoulLuque:
Since this PR seems stale, I was wondering if it would be alright if I would pick it up. That is, fork your branch and finish it such that your work gets merged : )

uulm-janbaudisch:
Sorry for the delay. I added a note to the documentation.
The added test is currently failing. The edge count seems to not be correctly calculated.
@RaoulLuque it would be great if you or someone else could take a look at it. I would be fine with you taking over the PR.

RaoulLuque:
> Sorry for the delay. I added a note to the documentation.
> The added test is currently failing. The edge count seems to not be correctly calculated.
> @RaoulLuque it would be great if you or someone else could take a look at it. I would be fine with you taking over the PR.

No worries at all :) Alright, I see 🤔 I'll look into it then and prob take over the PR 👍 I appreciate the quick response ☺️

RaoulLuque:
Okay, I will close this in favor of #861 then : )

PR #785, opened 2025-05-05
fix!: Conditionally use `std` hash set in impl of `Visitable` for `GraphMap`

### LINKED ISSUE #784: Can't construct `Dfs<_, std::collections::HashSet<_>` for `DiGraphMap`
### Summary

As of 0.8, it is no longer possible to directly construct a `Dfs<_, H>` for `DiGraphMap` (presumably any `GraphMap`?) when `H = std::collections::HashSet`.

#### Repro

The following code compiled and worked on 0.7, but now fails to compile:
```rust
use std::collections::HashSet;
use petgraph::graphmap::DiGraphMap;
use petgraph::visit::Dfs;

fn bug() -> Dfs<u32, HashSet<u32>> {
    let g: DiGraphMap<u32, ()> = DiGraphMap::new();
    Dfs::new(&g, 0u32)
}
```
#### Error
```
error[E0308]: mismatched types
   --> src/lib.rs:10:5
    |
8   | fn bug() -> Dfs<u32, HashSet<u32>> {
    |             ---------------------- expected `Dfs<u32, HashSet<u32>>` because of return type
9   |     let g: DiGraphMap<u32, ()> = DiGraphMap::new();
10  |     Dfs::new(&g, 0u32)
    |     ^^^^^^^^^^^^^^^^^^ expected `HashSet<u32>`, found `hashbrown::set::HashSet<u32>`
    |
    = note: `hashbrown::set::HashSet<u32>` and `HashSet<u32>` have similar names, but are actually distinct types
note: `hashbrown::set::HashSet<u32>` is defined in crate `hashbrown`
   --> /playground/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hashbrown-0.15.2/src/set.rs:114:1
    |
114 | pub struct HashSet<T, S = DefaultHashBuilder, A: Allocator = Global> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: `HashSet<u32>` is defined in crate `std`
   --> /playground/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/collections/hash/set.rs:125:1
    |
125 | pub struct HashSet<T, S = RandomState> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0308`.
```

### Analysis

The cause appears to be [this change](https://github.com/petgraph/petgraph/pull/747/files#diff-5c7f2bec4347c157a23a27560c059601d3ff3a348c46b335b98b6e9cf761470dR16), where the imported `HashSet` was changed to `hashbrown::HashSet`.

That import becomes observable [here](https://github.com/bushrat011899/petgraph/blob/f5399a4cc10b9936a4912a6c61161927c06825e3/src/graphmap.rs#L1154), where use the imported `HashSet` as the associated `Visitable::Map`.

Related to #763. That case could be fixed by just adding an additional impl, but in this case, I think the change may need to be breaking.

ranweiler:
Fixes #784.

This is a very narrow fix. A more comprehensive fix could be to create a `hash` subdomule that encapsulates the `std`-conditional choice of `HashSet`/`HashMap` and then require all modules to import from that. This would catch any other similarly-leaked changes of hash containers. I'll leave that to your judgement.

BREAKING CHANGE: when `std` is enabled, impl `Visitable<Map = std::collections::HashSet>` for `GraphMap`.

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -13,7 +13,6 @@ use core::{
     slice::Iter,
 };
 
-use hashbrown::HashSet;
 use indexmap::{
     map::{Iter as IndexMapIter, IterMut as IndexMapIterMut, Keys},
     IndexMap,
@@ -24,6 +23,12 @@ use crate::{
     graph::{node_index, Graph},
     visit, Directed, Direction, EdgeType, Incoming, IntoWeightedEdge, Outgoing, Undirected,
 };
+
+#[cfg(feature = "std")]
+use std::collections::hash_set::HashSet;
+
+#[cfg(not(feature = "std"))]
+use hashbrown::HashSet;
 
 #[cfg(feature = "std")]
 use std::collections::hash_map::RandomState;


RaoulLuque:
I think, this is a great suggestion, however I would prefer to implement it directly for all imports, as you suggested. Using a central point in petgraph from where other files import their HashMaps / etc needs ^^ Thus, if you don't want to implement that, I would probably open a PR soon with that and close this one in favor of it :)

ranweiler:
Sorry @RaoulLuque, I just now saw a notification for your comment! I'll switch to the module refactoring strategy, and we can close this PR once I send the new one.

RaoulLuque:
Hi, no worries at all :) Note the issue #891 which announces some major breaking changes soon, among other things also big refactorings of the current code base. Just as a heads up, that it might be worthwhile postponing working on such a refactor until when the 0.9 is in the making - to avoid doing work twice :)

ranweiler:
@RaoulLuque, thanks for the heads-up on the roadmap. Feel free to tag me in when it makes sense to tackle that. I'll just close this out since you've captured the followup work in #892.

RaoulLuque:
> @RaoulLuque, thanks for the heads-up on the roadmap. Feel free to tag me in when it makes sense to tackle that. I'll just close this out since you've captured the followup work in #892.

Will do :) Thanks for your work and letting us know about the bug / issue in the first place :heart_hands:

Issue #637, opened 2024-04-24
Document iterating order


vporton:
[Docs](https://docs.rs/petgraph/latest/petgraph/graph/struct.Graph.html#method.neighbors) for `neighbors`, `neighbors_directed`, and `neighbors_undirected`, as well as `edges`, `edges_directed`, `edges_connecting` should tell about the iterating order.

Particularly, is iterating order non-deterministic?

FliegendeWurst:
As far as I can tell the iteration order is always the reverse of the insertion order (if you never remove edges, not sure what happens if you do).

PR #790, opened 2025-05-18
docs: Specify iteration order for neighbors and edges and their variants

RaoulLuque:
This should fix https://github.com/petgraph/petgraph/issues/637.

I tried my best to deduce the iteration order from the code and various tests. It should also be considered, whether one actually wants to "fix" the iteration order in the docs of the respective methods. That is, whether there should be a guarantee for the user that this iteration may not change in the future.

Another possibility would be to add the iteration order, as suggested in this PR with a warning that this iteration might change in the future.

On a similar note, would it make sense to create tests that check whether this guarantee holds, to make sure that, also in the future, the docs of the methods actually reflect the reality?


starovoid:
Hi, please also fix clippy warnings

RaoulLuque:
Hi, not entirely sure, but the clippy errors don't seem to stem from my changes. Therefore I am not sure how to fix them ? Could there be an issue with clippy/rust nightly ?

starovoid:
I fixed the warnings in #791, now everything should be fine.

starovoid:
It also took me a long time to overread the code, but it seems to me that you have written everything correctly. The iteration order is something that may change one day in the future, but since users prefer to see a note about the iteration order, we really should make some efforts to maintain clarity.

So (to prove the words about the iteration order), I suggest writing the some tests:
1. one small test for each case;

starovoid:
> Could you maybe specify a bit more what you had in mind for the randomized tests? 

I mean the tests where we create a graph and remember the order of edge insertion, then iterate over the edges/neighbors comparing the orders. Repeat each such check several times, processing a new random graph each time. 
To generate random graphs, there are functions in `petgraph` or `graphalgs`. Feel free to adopt it.

RaoulLuque:
Okay, I see, I am just hesitant with combining the randomness and remembering the order of edge insertion, but I'll work something out I guess :+1:

RaoulLuque:
Okay, so I added unit tests for `Graph` and `StableGraph`, both directed and undirected. I tried to come up with Quickchecks, but nothing seemed very sensible in my opinion.

The problem being that one does not know what edges the arbitrary graph has. And this will definitely influence the order in which neighbors may appear in the different iterators. I tried adding fixed edges to the arbitrary graphs and then checking their order, but these edges might already exist which would confuse the ordering. Of course, one could select certain pairs of nodes, remove existing edges between them and then add edges there and check their order, but at that point it seems somewhat useless to use quickcheck in the first place :thinking: 

What do you think @starovoid ? Do you have a specific idea on how the quickchecks should look like or would you also be fine with "just" having unit tests, which test for the iteration order of the different functions ?

starovoid:
> What do you think @starovoid ? Do you have a specific idea on how the quickchecks should look like or would you also be fine with "just" having unit tests, which test for the iteration order of the different functions ?

Hi! Unit tests look great, I think it makes no sense to overcomplicate the task using quickcheck. 
`+1330` lines of code, I'm impressed 😃

RaoulLuque:
Oops, just messed something up with the rebase I guess :thinking:

RaoulLuque:
This is failing because of https://github.com/petgraph/petgraph/pull/878, could you have a look at the other PR such that we can fix the errors in our CI @starovoid  :)?

PR #794, opened 2025-05-20
feat: Add `map_owned` method for `Graph`

### LINKED ISSUE #542: Add another `map` method taking `self` instead of `&self`
### Summary

Add another `map` method taking `self` instead of `&self`.

### Motivation

This helps when the node weight or edge weight is some data structure that needs to be moved to another graph.

### Details

Add a new method (`map_owned`?) to `Graph` and `StableGraph`.

No.

  implement it?

I can implement it myself. Any suggestions on the method name?

petehayes102:
Closes #542

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -1493,6 +1493,31 @@ where
         g
     }
 
+    /// Create a new `Graph` by mapping node and edge weights to new values,
+    /// consuming the current graph.
+    ///
+    /// The resulting graph has the same structure and the same graph indices
+    /// as `self`.
+    ///
+    /// If you want a non-destructive version of this function, see `map`.
+    pub fn map_owned<F, G, N2, E2>(self, mut node_map: F, mut edge_map: G) -> Graph<N2, E2, Ty, Ix>
+    where
+        F: FnMut(NodeIndex<Ix>, N) -> N2,
+        G: FnMut(EdgeIndex<Ix>, E) -> E2,
+    {
+        let mut g = Graph::with_capacity(self.node_count(), self.edge_count());
+        g.nodes.extend(enumerate(self.nodes).map(|(i, node)| Node {
+            weight: node_map(NodeIndex::new(i), node.weight),
+            next: node.next,
+        }));
+        g.edges.extend(enumerate(self.edges).map(|(i, edge)| Edge {
+            weight: edge_map(EdgeIndex::new(i), edge.weight),
+            next: edge.next,
+            node: edge.node,
+        }));
+        g
+    }
+
     /// Create a new `Graph` by mapping nodes and edges.
     /// A node or edge may be mapped to `None` to exclude it from
     /// the resulting graph.
--- a/tests/graph.rs
+++ b/tests/graph.rs
@@ -1597,6 +1597,39 @@ fn test_has_path() {
 }
 
 #[test]
+fn map_map_owned() {
+    let mut g = Graph::new_undirected();
+    let a = g.add_node("A");
+    let b = g.add_node("B");
+    let c = g.add_node("C");
+    let d = g.add_node("D");
+    let e = g.add_node("E");
+    let f = g.add_node("F");
+    let ab = g.add_edge(a, b, 7);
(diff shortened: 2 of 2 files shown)


src/graph_impl/mod.rs
```diff
@@ -1493,6 +1493,31 @@ where
         g
     }
 
+    /// Create a new `Graph` by mapping node and edge weights to new values,
+    /// consuming the current graph.
+    ///
+    /// The resulting graph has the same structure and the same graph indices
+    /// as `self`.
+    ///
+    /// If you want a non-destructive version of this function, see `map`.
```
RaoulLuque, line 1502:
```suggestion
    /// If you want a non-consuming version of this function, see `map`.
```
I think, I would prefer to write non-consuming instead of destructive, since destruction might imply more than just giving up ownership of the value.

src/graph_impl/mod.rs
```diff
@@ -1493,6 +1493,31 @@ where
...
+    pub fn map_owned<F, G, N2, E2>(self, mut node_map: F, mut edge_map: G) -> Graph<N2, E2, Ty, Ix>
```
RaoulLuque, line 1503: Possible instead of `map_owned` we could call this method `into_map` aligning with naming conventions from std, such as `into_iter` or `into_bytes`. On the other hand, we could also go for `map_into` which is also used in itertools for example:
https://github.com/rust-itertools/itertools/blob/b0942d6b84d47dec75c19e541859211423d857d9/src/adaptors/map.rs#L125

`map_into` would make sense since we don't actually convert into a map (into an interator) but instead convert to a graph immediately (similar to applying `collect` to an iterator in the iterator analogy). Therefore we 'map' 'into' another graph. `into_map` on the other hand would suggest that we would return a map of some sort and consume the provided graph.

tests/graph.rs
```diff
@@ -1596,6 +1596,27 @@ fn test_has_path() {
...
+    assert_eq!(g2.node_count(), 3);
+    assert_eq!(g2.node_weight(a).map(|s| &**s), Some("map-A"));
+    assert_eq!(g2.node_weight(b).map(|s| &**s), Some("map-B"));
+    assert_eq!(g2.node_weight(c).map(|s| &**s), Some("map-C"));
+    assert_eq!(g2.edge_count(), 3);
+    assert_eq!(g2.edge_weight(ab), Some(&14));
+    assert_eq!(g2.edge_weight(bc), Some(&28));
+    assert_eq!(g2.edge_weight(ca), Some(&18));
+}
+
```
RaoulLuque, line 1619: Not necessarily your concern but I am just noticing that we don't have any (normal) map tests for graph. So maybe add one for that as well?

RaoulLuque (commented):
Hey, first off, thanks for the PR 🦕

Seems like a good addition, I just have a few notes :)

One thing I didn't include in the code review:
Is there any reason why this wouldn't be implemented for StableGraphs while we're at it?

Additionally, in the [original issue](https://github.com/petgraph/petgraph/issues/542#issuecomment-1544246603) it was suggested to add similar functionality for `filter_map`, so maybe add that as well?

starovoid:
I suggest also implement `filter_map_owned`, and then I'll include it all in `0.8.3`.

RaoulLuque:
Referenced this in the marked PRs by accident, sry ^^

RaoulLuque:
Since this PR seems stale, I was wondering if it would be alright if I would pick it up. That is, fork your branch and finish it such that your work gets merged : )

itpetey:
@RaoulLuque Hey mate sorry. Been really busy over the last...yeesh, over a month. I'd really appreciate you picking it up :) Thanks!!

RaoulLuque:
> @RaoulLuque Hey mate sorry. Been really busy over the last...yeesh, over a month. I'd really appreciate you picking it up :) Thanks!!

Hey, no worries at all :) Just wanted to make sure I'd ask in order not step on anyone's toes 🌞 Appreciate your quick response :)

RaoulLuque:
Closed in favor of #863.

PR #800, opened 2025-05-27
fix: Run Maximal Cliques Quickcheck only on Digraphs which are symmetrical

RaoulLuque:
Splitted off from https://github.com/petgraph/petgraph/pull/798. See there for details

### DIFF
--- a/src/algo/maximal_cliques.rs
+++ b/src/algo/maximal_cliques.rs
@@ -6,6 +6,13 @@ use hashbrown::HashSet;
 
 /// Finds maximal cliques containing all the vertices in r, some of the
 /// vertices in p, and none of the vertices in x.
+///
+/// By default, only works on undirected graphs. It can be used on directed graphs
+/// if the graph is symmetric. I.e., if an edge (u, v) exists, then (v, u) also exists.
+///
+/// Uses the [Bron–Kerbosch algorithm][1] with pivoting.
+///
+/// [1]: https://en.wikipedia.org/wiki/Bron%E2%80%93Kerbosch_algorithm
 fn bron_kerbosch_pivot<G>(
     g: G,
     adj_mat: &G::AdjMatrix,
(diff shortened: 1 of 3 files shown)


starovoid:
I had an alternative idea, which was to use an [`UndirectedAdaptor`](https://docs.rs/petgraph/latest/petgraph/visit/struct.UndirectedAdaptor.html). But this requires breaking API changes, so for now it's just an idea for further major release.
In the future, it might look something like this:

```rust
impl<G> GetAdjacencyMatrix for UndirectedAdaptor<G>
where
    G: GetAdjacencyMatrix,
{
    type AdjMatrix = G::AdjMatrix;

    fn adjacency_matrix(&self) -> Self::AdjMatrix {
        self.0.adjacency_matrix()
    }

    fn is_adjacent(&self, matrix: &Self::AdjMatrix, a: G::NodeId, b: G::NodeId) -> bool {
        self.0.is_adjacent(matrix, a, b) || self.0.is_adjacent(matrix, b, a)
    }
}
```

```rust
pub fn maximal_cliques<G>(g: G) -> Vec<HashSet<G::NodeId>>
where
    G: GetAdjacencyMatrix + IntoNodeIdentifiers + IntoNeighbors + IntoNeighborsDirected,
    G::NodeId: Eq + Hash,
{
    let g = UndirectedAdaptor(g);
    let adj_mat = g.adjacency_matrix();
    let r = HashSet::new();
    let p = g.node_identifiers().collect::<HashSet<G::NodeId>>();
    let x = HashSet::new();
    bron_kerbosch_pivot(g, &adj_mat, r, p, x)
}
```

The breaking change occurs because we require `IntoNeighborsDirected` for the `UndirectedAdaptor`. I will open an issue with this proposal a little later, but with a high chance it will lose relevance after the upcoming big refactoring.

PR #801, opened 2025-05-27
fix: Run Steiner Tree Quickcheck on the connected components to properly support disconnected graphs

RaoulLuque:
Splitted off from https://github.com/petgraph/petgraph/pull/798. See there for details

### DIFF
--- a/tests/quickcheck.rs
+++ b/tests/quickcheck.rs
@@ -63,7 +63,6 @@ where
 
 use core::fmt;
 use petgraph::algo::articulation_points::articulation_points;
-use petgraph::unionfind::UnionFind;
 
 quickcheck! {
     fn mst_directed(g: Small<Graph<(), u32>>) -> bool {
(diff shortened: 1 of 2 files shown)


RaoulLuque:
@starovoid , maybe it's me but I don't see a review or anything of the sort.

tests/quickcheck.rs
```diff
@@ -1507,22 +1507,72 @@ quickcheck! {
...
+                        if component.contains(&(edge.0)) && component.contains(&(edge.1)) {
+                            Some(*edge_weight)
+                        } else {
+                            None
+                        }
+                    },
+                );
 
-        spans_terminals
+                let mut terminals = g.node_indices().collect::<Vec<_>>();
```
starovoid, line 1558: Why not simplify 
```rust
let mut terminals = g.node_indices().collect::<Vec<_>>();
terminals = terminals.into_iter().take(5).collect();
```

to
```rust 
let terminals = g.node_indices().take(5).collect::<Vec<_>>();
```

PR #808, opened 2025-05-31
docs: Add examples to minimum spanning tree functions

RaoulLuque:
This PR simply adds examples (really the same one for both) to the minimum spanning tree functions (the default one and the prim variant). Adding these makes them more consistent with the rest of the doc strings for other functions, since most of them include examples of how to use them.

This seems particularly useful, since in case of computing msts with petgraph, one needs to use `from_elements`, which is mentioned in the docstring already, but it might not be clear how to actually use `from_elements` to obtain a graph.

### DIFF
--- a/src/algo/min_spanning_tree.rs
+++ b/src/algo/min_spanning_tree.rs
@@ -24,9 +24,58 @@ use crate::visit::{IntoEdgeReferences, NodeIndexable};
 ///
 /// Use `from_elements` to create a graph from the resulting iterator.
 ///
-/// See also: [`.min_spanning_tree_prim(g)`][1] for implementation using Prim's algorithm.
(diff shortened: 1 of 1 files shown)


src/algo/min_spanning_tree.rs
```diff
@@ -141,6 +190,55 @@ where
 /// See also: [`.min_spanning_tree(g)`][1] for implementation using Kruskal's algorithm and support for minimum spanning forest.
```
starovoid, line 190: A small fix:

```suggestion
/// See also: [`min_spanning_tree`][1] for an implementation using Kruskal's algorithm and support for minimum spanning forest.
```

starovoid (changes requested):

starovoid:
A nice addition to the upcoming release. Thanks!

PR #809, opened 2025-05-31
ci: Remove one test from miri test

RaoulLuque:
There is one test, namely b07_vienna_test which seems to take ~5-10 minutes for each run of miri. Therefore, I would suggest excuding this test when running miri, since the test is also run in the various 'normal' cargo test runs.

For the workflows of this PR, a significant speedup should already be seen.


starovoid:
Where can we read about the purpose and design of this test? I like the idea of accelerating CI, but I want to make sure that we don't lose coverage.

PR #810, opened 2025-05-31
ci: Speedup miri by using nextest

RaoulLuque:
[Nextest](https://github.com/nextest-rs/nextest) is an alternative test runner for rust projects which should speedup the tests, which is particularly interesting for the miri test run, since that one takes very long in comparison to the others.

I also included the changes from PR #809 already. If those should not be desireable, I can adjust the PR accordingly. However, the speedup should be even more noticeable if the test as suggest in PR #809 is removed, since most notably, nextest runs some tests in parallel, which is not very helpful if there are single tests which take extraordinarily long. 

There is an [integration with miri](https://nexte.st/docs/integrations/miri/) which is used for this PR, for which there is one thing to note however:

"Note, however, that `cargo miri test` is able to detect data races where two tests race on a shared resource. Miri with nextest will not detect such races."

I thought that covering such cases is not important for `petgraph`, but I am open to discussion on that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: '*'
   merge_group:
     types: [ checks_requested ]
-  workflow_dispatch: {} 
+  workflow_dispatch: { }
 
 name: Continuous integration
 
@@ -120,4 +120,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@nightly
         with:
           components: miri
-      - run: cargo miri test
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: nextest
+      - run: cargo miri nextest test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,4 +123,4 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest
-      - run: cargo miri nextest test
+      - run: cargo miri nextest run
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,4 +123,5 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest
-      - run: cargo miri nextest run
+      # Exclude the `b07_vienna_test` test, as it takes very long to run with miri
+      - run: cargo miri nextest run -- --skip b07_vienna_test


RaoulLuque:
Update: As can be seen, this leads to a runtime increase the miri CI job from [14 minutes](https://github.com/petgraph/petgraph/actions/runs/15363867299/job/43234281588?pr=809) to 11 minutes.

starovoid:
> "Note, however, that `cargo miri test` is able to detect data races where two tests race on a shared resource. Miri with nextest will not detect such races."
> 
> I thought that covering such cases is not important for `petgraph`, but I am open to discussion on that.

I think we still need to cover potential data races, since it is planned to add a large number of parallel algorithms in the future, some of which may contain unsafe optimizations. 

Therefore, switching to `nextest` seems like an ambiguous decision and requires additional discussions, and we probably won't have time to include it in the next release.

ABorgna:
This is great, I agree that sometimes we may want the extra miri detection but it's not required for every commit made to a PR.

Since there's no code change needed, there's some options here,
- Run the regular `miri test` only for commits to `main` (i.e. `if: github.event_name == 'push'`). This lets us speedup development, while still getting notified about failing checks.
- Add a special PR label that triggers CI to run `test` instead of `nextest`. This is a bit annoying since it requires remembering to manually mark PRs, and re-run CI afterwards.

I say we should do (1), as the CI running time after merging to `main` is not too important.
(2) I'm not so sure, I don't think it's required.

> we probably won't have time to include it in the next release.

Note that this doesn't affect the released crate in any way, so it shouldn't be considered for the release anyways

starovoid:
>   Run the regular `miri test` only for commits to `main` (i.e. `if: github.event_name == 'push'`). This lets us speedup development, while still getting notified about failing checks.

That sounds smart, let's do it.

> Note that this doesn't affect the released crate in any way, so it shouldn't be considered for the release anyways

With the new offer, we are not losing our test coverage, so we really don't have to wait any longer.


@RaoulLuque Can you prepare this CI update? Especially in the new position 🎉

RaoulLuque:
That does indeed sound like a very neat solution :)


With your suggested solution I am wondering whether we should/could exclude more of the very slow steiner tree /graph6 encoding tests from miri when developing ? (Like the one in [this PR](https://github.com/petgraph/petgraph/pull/809) ).
Among would be the following (timings are from my local machine using `ts` from `multitools` - CI times seem to be ~3/4 as long as the ones on my local machine):
```
graph6 encoding:
~40s:  generic_graph6_encoder_test_cases
~160s: graph6_for_csr_test_cases
~196s: graph6_for_graph_map_test_cases
~100s: graph6_for_graph_test_cases
~95s:  graph6_for_matrix_graph_test_case
~100s: graph6_for_stable_graph_test_cases

Steiner trees:
~160s: b01_vienna_test
~510s: b07_vienna_test (the test removed in the PR linked above)
```
Please let me know whether you think we should exclude more and which ones :)

I suppose then I would also make the '`Push` run' include all tests even the one excluded in the PR linked above?

ABorgna:
That list looks fine 👍

> I suppose then I would also make the 'Push run' include all tests even the one excluded in the PR linked above?

Yes, commits to `master` should run everything if possible.

Also, related to #811. Do you know if other projects have "flag labels" for things like this? We should do something like a `F-run-full-CI` to unskip the tests here. (I can add it afterwards if you prefer to merge without the skip).

RaoulLuque:
Huh, that's a good question. I actually don't know how other projects handle that :thinking: But it sounds like a good idea ! It should definitely be possible. I guess it would be nice to be able to run the CI before actually merging and not having the 'breaking' changes already on master when one notices them

RaoulLuque:
I think I'll create a new PR for that anyways since I predict that it'll take some fiddling around to obtain the desired behaviour. Do you know of any CI plugin which could achieve that or should we custom write some bash fetching of the PR labels?

ABorgna:
>  think I'll create a new PR for that

👍 

> Do you know of any CI plugin which could achieve that or should we custom write some bash fetching of the PR labels?

It can be done directly in github actions conditionals.
This should do it (gh actions syntax is... not the best).
```yml
extra-miri:
    if: ${{ github.event_name != 'merge_group' && ( github.event_name != 'pull_request' || contains( github.event.pull_request.labels.*.name, 'run-ci-checks') ) }}
        ...
```

RaoulLuque:
Wow, this landed us at a duration for the miri checks of right below 6 mins :tada:

RaoulLuque:
I'm gonna test now whether assigning the new label to the PR will trigger the other CI run as expected

RaoulLuque:
Okay, so after some further testing, it seems to be working as intended now. I also already included the conditional triggers based on labels etc. ^^

Here is the first run: https://github.com/petgraph/petgraph/actions/runs/15810905808/job/44562063248
This is the second run when I added the label (more than 3 times longer): https://github.com/petgraph/petgraph/actions/runs/15810964877/job/44562186832
And this is the one after I removed the label again: https://github.com/petgraph/petgraph/actions/runs/15811138729/job/44562571257?pr=810

We'll have to see whether the specification to run on `push` works as well once we merge I guess ^^

If you have any suggestions to renaming the label or anything, I am open to that :)

Otherwise this should be ready @ABorgna , @starovoid  :sauropod:

.github/workflows/ci.yml
```diff
@@ -110,15 +110,41 @@ jobs:
       - name: Check
         run: cargo check --no-default-features -p petgraph --target wasm32v1-none --features graphmap,serde-1,stable_graph,matrix_graph,generate,unstable
 
-  miri:
-    name: Unsoundness check
-    if: github.event_name != 'merge_group'
+  fast-miri:
+    # This job runs on pull_request without the `S-run-thorough-ci-tests` label.
+    name: Unsoundness check fast
+    if: ${{ github.event_name != 'merge_group' && github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'S-run-thorough-ci-tests') }}
```
ABorgna: Can be removed, unless you wanted to keep it explicitly. (same with the other `if`)
```suggestion
    if: ${{ github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'S-run-thorough-ci-tests') }}
```

ABorgna (approved):
:shipit:

.github/workflows/ci.yml
RaoulLuque: Lol, yeah 😅 Boolean logic ... 🤷🏻🤦‍♂️

RaoulLuque:
Just tested one last time whether it behaves as expected, looking good :)

Issue #806, opened 2025-05-30
A* should tie-break on lower h-values


Dietr1ch:
Ranking with min `(f, h)` instead of min `f` tends to do fewer expansions.

The idea is that among the many search nodes with best f-value, the ones further away (higher g, lower h values) are the ones that will move the fringe closer to a goal.

Having the same f-values on many nodes is common, although it's not easy to describe exactly when there's a lot of ties in a given search problem.

---

https://github.com/petgraph/petgraph/blob/master/src/algo/astar.rs#L137

RaoulLuque:
Sounds like a very good idea to me :) 

Would you maybe feel comfortable implementing this? Should be relatively simple since h = f - g. Possibly you would need to implement some traits for MinScored to work but should work out of the box I think :)

If so, it would be very interesting to see a before/after benchmark comparison 🦕

Dietr1ch:
Added a synthetic benchmark on a 500x500 empty maze (4-connected) which is a pretty good case since every node in the rectangle between the start and goal has the same f-value. With this tweak only the diagonal is expanded, instead of maybe even running like Dijkstra's within the rectangle and expanding every node there.

These are not the same commits as my branches since I rebased things out of my nix flake base commit.

This results in a ~20x speedup in the benchmark, but since this essentially compares `n^2` against `n` you can make it as big as you want by using a larger maze.

#### `762a1a5` Bench: Add pathfinding benchmark
test maze_astar    ... bench:   3,936,871.05 ns/iter (+/- 277,951.85)    `(100%)`
test maze_dijkstra ... bench:  15,274,596.40 ns/iter (+/- 389,289.97)


#### `8dd941f` Alg,A*: Tie break ranking on lower h-values
test maze_astar    ... bench:     193,009.44 ns/iter (+/- 8,694.86)      `(4.9%)`
test maze_dijkstra ... bench:  15,360,808.50 ns/iter (+/- 243,655.03)

RaoulLuque:
Resolved in #882

PR #813, opened 2025-06-03
perf: Make A* tie break on lower h-values

Dietr1ch:
This implements the trick from #806 and adds an empty 2d maze benchmark

Update: This dropped original breaking change of making cost be `core::ops::Sub`, so it can be merged earlier.

### DIFF
--- a/benches/astar.rs
+++ b/benches/astar.rs
@@ -1,0 +1,103 @@
+#![feature(test)]
+
+extern crate petgraph;
+extern crate test;
+
+use petgraph::prelude::*;
+use test::Bencher;
+
+mod common;
+use petgraph::algo::astar;
+
+#[bench]
+fn astar_2d_grid_line_bench(bench: &mut Bencher) {
+    // Same line grid as ./dijkstra.rs
+    const N: usize = 500;
+    const M: usize = 5;
+    let g = common::build_2d_grid(N, M);
+
+    let heuristic_for = |goal: NodeIndex| {
+        let g = &g;
+        move |node: NodeIndex| -> usize {
+            let (x1, y1): (usize, usize) = g[node];
+            let (x2, y2): (usize, usize) = g[goal];
+
+            x2.abs_diff(x1) + y2.abs_diff(y1)
+        }
+    };
+
+    let start_i = 0;
+    let goal_i = N * M - 1;
+    let start = NodeIndex::new(start_i);
+    let goal = NodeIndex::new(goal_i);
+
+    if let Some((cost, path)) = astar(
+        &g,
+        start,
+        |n| n == goal,
+        |e| *e.weight(),
+        heuristic_for(goal),
+    ) {
+        assert_eq!(cost, N + M - 2);
+
+        assert_eq!(path.first().unwrap().index(), start_i);
+        assert_eq!(path.last().unwrap().index(), goal_i);
+        assert_eq!(path.len(), cost + 1);
+    }
+
+    bench.iter(|| {
+        let _ = astar(
+            &g,
+            start,
+            |n| n == goal,
+            |e| *e.weight(),
+            heuristic_for(goal),
+        );
+    });
+}
+
+#[bench]
+fn astar_2d_grid_bench(bench: &mut Bencher) {
+    // Same 2D grid as ./dijkstra.rs
+    const N: usize = 500;
+    let g = common::build_2d_grid(N, N);
+
+    let heuristic_for = |goal: NodeIndex| {
+        let g = &g;
+        move |node: NodeIndex| -> usize {
+            let (x1, y1): (usize, usize) = g[node];
+            let (x2, y2): (usize, usize) = g[goal];
+
+            x2.abs_diff(x1) + y2.abs_diff(y1)
+        }
+    };
+
+    let start_i = 0;
+    let goal_i = N * N - 1;
+    let start = NodeIndex::new(start_i);
+    let goal = NodeIndex::new(goal_i);
+
+    if let Some((cost, path)) = astar(
+        &g,
+        start,
+        |n| n == goal,
+        |e| *e.weight(),
+        heuristic_for(goal),
+    ) {
+        assert_eq!(cost, 2 * (N - 1));
+
+        assert_eq!(path.first().unwrap().index(), start_i);
+        assert_eq!(path.last().unwrap().index(), goal_i);
+        assert_eq!(path.len(), cost + 1);
+    }
+
+    bench.iter(|| {
+        let _ = astar(
+            &g,
+            start,
+            |n| n == goal,
+            |e| *e.weight(),
+            heuristic_for(goal),
+        );
+    });
+}
--- a/benches/common/factories.rs
+++ b/benches/common/factories.rs
@@ -355,3 +355,24 @@ pub fn build_graph(node_count: usize, dense: bool) -> Graph<usize, i32, Undirect
     }
     graph
 }
+
+pub fn build_2d_grid(width: usize, height: usize) -> Graph<(usize, usize), usize, Undirected> {
+    let mut g = Graph::new_undirected();
+
+    for j in 0..height {
+        for i in 0..width {
+            let node = g.add_node((i, j));
+
+            if j > 0 {
+                let up = NodeIndex::new(node.index() - width);
+                g.add_edge(node, up, 1);
+            }
+            if i > 0 {
+                let right = NodeIndex::new(node.index() - 1);
+                g.add_edge(node, right, 1);
+            }
+        }
+    }
+
+    g
+}
--- a/benches/dijkstra.rs
+++ b/benches/dijkstra.rs
@@ -33,7 +33,6 @@ fn dijkstra_bench(bench: &mut Bencher) {
     });
 }
 
-
 #[bench]
 fn dijkstra_2d_grid_line_bench(bench: &mut Bencher) {
     // Same line grid as ./astar.rs
--- a/benches/dijkstra.rs
+++ b/benches/dijkstra.rs
@@ -3,10 +3,11 @@
 extern crate petgraph;
 extern crate test;
 
+use core::cmp::{max, min};
 use petgraph::prelude::*;
-use std::cmp::{max, min};
 use test::Bencher;
 
+mod common;
 use petgraph::algo::dijkstra;
 
 #[bench]
@@ -29,5 +30,45 @@ fn dijkstra_bench(bench: &mut Bencher) {
 
     bench.iter(|| {
         let _scores = dijkstra(&g, nodes[0], None, |e| *e.weight());
+    });
+}
+
+
+#[bench]
+fn dijkstra_2d_grid_line_bench(bench: &mut Bencher) {
+    // Same line grid as ./astar.rs
+    const N: usize = 500;
+    const M: usize = 5;
+    let g = common::build_2d_grid(N, M);
+
+    let start_i = 0;
+    let goal_i = N * M - 1;
+    let start = NodeIndex::new(start_i);
+    let goal = NodeIndex::new(goal_i);
+
+    let costs = dijkstra(&g, start, None, |e| *e.weight());
+    assert_eq!(costs[&goal], N + M - 2);
+
+    bench.iter(|| {
+        let _ = dijkstra(&g, start, None, |e| *e.weight());
+    });
+}
+
+#[bench]
+fn dijkstra_2d_grid_bench(bench: &mut Bencher) {
+    // Same 2D grid as ./astar.rs
+    const N: usize = 500;
+    let g = common::build_2d_grid(N, N);
+
+    let start_i = 0;
+    let goal_i = N * N - 1;
+    let start = NodeIndex::new(start_i);
+    let goal = NodeIndex::new(goal_i);
+
+    let costs = dijkstra(&g, start, None, |e| *e.weight());
+    assert_eq!(costs[&goal], 2 * (N - 1));
+
+    bench.iter(|| {
+        let _ = dijkstra(&g, start, None, |e| *e.weight());
     });
 }
--- a/src/algo/astar.rs
+++ b/src/algo/astar.rs
@@ -91,61 +91,69 @@ where
     H: FnMut(G::NodeId) -> K,
     K: Measure + Copy,
 {
+    // The Open set
     let mut visit_next = BinaryHeap::new();
-    let mut scores = HashMap::new(); // g-values, cost to reach the node
-    let mut estimate_scores = HashMap::new(); // f-values, cost to reach + estimate cost to goal
+    // A node -> (f, h, g) mapping
+    // TODO: Derive `g` from `f` and `h`.
+    let mut scores = HashMap::new();
+    // The search tree
     let mut path_tracker = PathTracker::<G>::new();
 
-    let zero_score = K::default();
-    scores.insert(start, zero_score);
-    visit_next.push(MinScored(estimate_cost(start), start));
+    let zero: K = K::default();
+    let g: K = zero;
+    let h: K = estimate_cost(start);
+    let f: K = g + h;
+    scores.insert(start, (f, h, g));
+    visit_next.push(MinScored((f, h, g), start));
 
-    while let Some(MinScored(estimate_score, node)) = visit_next.pop() {
+    while let Some(MinScored((f, h, g), node)) = visit_next.pop() {
         if is_goal(node) {
             let path = path_tracker.reconstruct_path_to(node);
-            let cost = scores[&node];
-            return Some((cost, path));
+            let (goal_f, goal_h, goal_g) = scores[&node];
+            debug_assert_eq!(goal_h, zero);
+            debug_assert_eq!(goal_f, goal_g);
+            return Some((goal_f, path));
         }
 
-        // This lookup can be unwrapped without fear of panic since the node was necessarily scored
-        // before adding it to `visit_next`.
-        let node_score = scores[&node];
-
-        match estimate_scores.entry(node) {
+        match scores.entry(node) {
             Occupied(mut entry) => {
-                // If the node has already been visited with an equal or lower score than now, then
-                // we do not need to re-visit it.
-                if *entry.get() <= estimate_score {
+                let (_, _, old_g) = *entry.get();
+                // The node has already been expanded with a better cost.
+                if old_g < g {
                     continue;
                 }
-                entry.insert(estimate_score);
+                // NOTE: Because there's no closed set, we don't know if we expanded this node.
+                // if old_g = g we may be re-expanding this node, but won't insert new neigbours.
+                entry.insert((f, h, g));
             }
             Vacant(entry) => {
-                entry.insert(estimate_score);
+                entry.insert((f, h, g));
             }
         }
 
         for edge in graph.edges(node) {
-            let next = edge.target();
-            let next_score = node_score + edge_cost(edge);
+            let neigh = edge.target();
+            let neigh_g = g + edge_cost(edge);
+            let neigh_h = estimate_cost(neigh);
+            let neigh_f = neigh_g + neigh_h;
+            let neigh_score = (neigh_f, neigh_h, neigh_g);
 
-            match scores.entry(next) {
+            match scores.entry(neigh) {
                 Occupied(mut entry) => {
-                    // No need to add neighbors that we have already reached through a shorter path
-                    // than now.
-                    if *entry.get() <= next_score {
+                    let (_, _, old_neigh_g) = *entry.get();
+                    if neigh_g >= old_neigh_g {
+                        // New cost isn't better
                         continue;
                     }
-                    entry.insert(next_score);
+                    entry.insert(neigh_score);
                 }
                 Vacant(entry) => {
-                    entry.insert(next_score);
+                    entry.insert(neigh_score);
                 }
             }
 
-            path_tracker.set_predecessor(next, node);
-            let next_estimate_score = next_score + estimate_cost(next);
-            visit_next.push(MinScored(next_estimate_score, next));
+            path_tracker.set_predecessor(neigh, node);
+            visit_next.push(MinScored(neigh_score, neigh));
         }
     }
 
--- a/tests/graph.rs
+++ b/tests/graph.rs
@@ -501,7 +501,7 @@ fn test_astar_manhattan_heuristic() {
     let b = g.add_node((2., 0.));
     let c = g.add_node((1., 1.));
     let d = g.add_node((0., 2.));
-    let e = g.add_node((3., 3.));
+    let e = g.add_node((3., 2.));
     let f = g.add_node((4., 2.));
     let _ = g.add_node((5., 5.)); // no path to node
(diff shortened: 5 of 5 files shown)


src/algo/mod.rs
Dietr1ch, line 1: I figured that adding `Sub<M, Output = M>` would be ok, but I guess I can move the extra requirement into A*.

Adding this was needed to recover `g` as `f-g`

RaoulLuque:
First off, thanks for taking the time to create a PR 🦕

Okay, well changing the Measure trait / the trait bounds on A* would be a breaking change and thus move the adoption of this change to earliest 0.9. With 0.9, the trait system might have changed drastically, so one would have to see how this change might fit in.

Would there possibly be a way to express the subtraction with the current definition of Measure (i.e. with addition)? I am thinking about something like keeping track of all three relevant values f,g,h, or only g and h somehow, since we can construct f from both of them?

Also the Partial Ord trait bound on K seems redundant since Measure is a subtrait of PartialOrd already.

Additionally, we usually call the benchmark files according to the algorithms. I am kind of surprised there is no file for A* yet, but please rename the file accordingly :) Moreover, since there are not other benchmarks yet, I would really like 1/2 more benchmarks which might not be as biased towards the optimization. Feel free to look at something like Dijkstra benchmarks / the build_graph function in benches/common/factories.rs to create some sufficiently large graphs (of course you'd have to check which nodes are connected and sufficiently far away from each other).

At last, please rename your PR to fit the conventional commit/PR format.

src/algo/mod.rs
RaoulLuque: Moving it to A* will also be a breaking change

starovoid:
Since the algorithm has a fairly large number of parameters, it already makes sense to switch to instance implementation. There have already been such proposals, for example #359. It won't help to get rid of the breaking changes, but we could collect all the suggestions about A* and prepare them to be combined for the new major release.

RaoulLuque:
> Since the algorithm has a fairly large number of parameters, it already makes sense to switch to instance implementation. There have already been such proposals, for example #359. It won't help to get rid of the breaking changes, but we could collect all the suggestions about A* and prepare them to be combined for the new major release.

Good point, I haven't thought about combining the changes with previous efforts 👍
Are you sure you mean major release? Since to me that would mean 0.9 and I think we could incorporate some optimizations before this release already.
Also, what do you mean by "instance implementation"?

starovoid:
> Also, what do you mean by "instance implementation"?

I mean the approach to describing an algorithm through a structure. For example, in #359:

```rust
pub struct AstarInstance<G, F, H, K, IsGoal>
where
    G: IntoEdges + Visitable,
    IsGoal: FnMut(G::NodeId) -> bool,
    G::NodeId: Eq + Hash,
    F: FnMut(G::EdgeRef) -> K,
    H: FnMut(G::NodeId) -> K,
    K: Measure + Copy,
{
    graph: G,
    is_goal: IsGoal,
    edge_cost: F,
    estimate_cost: H,

    estimate_costs: HashMap<G::NodeId, K>,
}
```

>  I think we could incorporate some optimizations before this release already.

All non-breaking optimizations can indeed be included in the upcoming minor releases, but so far it seems to me that the most "interesting" thing is still breaking. I will be glad to make A* roadmap after the release of `0.8.2` and think about the sequence of changes in more detail.

RaoulLuque:
Ah, that makes sense. Thanks for the explanation 🙏

That kind of roadmap sounds great :)

I actually think that this optimization could be implemented without changing the API, no? As I stated above, instead of using subtraction, one could use addition. There would be some overhead, but I'd think that it is negligible in comparison to the speedup :) If I have time, I'll try to work on an example implementing that change, including some benchmarks :sun_with_face: 

In any case, I agree that we could look at actually incorporating such optimizations after 0.8.2 :+1:

src/algo/mod.rs
Dietr1ch: It's a smaller breaking change though. I moved it to scope down the breakage, but I feel that a Measure should be `Sub<Self, Output = Self>`.

I'm not sure if these breakages are meaningfully different for the library though. It might not be worth making the distinction and having the intermediate breakage.

RaoulLuque: In the comment right below this I already stated on how a breaking change might affect the merger of this PR. As also stated there, I would suggest trying to solve the problem without an API change (that is using addition instead of Sub). Unfortunately, the "size" of the breaking change does not matter, as any breaking change counts as breaking and will thus be delayed since a major release would be necessary.

The situation currently is, that probably 0.9 there are gonna be a lot of changes to the current internal implementation of things ( in particular to a lot of traits ) and thus, I am not sure whether this change will even be applicable anymore. Hence, I would either recommend to try to implement it now, or pause the PR until then and see how it fits into the possibly new design with 0.9 :)

Dietr1ch:
The implementation needs to rank with either `(f, h)` (using `Sub` to recover `g`) or `(f, -g)` (which requires `Neg`).

Alternatively, you could rank with `(f, h, g)` to keep all 3 values in the heap, which seems a bit silly, but would work without adding new requirements to `K` (nor `Measure`).

---

Oh, BTW having a structure representing a run of A* allows to implement `Iterator` and offer stats to peer into the algorithm's execution internals.

My own experimental implementation[^0] looks like this,

```rust
struct AStarSearch {..}

impl AStarSearch {
  pub fn new(..) -> Self {..}
  pub fn find_next_goal(&mut self) -> Option<Path> {..}
  
  #[cfg(feature = "stats")]
  pub fn stats() -> Stats {..}
}

impl Iterator for AStarSearch {
    type Item = Path;
        self.find_next_goal()
    }
}
```

src/algo/mod.rs
Dietr1ch: And where is this `0.9` version being drafted? I see the `next` branch is stale and there's no [other branch](https://github.com/petgraph/petgraph/branches) that stands out.

There's a hack to implement this without `Sub` nor `Neg` a the expense of the heap entries being larger that should work. Should I just go with that and leave a note about the possible data trim?

Dietr1ch:
I implemented the non-breaking workaround since 0.9 doesn't have a release branch that can take the breaking change.

RaoulLuque:
Nice!

Would you please incorporate the changes I requested in the comments above. That is, add more benchmarks than only the maze one. Also, please provide before/after benchmark result comparisons. In particular, a graph which is just a line would be interesting, since in that case the new version should loose the most performance with the unnecessary entries in the binary tree.

Once that is done and the benchmark results look good, I'll review the changes once I've got time :)

src/algo/mod.rs
RaoulLuque: Yes, I would second the solution you suggested :) At least if the benchmarks show that it is actually worth it ! Otherwise we'll have to just pause this PR and see until 0.9 is around the corner.

Regarding the branches: The next branch is supposed to be the reworked one. As you correctly pointed out, currently it is not being worked on and is thus behind the master branch because the main author/maintainer took a pause and said they'd resume work in the course of this year :)

Dietr1ch: It's definitely worth it, but yes, you shouldn't take my word for granted. It'd be nice to have more benchmarks available. I'll try to come up with something different, yet simple, but I'm past my detouring quota already, so it might take me some time.

Dietr1ch:
> a graph which is just a line would be interesting, since in that case the new version should lose the most performance with the unnecessary entries in the binary tree.

Well, given the current implementation uses 2 hashmaps it should still lose to the new implementation as the overhead around the keys is larger than having a larger value.

Still interesting though, because that kind of graph stresses some of the internals the most, but the real optimisation comes from being significantly better at exploring the graph and not just faster at picking the next node. From my time doing research around this, measuring the number of nodes expanded was the best measure as time ends up being linear in them, but noisy and dependent on things that were harder to reproduce (machine (cpu, cache, mem and their timings), load, compiler, linker) and compare.

Dietr1ch:
Thinking about breaking APIs, is there a plan to break `Graph` in `0.9`?

I'm asking because the other classic problems that are also widely used for benchmarking might require an abstract Graph representation to avoid materialising the entire state space.[0] 

The classic [15-puzzle](https://en.wikipedia.org/wiki/15_puzzle) shouldn't need building `16!/2` nodes and their edges. Instead keeping a `[u8; 16]` around and operating on it to generate nodes on the fly is more efficient.


[0]: It seems that [`Graph`](https://github.com/petgraph/petgraph/blob/2f8b190caa24b741cc30e98b5b0c4ae40b4b2133/src/graph_impl/mod.rs#L391-L395) instantiates the whole graph. I can see how that's convenient for some graphs, but this struct should be behind some abstract trait. (Maybe I missed something when skimming around?)

RaoulLuque:
I am not sure this is the right place for such a discussion. Apart from that, I think that I don't quite understand what you mean exactly. If you would like to go into detail about this or make a suggestion please consider creating something like a conversation in the discussion tab 👍

I am glad to review your PR tho once you incorporate the changes I asked for above. Please consider turning this into a draft PR until then :)

Dietr1ch:
I moved the discussion about `Graph` to a proper discussion on [#821](https://github.com/petgraph/petgraph/discussions/821).

I think the PR is in a state that can me merged, so I'm not making it a draft. If you feel strongly about adding more problems and benchmarks I think `petgraph` should make `Graph` more general before getting those.

RaoulLuque:
Okay. In that case someone else will have to give their opinion I guess 👍 Maybe you want to chime in @starovoid ?

My point still stands that the benchmarks to compare the performance of the new and old variant should cover different cases and not only the one where we know that it will favor the new implementation 👍 If the benchmarks show improvements in every case I would be glad to review the changes :)

I also don't understand the argument of changing the graph type before adding more benchmarks, since if we implement this PR, it will mostly affect people using the graph type as it currently is and the new Astar (in particular most people probably don't implement their own graph types for virtualized/unmaterialized graphs, but instead use the types provided by petgraph). In that case it would be interesting to see exactly that combination, of how the current graph type performs in different scenarios with the modifications of Astar 👍

Dietr1ch:
> I also don't understand the argument of changing the graph type before adding more benchmarks

I said this thinking about adding more variety to the graphs. I can do the line graph, or a small road network, but to me the interesting graphs are the ones that are large enough to make the decision-making while exploring the graph more important than details like how quick is the hash table.

Representing the classical 16-puzzle I mentioned would take about 50TB-300TB of memory depending on how "clever" (naughty) you get. It's completely unreasonable to try doing it without "abstract" graph support.

Dietr1ch:
Got benchmarks on the "line" graph (actually an aisle since it leads to slightly more ranking work in the heap).

There's a 2x speedup on it, even though my change should've made things slightly slower, but this is just due to removing the ~redundant hashmap as I made the ranking adjustment.

```text
 name                     dev__pathfinding_bench ns/iter  dev__astar_rank_tie_breaking ns/iter  diff ns/iter   diff %  speedup
 astar_maze2d_bench       3,873,653                       208,091                                 -3,665,562  -94.63%  x 18.62
 astar_maze2d_line_bench  161,513                         63,586                                     -97,927  -60.63%   x 2.54
 ```



---

BTW, I also have an environment commit that adds a simple `flake.nix` file and a simple bash script to run a benchmark comparing 2..N branches. I can contribute that too, but I'll fork that into a new PR.

File: benches/astar.rs [hunk 1]
```diff
@@ -0,0 +1,103 @@
...
+
+use petgraph::prelude::*;
+use test::Bencher;
+
+mod common;
+use petgraph::algo::astar;
+
+#[bench]
+fn astar_maze2d_line_bench(bench: &mut Bencher) {
+    // Same maze as ./dijkstra.rs
```
RaoulLuque:
```suggestion
    // Same maze as ./dijkstra.rs
```
This might have been copied by accident. Either remove this or add the same benchmark to dijkstra as well :)

benches/common/factories.rs
```diff
@@ -355,3 +355,24 @@ pub fn build_graph(node_count: usize, dense: bool) -> Graph<usize, i32, Undirect
     }
     graph
 }
+
+pub fn build_2d_maze(width: usize, height: usize) -> Graph<(usize, usize), usize, Undirected> {
```
RaoulLuque:
```suggestion

/// Builds a 2d grid graph with the provided width and height
pub fn build_2d_grid(width: usize, height: usize) -> Graph<(usize, usize), usize, Undirected> {
```
According to [Wikipedia](https://en.wikipedia.org/wiki/Lattice_graph) graphs like this are rather referred to as lattice or grid graphs. Since lattice has some other meanings as well in Computer Science, I would suggest grid :)

benches/astar.rs
```diff
@@ -0,0 +1,103 @@
...
+            heuristic_for(goal),
+        );
+    });
+}
+
+#[bench]
+fn astar_maze2d_bench(bench: &mut Bencher) {
+    // Same maze as ./dijkstra.rs
+    const N: usize = 500;
+    let g = common::build_2d_maze(N, N);
```
RaoulLuque:
```suggestion
    let g = common::build_2d_grid(N, N);
```
See comment on method itself

benches/astar.rs
```diff
@@ -0,0 +1,103 @@
...
+
+mod common;
+use petgraph::algo::astar;
+
+#[bench]
+fn astar_maze2d_line_bench(bench: &mut Bencher) {
+    // Same maze as ./dijkstra.rs
+    const N: usize = 500;
+    const M: usize = 5;
+    let g = common::build_2d_maze(N, M);
```
RaoulLuque:
```suggestion
    let g = common::build_2d_grid(N, M);
```
See comment on method itself

benches/dijkstra.rs
```diff
@@ -31,3 +32,22 @@ fn dijkstra_bench(bench: &mut Bencher) {
         let _scores = dijkstra(&g, nodes[0], None, |e| *e.weight());
     });
 }
+
+#[bench]
+fn dijkstra_maze2d_bench(bench: &mut Bencher) {
+    // Same maze as ./astar.rs
```
RaoulLuque:
```suggestion
fn dijkstra_grid2d_bench(bench: &mut Bencher) {
    // Same grid as in ./astar.rs
```
See comment on grid/maze method

benches/dijkstra.rs
```diff
@@ -31,3 +32,22 @@ fn dijkstra_bench(bench: &mut Bencher) {
...
+    const N: usize = 500;
+    let g = common::build_2d_maze(N, N);
```
RaoulLuque:
```suggestion
    let g = common::build_2d_grid(N, N);
```
See comment on grid method

benches/astar.rs
```diff
@@ -0,0 +1,103 @@
...
+            |n| n == goal,
+            |e| *e.weight(),
+            heuristic_for(goal),
+        );
+    });
+}
+
+#[bench]
+fn astar_maze2d_bench(bench: &mut Bencher) {
+    // Same maze as ./dijkstra.rs
```
RaoulLuque:
```suggestion
fn astar_grid2d_bench(bench: &mut Bencher) {
    // Same grid as in ./dijkstra.rs
```
See comment on method itself

benches/astar.rs
```diff
@@ -0,0 +1,103 @@
...
+extern crate test;
+
+use petgraph::prelude::*;
+use test::Bencher;
+
+mod common;
+use petgraph::algo::astar;
+
+#[bench]
+fn astar_maze2d_line_bench(bench: &mut Bencher) {
```
RaoulLuque:
```suggestion
fn astar_grid2d_line_bench(bench: &mut Bencher) {
```
See comment on grid/maze factory method itself

src/algo/astar.rs
```diff
@@ -91,61 +91,68 @@ where
     H: FnMut(G::NodeId) -> K,
     K: Measure + Copy,
 {
+    // The Open set
     let mut visit_next = BinaryHeap::new();
-    let mut scores = HashMap::new(); // g-values, cost to reach the node
-    let mut estimate_scores = HashMap::new(); // f-values, cost to reach + estimate cost to goal
+    // A node -> (f, h, g) mapping. Should be (f, h) since g can be recovered.
```
RaoulLuque:
```suggestion
    // A node -> (f, h, g) mapping. TODO: With a breaking change this should be (f, h) since g can be recovered if subtraction is available.
```
Seems clearer

tests/graph.rs
```diff
@@ -501,7 +501,7 @@ fn test_astar_manhattan_heuristic() {
     let b = g.add_node((2., 0.));
     let c = g.add_node((1., 1.));
     let d = g.add_node((0., 2.));
-    let e = g.add_node((3., 3.));
+    let e = g.add_node((3., 2.));
```
RaoulLuque, line 504: Why did this need to be changed? It seems to fail with the old value.

RaoulLuque (commented):
The performance improvement is indeed very impressive !

I had some time to look through the files :) Since we don't have any scripts of sorts in petgraph, please remove the bench script you also added.

tests/graph.rs
Dietr1ch: It fails because the heuristic provided here isn't admissible, which misleads the search into finding a sub-optimal path (and thus the assertion later on having optimal cost fails).

benches/astar.rs
Dietr1ch: Renamed all instances preferring grid over maze.

benches/astar.rs
Dietr1ch: Renamed all instances preferring grid over maze.

benches/dijkstra.rs
Dietr1ch: Renamed all instances preferring grid over maze.

benches/dijkstra.rs
Dietr1ch: Renamed all instances preferring grid over maze.

benches/astar.rs
Dietr1ch: Renamed all instances preferring grid over maze.

benches/common/factories.rs
Dietr1ch: Fair, it's a name that I just grabbed from work with real mazes, but since I'm using empty ones here it seems kind of misleading too.

It seems that my understanding of lattices being tied to power set inclusions was not broad enough.

benches/astar.rs
Dietr1ch: Renamed all instances preferring grid over maze.

benches/astar.rs
Dietr1ch: Not copied by accident. I'm just keeping bi-directional file links to show that the same problem is being used to bench both, A* and Dijkstra. Maybe I forgot to grab the change into `bench/dijkstra.rs` into this commit.

Dietr1ch:
Oh, I screwed up by switching back and forth between laptop/workstation believing that my github remote was updated and ended up resolving the comments over an older commit that didn't do the ranking trick that avoids requiring `Sub` and breaking the API. I'll re-do that now.

Dietr1ch:
I re-ran the bench on my workstation using the script I had around.

Same remarks apply.
- The ranking speed-up numbers are kind of made up. Empty grids essentially compare `dim^N` vs `dim`, so if you want to see >1000x out of it, just crank up the problem size, but to keep things comparable we should settle in the current instance.
- The ~2x speed-up in the 1-D version comes from using a single hashmap instead of two.
- This is not super scientific as sampling-based benchmarks are naturally noisy and get worse when execution isn't parallel, but I've seen ~consistent 18x-22x on the 2D grid bench and 1.7x-2.5x on the "1D" grid on both, my laptop and workstation.


| name                      | [dev/pathfinding_bench](https://github.com/petgraph/petgraph/pull/813/commits/96cdd011025d79644591b7e69c83c927e9771c42) ns/iter  | [dev/astar_rank_tie_breaking](https://github.com/petgraph/petgraph/pull/813/commits/2b866b30d3638c944981d9f9b3143bd00a4acab6) ns/iter | diff ns/iter   | diff %  | speedup |
|---------------------------|--------------------------------|-------------------------------------|---------------|---------|---------|
| `astar_2d_grid_bench`       | 3,872,858                       | 210,172 | -3,662,686  | -94.57%  | x18.43 |
| `astar_2d_grid_line_bench`  | 164,219                         | 67,211  | -97,008  | -59.07%   | x2.44 |

benches/astar.rs
RaoulLuque: Ah, I see :)

RaoulLuque:
Alright, well if nothing else changed, LGTM :) :+1:

Thanks for taking the suggestions into consideration and all the work you put in so far :sauropod: 

Gonna ping @starovoid , so he can say how to proceed :relaxed:

RaoulLuque:
Well, actually the only other thing I think we could do while we are at it, is add a quickcheck for `astar`. It could be similar to the one for dijkstra, checking whether computed paths satisfy the triangle inequality using something like the null heuristic?

It would go like right below the `fn dijkstra_triangle_ineq` in `tests/quickcheck.rs`

Dietr1ch:
> checking whether computed paths satisfy the triangle inequality using something like the null heuristic?

There's 2 strong arguments IMO against doing this.

- First, it seems a lot like testing the test, so what's stopping us from unproductively recursing into testing^+ the test indefinitely? At some point you need to simply be able to stare at the tests long enough to convince yourself things are fine, and in this case I think we try hard enough by comparing against Dijkstra's 1->N optimal paths.

- Second, this is something that when studying A* is explicitly warned against trying to do, since doing that is more expensive than running the search, so it'd only make sense if you know your problem in advance, but can't afford the memory to just run Floyd-Warshall once.


That said, there's a weaker version of this, which is testing the heuristic along the optimal path retrieved, which comes almost for free since you already went through it. Along this path `h` should't over-estimate `c` (`h(a, b) <= c(a, b)`). This is something that would expose the heuristic is inconsistent, but not enough to prove the path isn't optimal in case it's inconsistent, and neither enough to prove the heuristic is indeed consistent.

Proofs around heuristics often come from the ideas that allow you to have an heuristic in the first place, so I think are rarely needed once you see your heuristic is just an optimal solution in a relaxed version of your problem.

If you really want to add this weaker check, I suggest you just add some `debug_assert`ions, but keep that away from the tests and runtime on release runs.

RaoulLuque:
Thanks for the response and the insights. I am not sure however, whether we are on the same page here. I am not suggesting to add another unit test. Instead I am suggesting to add a [quickcheck](https://github.com/BurntSushi/quickcheck) like the ones in `tests/quickcheck.rs`.

This would allow the algorithm (A*) to be tested on a multitude of random graphs instead of just the one from the unit test. I do see the point of possibly missing bugs when comparing with the triangle inequality since that should be given with A*, however it could still catch some bugs. However, we could also just compare it with Dijkstra in a Quickcheck setting as well. That would rely on Dijkstra working of course, but that seems like a reasonable assumption I think.

Maybe I also misunderstand you but it sounds like you thought I'd suggest adding another unit test. Instead I suggested adding a quickcheck which has the great benefit of systematically testing pseudo-random graphs, which make testing edge-case graphs way easier.

RaoulLuque:
Hey, just to let you know. I would fork and pickup the PR myself in a few days and just implement the last suggestions regarding quickcheck to merge your work, since I think that it is very valuable :)

RaoulLuque:
Closed in favor of #882

PR #815, opened 2025-06-04
docs: Unify algo docs

starovoid:
This PR aims to bring uniformity to the documentation of algorithms.

Now the doc comments will have the following structure:

```markdown
Main description...
...
...

# Arguments
...

# Returns
...

# Complexity
* Time complexity: ...
* Space complexity: ...

# Examples

```

For example, the `spfa` doc comment might look like:
```markdown
/// \[Generic\] Compute shortest paths from node `source` to all other.
///
/// Using the [Shortest Path Faster Algorithm][spfa].
/// Compute shortest distances from node `source` to all other.
///
/// Compute shortest paths lengths in a weighted graph with positive or negative edge weights,
/// but with no negative cycles.
///
/// ## Arguments
/// * `graph`: weighted graph.
/// * `source`: the source vertex, for which we calculate the lengths of the shortest paths to all the others.
/// * `edge_cost`: closure that returns cost of a particular edge
///
/// ## Returns
/// * `Err`: if graph contains negative cycle.
/// * `Ok`: a pair of a vector of shortest distances and a vector
///   of predecessors of each vertex along the shortest path.
///
/// ## Complexity
/// * Time complexity: **O(|V||E|)**, but it's generally assumed that in the average case it is **O(|E|)**
/// * Space complexity: **O(|V|)**
///
/// |V| is the number of nodes, |E| is the number of edges in the input graph.
///
///
/// [spfa]: https://www.geeksforgeeks.org/shortest-path-faster-algorithm/
///
/// # Example
```

### DIFF
--- a/src/algo/astar.rs
+++ b/src/algo/astar.rs
@@ -25,9 +25,6 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// it should never overestimate the actual cost to get to the nearest goal node. Estimate costs
 /// must also be non-negative.
 ///
-/// The graph should be `Visitable` and implement `IntoEdges`.
-///
-///
 /// # Arguments
 /// * `graph`: weighted graph.
 /// * `start`: the start node.
@@ -40,9 +37,9 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// * `None` - if such a path was not found.
 ///
 /// # Complexity
-/// * Time complexity: **O(b^d)**, where **b** is the branching factor (the average number of successors per state)
-///   and **d** is the depth of *goal* node.
+/// The time complexity largely depends on the heuristic used. You can contribute to improve the documentation in this place ;)
 ///
+/// With a trivial heuristic, the algorithm will work like [`fn@crate::algo::dijkstra`], but other situations are possible too.
 ///
 /// # Example
 /// ```
--- a/src/algo/astar.rs
+++ b/src/algo/astar.rs
@@ -25,7 +25,27 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// it should never overestimate the actual cost to get to the nearest goal node. Estimate costs
 /// must also be non-negative.
 ///
+/// Returns the total cost + the path of subsequent `NodeId` from start to finish, if one was
+/// found.
+///
 /// The graph should be `Visitable` and implement `IntoEdges`.
+///
+///
+/// # Arguments
+/// * `graph`: weighted graph
+/// * `start`: the start node
+/// * `is_goal`: the callback defines the goal node
+/// * `edge_cost`: closure that returns cost of a particular edge
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node
+///
+/// # Returns
+/// * `Some(...)` - the total cost and path from start to finish, if one was found.
+/// * `None` - if such path was not found.
+///
+/// # Complexity
+/// * Time complexity: **O(b^d)**, where **b** is branching factor (the average number of successors per state)
+///   and **d** is the depth of *goal* node.
+///
 ///
 /// # Example
 /// ```
@@ -62,9 +82,6 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// let path = astar(&g, a, |finish| finish == f, |e| *e.weight(), |_| 0);
 /// assert_eq!(path, Some((6, vec![a, d, e, f])));
 /// ```
-///
-/// Returns the total cost + the path of subsequent `NodeId` from start to finish, if one was
-/// found.
 pub fn astar<G, F, H, K, IsGoal>(
     graph: G,
     start: G::NodeId,
--- a/src/algo/astar.rs
+++ b/src/algo/astar.rs
@@ -37,9 +37,9 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// * `None` - if such a path was not found.
 ///
 /// # Complexity
-/// The time complexity largely depends on the heuristic used. You can contribute to improve the documentation in this place ;)
+/// The time complexity largely depends on the heuristic used. Feel free to contribute and provide the exact time complexity :)
 ///
-/// With a trivial heuristic, the algorithm will work like [`fn@crate::algo::dijkstra`], but other situations are possible too.
+/// With a trivial heuristic, the algorithm will behave like [`fn@crate::algo::dijkstra`].
 ///
 /// # Example
 /// ```
--- a/src/algo/astar.rs
+++ b/src/algo/astar.rs
@@ -25,25 +25,22 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// it should never overestimate the actual cost to get to the nearest goal node. Estimate costs
 /// must also be non-negative.
 ///
-/// Returns the total cost + the path of subsequent `NodeId` from start to finish, if one was
-/// found.
-///
 /// The graph should be `Visitable` and implement `IntoEdges`.
 ///
 ///
 /// # Arguments
-/// * `graph`: weighted graph
-/// * `start`: the start node
-/// * `is_goal`: the callback defines the goal node
-/// * `edge_cost`: closure that returns cost of a particular edge
-/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node
+/// * `graph`: weighted graph.
+/// * `start`: the start node.
+/// * `is_goal`: the callback defines the goal node.
+/// * `edge_cost`: closure that returns cost of a particular edge.
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node.
 ///
 /// # Returns
-/// * `Some(...)` - the total cost and path from start to finish, if one was found.
-/// * `None` - if such path was not found.
+/// * `Some(K, Vec<G::NodeId>)` - the total cost and path from start to finish, if one was found.
+/// * `None` - if such a path was not found.
 ///
 /// # Complexity
-/// * Time complexity: **O(b^d)**, where **b** is branching factor (the average number of successors per state)
+/// * Time complexity: **O(b^d)**, where **b** is the branching factor (the average number of successors per state)
 ///   and **d** is the depth of *goal* node.
 ///
 ///
--- a/src/algo/bellman_ford.rs
+++ b/src/algo/bellman_ford.rs
@@ -31,7 +31,7 @@ pub struct Paths<NodeId, EdgeWeight> {
 ///
 /// # Complexity
 /// * Time complexity: **O(|V||E|)**.
-/// * Space complexity: **O(|V|)**.
+/// * Auxiliary space: **O(|V|)**.
 ///
 /// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
@@ -138,7 +138,7 @@ where
 ///
 /// # Complexity
 /// * Time complexity: **O(|V||E|)**.
-/// * Space complexity: **O(|V|)**.
+/// * Auxiliary space: **O(|V|)**.
 ///
 /// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
--- a/src/algo/bellman_ford.rs
+++ b/src/algo/bellman_ford.rs
@@ -24,6 +24,21 @@ pub struct Paths<NodeId, EdgeWeight> {
 /// out the predecessor of a node along a shortest path. The vectors
 /// are indexed by the graph's node indices.
 ///
+/// # Arguments
+/// * `g`: graph with no negative cycle
+/// * `source`: the source node
+///
+/// # Returns
+/// * `Ok`: (if graph contains no negative cycle) a struct [`Paths`] containing distances and
+///   predecessors along each shortest path.
+/// * `Err`: if graph contains negative cycle.
+///
+/// # Complexity
+/// * Time complexity: **O(|V||E|)**
+/// * Space complexity: **O(|V|)**
+///
+/// where **|V|** is the number of nodes and **|E|** is the number of edges.
+///
 /// [bf]: https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm
 ///
 /// # Example
@@ -117,7 +132,20 @@ where
 ///
 /// If a negative cycle is found from source, return one vec with a path of `NodeId`s.
 ///
-/// The time complexity of this algorithm should be the same as the Bellman-Ford (O(|V|·|E|)).
+/// # Arguments
+/// * `g`: graph
+/// * `source`: the source node
+///
+/// # Returns
+/// * `Some(Vec<G::NodeId>)` - the path of the negative cycle (if found)
+/// * `None` - if `g` doesn't contain negative cycles reachable from `source`
+///
+/// # Complexity
+/// * Time complexity: **O(|V||E|)**
+/// * Space complexity: **O(|V|)**
+///
+/// where **|V|** is the number of nodes and **|E|** is the number of edges.
+///
 ///
 /// [nc]: https://blogs.asarkar.com/assets/docs/algorithms-curated/Negative-Weight%20Cycle%20Algorithms%20-%20Huang.pdf
 /// [bf]: https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm
--- a/src/algo/bellman_ford.rs
+++ b/src/algo/bellman_ford.rs
@@ -20,22 +20,18 @@ pub struct Paths<NodeId, EdgeWeight> {
 /// permitted, but the graph must not have a cycle of negative weights
 /// (in that case it will return an error).
 ///
-/// On success, return one vec with path costs, and another one which points
-/// out the predecessor of a node along a shortest path. The vectors
-/// are indexed by the graph's node indices.
-///
 /// # Arguments
-/// * `g`: graph with no negative cycle
-/// * `source`: the source node
+/// * `g`: graph with no negative cycle.
+/// * `source`: the source node.
 ///
 /// # Returns
 /// * `Ok`: (if graph contains no negative cycle) a struct [`Paths`] containing distances and
-///   predecessors along each shortest path.
+///   predecessors along each shortest path. The vectors in [`Paths`] are indexed by the graph's node indices.
 /// * `Err`: if graph contains negative cycle.
 ///
 /// # Complexity
-/// * Time complexity: **O(|V||E|)**
-/// * Space complexity: **O(|V|)**
+/// * Time complexity: **O(|V||E|)**.
+/// * Space complexity: **O(|V|)**.
 ///
 /// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
@@ -127,22 +123,22 @@ where
 
 /// \[Generic\] Find the path of a negative cycle reachable from node `source`.
 ///
-/// Using the [find_negative_cycle][nc]; will search the Graph for negative cycles using
+/// Using the [find_negative_cycle][nc]; will search the graph for negative cycles using
 /// [Bellman–Ford algorithm][bf]. If no negative cycle is found the function will return `None`.
 ///
 /// If a negative cycle is found from source, return one vec with a path of `NodeId`s.
 ///
 /// # Arguments
-/// * `g`: graph
-/// * `source`: the source node
+/// * `g`: graph.
+/// * `source`: the source node.
 ///
 /// # Returns
-/// * `Some(Vec<G::NodeId>)` - the path of the negative cycle (if found)
-/// * `None` - if `g` doesn't contain negative cycles reachable from `source`
+/// * `Some(Vec<G::NodeId>)` - the path of the negative cycle (if found).
+/// * `None` - if `g` doesn't contain negative cycles reachable from `source`.
 ///
 /// # Complexity
-/// * Time complexity: **O(|V||E|)**
-/// * Space complexity: **O(|V|)**
+/// * Time complexity: **O(|V||E|)**.
+/// * Space complexity: **O(|V|)**.
 ///
 /// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
--- a/src/algo/bridges.rs
+++ b/src/algo/bridges.rs
@@ -5,11 +5,23 @@ use crate::visit::{
 use alloc::{vec, vec::Vec};
 
 /// Find all [bridges](https://en.wikipedia.org/wiki/Bridge_(graph_theory)) in a simple undirected graph.
-/// The algorithm is **O(|E|)** where **|E|** is the number of edges in the graph.
 ///
-/// Returns the vector of pairs `(G::NodeID, G:: NodeID)`,
+/// Returns the iterator of edge references `G::EdgeRef`,
 /// representing the edges of the input graph that are bridges.
-/// The order of the vertices in the pair and the order of the edges themselves are arbitrary.
+/// The order of the edges is arbitrary.
+///
+/// # Arguments
+/// * `graph`: a simple undirected graph
+///
+/// # Returns
+/// * `impl Iterator`:  the iterator of edge references `G::EdgeRef`
+///
+/// # Complexity
+/// * Time complexity: **O(|V| + |E|)**
+/// * Space complexity: **O(|V| + |E|)**
+///
+/// where **|V|** is the number of nodes and **|E|** is the number of edges.
+///
 ///
 /// # Examples
 ///
--- a/src/algo/bridges.rs
+++ b/src/algo/bridges.rs
@@ -15,7 +15,7 @@ use alloc::{vec, vec::Vec};
 ///
 /// # Complexity
 /// * Time complexity: **O(|V| + |E|)**.
-/// * Space complexity: **O(|V|)**.
+/// * Auxiliary space: **O(|V|)**.
 ///
 /// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
--- a/src/algo/bridges.rs
+++ b/src/algo/bridges.rs
@@ -6,19 +6,16 @@ use alloc::{vec, vec::Vec};
 
 /// Find all [bridges](https://en.wikipedia.org/wiki/Bridge_(graph_theory)) in a simple undirected graph.
 ///
-/// Returns the iterator of edge references `G::EdgeRef`,
-/// representing the edges of the input graph that are bridges.
-/// The order of the edges is arbitrary.
-///
 /// # Arguments
-/// * `graph`: a simple undirected graph
+/// * `graph`: a simple undirected graph.
 ///
 /// # Returns
-/// * `impl Iterator`:  the iterator of edge references `G::EdgeRef`
+/// * `impl Iterator`:  the iterator of edge references `G::EdgeRef` representing the edges
+///   of the input graph that are bridges. The order of the edges is arbitrary.
 ///
 /// # Complexity
-/// * Time complexity: **O(|V| + |E|)**
-/// * Space complexity: **O(|V| + |E|)**
+/// * Time complexity: **O(|V| + |E|)**.
+/// * Space complexity: **O(|V|)**.
 ///
 /// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
--- a/src/algo/coloring.rs
+++ b/src/algo/coloring.rs
@@ -13,9 +13,21 @@ use crate::visit::{IntoEdges, IntoNodeIdentifiers, NodeIndexable, VisitMap, Visi
 ///
 /// The graph must be undirected. It should not contain loops.
 /// It must implement `IntoEdges`, `IntoNodeIdentifiers` and `Visitable`
-/// Returns a tuple composed of a HashMap that associates to each `NodeId` its color and the number of used colors.
+/// Returns a tuple composed of a `HashMap` that associates to each `NodeId` its color and the number of used colors.
 ///
-/// Computes in **O((|V| + |E|)log(|V|)** time.
+///
+/// # Arguments
+/// * `graph`: undirected graph without loops
+///
+/// # Returns
+/// * `(HashMap, usize)`: [`struct@hashbrown::HashMap`] that associates to each `NodeId` its color
+///   and the `usize` number of used colors.
+///
+/// # Complexity
+/// * Time complexity: **O((|V| + |E|)log(|V|)**
+/// * Space complexity: **O(|V|)**
+///
+/// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
 /// [1]: https://en.wikipedia.org/wiki/DSatur
 ///
--- a/src/algo/coloring.rs
+++ b/src/algo/coloring.rs
@@ -10,22 +10,18 @@ use crate::visit::{IntoEdges, IntoNodeIdentifiers, NodeIndexable, VisitMap, Visi
 ///
 ///
 /// This is a heuristic. So, it does not necessarily return a minimum coloring.
-///
 /// The graph must be undirected. It should not contain loops.
-/// It must implement `IntoEdges`, `IntoNodeIdentifiers` and `Visitable`
-/// Returns a tuple composed of a `HashMap` that associates to each `NodeId` its color and the number of used colors.
-///
 ///
 /// # Arguments
-/// * `graph`: undirected graph without loops
+/// * `graph`: undirected graph without loops.
 ///
 /// # Returns
 /// * `(HashMap, usize)`: [`struct@hashbrown::HashMap`] that associates to each `NodeId` its color
 ///   and the `usize` number of used colors.
 ///
 /// # Complexity
-/// * Time complexity: **O((|V| + |E|)log(|V|)**
-/// * Space complexity: **O(|V|)**
+/// * Time complexity: **O((|V| + |E|)log(|V|)**.
+/// * Space complexity: **O(|V|)**.
 ///
 /// where **|V|** is the number of nodes and **|E|** is the number of edges.
 ///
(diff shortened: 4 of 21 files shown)


RaoulLuque:
Wow, that's a great idea !

I also like the layout of the doc comment you are suggesting :)

While we are at it, I always wondered what the `\[Generic\]` is supposed to be mean?

Also, maybe in the main description at the top, there could be a bit less repetition, by adjusting the following sentence like so:
From
```
/// Compute shortest paths lengths in a weighted graph with positive or negative edge weights,
/// but with no negative cycles.
```
To
```
/// Works on weighted graphs with positive or negative edge weights, but with no negative cycles.
```

That is, not state again what the algorithm does, and instead only state the requirements for the provided graph.
Maybe we could try to keep the requirements on the graph consistent somehow. I am not sure if it makes sense to make an entire bulletpoint (`#`) out of it, but instead adjusting the structure to be like so for example:

```
Main description...
...
...
Requirements for the provided graph

# Arguments
...
```
I think we could also reuse your layout/structure and add it to a PR template to make sure other contributions also stick to this layout :+1:  

At last, if you'd like, I would really like to review the changes :)

starovoid:
> At last, if you'd like, I would really like to review the changes :)

I'm glad to hear that! Moreover, I need help on several points:
- [x] Find out the space complexity of the `greedy_feedback_arc_set` algorithm
- [x] Make sure that `ford_fulkerson` is actually Edmonds-Karp (or not) and document the time and space complexity
- [ ] ~~Document all estimates regarding isomorphism algorithms~~. In fact, it looks so grand that I'm thinking of leaving it for the next release.

This list will probably be updated in the near future :)

---

> While we are at it, I always wondered what the \[Generic\] is supposed to be mean?

This means that the algorithm accepts an arbitrary graph type (satisfying the trait bounds) as input. I suggest to ignore this historical artifact for now.

---

> Also, maybe in the main description at the top, there could be a bit less repetition, by adjusting the following sentence like so...

Please suggest all such changes through the review mechanism, so it will be easier and faster to accept them ;)

---

> I think we could also reuse your layout/structure and add it to a PR template to make sure other contributions also stick to this layout

As for that, the contributing guide has been in need of updating for a long time, I will raise this issue among the maintainers a little later.

RaoulLuque:
> I'm glad to hear that! Moreover, I need help on several points:
> 
>     * [ ]  Find out the space complexity of the `greedy_feedback_arc_set` algorithm

I'll look into that :+1: 

>     * [ ]  Make sure that `ford_fulkerson` is actually Edmonds-Karp (or not) and document the time and space complexity

I can also look into that :+1: 

>     * [ ]  Document all estimates regarding isomorphism algorithms. In fact, it looks so grand that I'm thinking of leaving it for the next release.

I am not very familiar with this class of algorithms, so I'm not sure I'll have enough time to look into this in the near future.

> This means that the algorithm accepts an arbitrary graph type (satisfying the trait bounds) as input. I suggest to ignore this historical artifact for now.

Ah, I see. Because I find it somewhat confusing since one might think that the algorithm is generic in the sense that it can be used for arbitrary graphs. That is, something like directed/undirected or just arbitrary (cycles / no cycles). Because that is not true for spfa although there is the generic tag. 
I then wonder, if there is a reason to keep the generic tag? Because the rustdocs and compiler should tell users if the algorithm accepts arbitrary graphs or not (and which Traits should be implemented) and adding such a tag might be confusing as I described above.

> Please suggest all such changes through the review mechanism, so it will be easier and faster to accept them ;)

Alright I'll do that :) Should I already review the changes or wait until your PR is not in draft mode anymore :)?

> As for that, the contributing guide has been in need of updating for a long time, I will raise this issue among the maintainers a little later.

Alright :+1:

starovoid:
> Should I already review the changes or wait until your PR is not in draft mode anymore

You can do it right now💪

src/algo/astar.rs
```diff
@@ -25,8 +25,28 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
...
+/// # Arguments
+/// * `graph`: weighted graph
+/// * `start`: the start node
+/// * `is_goal`: the callback defines the goal node
+/// * `edge_cost`: closure that returns cost of a particular edge
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node
+///
+/// # Returns
+/// * `Some(...)` - the total cost and path from start to finish, if one was found.
+/// * `None` - if such path was not found.
```
RaoulLuque:
```suggestion
/// * `None` - if such a path was not found.
```

src/algo/astar.rs
```diff
@@ -25,8 +25,28 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// it should never overestimate the actual cost to get to the nearest goal node. Estimate costs
 /// must also be non-negative.
 ///
+/// Returns the total cost + the path of subsequent `NodeId` from start to finish, if one was
+/// found.
+///
 /// The graph should be `Visitable` and implement `IntoEdges`.
 ///
+///
```
RaoulLuque: The line right above this line "The graph should be `Visitable` and implement `IntoEdges`." should probably be removed, since it is not present in most other graphs and is also documented in the signature of the function.

src/algo/astar.rs
```diff
@@ -25,8 +25,28 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
...
+/// * `is_goal`: the callback defines the goal node
+/// * `edge_cost`: closure that returns cost of a particular edge
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node
+///
+/// # Returns
+/// * `Some(...)` - the total cost and path from start to finish, if one was found.
+/// * `None` - if such path was not found.
+///
+/// # Complexity
+/// * Time complexity: **O(b^d)**, where **b** is branching factor (the average number of successors per state)
```
RaoulLuque:
```suggestion
/// * Time complexity: **O(b^d)**, where **b** is the branching factor (the average number of successors per state)
```

src/algo/astar.rs
```diff
@@ -25,8 +25,28 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
 /// it should never overestimate the actual cost to get to the nearest goal node. Estimate costs
 /// must also be non-negative.
 ///
+/// Returns the total cost + the path of subsequent `NodeId` from start to finish, if one was
+/// found.
+///
```
RaoulLuque:
```suggestion
```
This comment possibly applies to more algs, but maybe we could leave out lines describing the return type in other places than in #Returns ? And if necessary, add more info to the # Returns section?

src/algo/astar.rs
```diff
@@ -25,8 +25,28 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
...
+/// * `edge_cost`: closure that returns cost of a particular edge
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node
+///
+/// # Returns
+/// * `Some(...)` - the total cost and path from start to finish, if one was found.
+/// * `None` - if such path was not found.
+///
+/// # Complexity
+/// * Time complexity: **O(b^d)**, where **b** is branching factor (the average number of successors per state)
+///   and **d** is the depth of *goal* node.
```
RaoulLuque: I am not sure about this time complexity. I would think it is something like `O((|V| + |E|) log |V|)`. In particular the current O(b^d) cannot be true for a graph which is a line for example. Where for arbitrary "length" of the line, the complexity would be constant which is not true of course.

src/algo/bellman_ford.rs
```diff
@@ -24,6 +24,21 @@ pub struct Paths<NodeId, EdgeWeight> {
 /// out the predecessor of a node along a shortest path. The vectors
 /// are indexed by the graph's node indices.
 ///
```
RaoulLuque: Here, the same comment from above applies, about moving this info exclusively into #Returns?

src/algo/astar.rs
```diff
@@ -25,8 +25,28 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
...
+///
+/// # Arguments
+/// * `graph`: weighted graph
+/// * `start`: the start node
+/// * `is_goal`: the callback defines the goal node
+/// * `edge_cost`: closure that returns cost of a particular edge
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node
+///
+/// # Returns
+/// * `Some(...)` - the total cost and path from start to finish, if one was found.
```
RaoulLuque:
```suggestion
/// * `Some(K, Vec<G::NodeId>)` - the total cost and path from start to finish, if one was found.
```

src/algo/astar.rs
```diff
@@ -25,8 +25,28 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
...
+///
 /// The graph should be `Visitable` and implement `IntoEdges`.
 ///
+///
+/// # Arguments
+/// * `graph`: weighted graph
+/// * `start`: the start node
+/// * `is_goal`: the callback defines the goal node
+/// * `edge_cost`: closure that returns cost of a particular edge
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node
```
RaoulLuque:
```suggestion
/// * `graph`: weighted graph.
/// * `start`: the start node.
/// * `is_goal`: the callback defines the goal node.
/// * `edge_cost`: closure that returns cost of a particular edge.
/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node.
```

Not sure about this, but that's how it is in your spfa example. Might apply to other algs aswell.

src/algo/bridges.rs
```diff
@@ -5,11 +5,23 @@ use crate::visit::{
 use alloc::{vec, vec::Vec};
 
 /// Find all [bridges](https://en.wikipedia.org/wiki/Bridge_(graph_theory)) in a simple undirected graph.
-/// The algorithm is **O(|E|)** where **|E|** is the number of edges in the graph.
 ///
-/// Returns the vector of pairs `(G::NodeID, G:: NodeID)`,
+/// Returns the iterator of edge references `G::EdgeRef`,
 /// representing the edges of the input graph that are bridges.
-/// The order of the vertices in the pair and the order of the edges themselves are arbitrary.
+/// The order of the edges is arbitrary.
```
RaoulLuque: Here the same from above applies about describing what is returned at multiple points in the docs. However, the info on the order of the edges is important, so maybe move that into the #Returns section?

src/algo/bridges.rs
```diff
@@ -5,11 +5,23 @@ use crate::visit::{
...
+///
+/// # Arguments
+/// * `graph`: a simple undirected graph
+///
+/// # Returns
+/// * `impl Iterator`:  the iterator of edge references `G::EdgeRef`
+///
+/// # Complexity
+/// * Time complexity: **O(|V| + |E|)**
+/// * Space complexity: **O(|V| + |E|)**
```
RaoulLuque: The space complexity here is interesting. I would say it is O(|V|) in most cases. However, since vecs with size graph.node_bound() are allocated, that might actually be arbitrarily big (if the graph is stable and had 100_000 nodes but now only has 10 for example). But maybe we write O(|V|) anyways here? We could of course also specify it as O(b) where b is the node_bound and describe it that way? Whilst also saying that usually b = |V|.
(This might apply to other algs aswell)

RaoulLuque (commented):
I only had enough time to get to `bridges.rs` (alphabetically) but these are some suggestions which contain some general ones as well :)

src/algo/astar.rs
starovoid: In A*, it is very difficult to give a general time estimate, since everything strongly depends on the heuristic. Perhaps, to avoid confusion, it is worth leaving the `#Complexity` section empty until a good enough description is found.

src/algo/bridges.rs
starovoid: Oh, my bad. Of course, O(|V|)

src/algo/astar.rs
starovoid: I decided on the following option:
```markdown
/// The time complexity largely depends on the heuristic used. You can contribute to improve the documentation in this place ;)
///
/// With a trivial heuristic, the algorithm will work like [`fn@crate::algo::dijkstra`], but other situations are possible too.
```

RaoulLuque: I see, nice :+1:

src/algo/astar.rs
```diff
@@ -25,7 +25,21 @@ use crate::visit::{EdgeRef, GraphBase, IntoEdges, Visitable};
...
+/// * `estimate_cost`: closure that returns the estimated cost to the finish for particular node.
+///
+/// # Returns
+/// * `Some(K, Vec<G::NodeId>)` - the total cost and path from start to finish, if one was found.
+/// * `None` - if such a path was not found.
+///
+/// # Complexity
+/// The time complexity largely depends on the heuristic used. You can contribute to improve the documentation in this place ;)
+///
+/// With a trivial heuristic, the algorithm will work like [`fn@crate::algo::dijkstra`], but other situations are possible too.
```
RaoulLuque: Regarding this, I would m
```suggestion
/// The time complexity largely depends on the heuristic used. Feel free to contribute and provide the exact time complexity :)
///
/// With a trivial heuristic, the algorithm will behave like [`fn@crate::algo::dijkstra`].
```
I would just suggest rephrasing it a bit, especially with the emoji since the winky face emoji has some unwanted connotations in some contexts ^^

starovoid:
Hey @RaoulLuque, I think I've finished the work, except for the points mentioned above. I suggest we leave the isomorphism algorithms for later and wait for your review :)

RaoulLuque:
Very nice :) I'll try to look into it later today and also give my opinion on:

> Find out the space complexity of the greedy_feedback_arc_set algorithm

What is your opinion on removing the `[Generic]` Tags on the algorithms with the arguments I gave above?

starovoid:
> What is your opinion on removing the `[Generic]` Tags on the algorithms with the arguments I gave above?

At the moment, there are still several algorithms that accept specific graph types. You can suggest a change in wording for them, and then nothing will stop us from deleting `[Generic]` labels.

RaoulLuque:
I see. The thing is, that I am not sure users will understand the `[Generic]` tag as meaning that. But I do see the point that it would be good to point out the difference to users between algs which work on generic graphs (fulfilling only some trait bounds) and algs that only work on specific graphs :thinking: I'll think about that then and maybe open a PR/Issue in the future about it :)

RaoulLuque:
Regarding this:
>     * [ ]  Find out the space complexity of the `greedy_feedback_arc_set` algorithm
I would say the runtime is in O(|V| + |E|) and the space complexity is the same.

src/algo/coloring.rs
```diff
@@ -6,16 +6,26 @@ use hashbrown::{HashMap, HashSet};
...
+/// # Arguments
+/// * `graph`: undirected graph without loops.
+///
+/// # Returns
+/// * `(HashMap, usize)`: [`struct@hashbrown::HashMap`] that associates to each `NodeId` its color
+///   and the `usize` number of used colors.
+///
+/// # Complexity
+/// * Time complexity: **O((|V| + |E|)log(|V|)**.
+/// * Space complexity: **O(|V|)**.
```
RaoulLuque: I would think the space complexity is O(|V| + |E|), since the adj_color_map variable can store a color for each neighbor of each node. That results in possibly |E| many colors being stored?
```suggestion
/// * Space complexity: **O(|V| + |E|)**.
```

starovoid:
> I would say the runtime is in O(|V| + |E|) and the space complexity is the same.

Is there a proof for this estimate for the execution time? The author of reference paper gives O(|E|).

src/algo/feedback_arc_set.rs
```diff
@@ -15,13 +15,21 @@ use self::linked_list::{LinkedList, LinkedListEntry};
...
 /// Loops (edges to and from the same node) are always included in the returned set.
 ///
+/// # Arguments
+/// * `g`: a directed graph.
+///
+/// # Returns
+/// * `impl Iterator`:  the iterator of edge references `G::EdgeRef`.
+///
+/// # Complexity
+/// * Time complexity: **O(|E|)**, where **|E|** is the number of edges in the input graph.
```
RaoulLuque: See GitHub comment
```suggestion
/// # Complexity
/// * Time complexity: **O(|V| + |E|)**.
/// * Space complexity: **O(|V| + |E|)**.
///
/// where **|V|** is the number of nodes and **|E|** is the number of edges.
```

src/algo/feedback_arc_set.rs
```diff
@@ -15,13 +15,21 @@ use self::linked_list::{LinkedList, LinkedListEntry};
...
 ///
 /// Does not consider edge/node weights when selecting edges for the feedback arc set.
 ///
 /// Loops (edges to and from the same node) are always included in the returned set.
 ///
+/// # Arguments
+/// * `g`: a directed graph.
+///
+/// # Returns
+/// * `impl Iterator`:  the iterator of edge references `G::EdgeRef`.
```
RaoulLuque:
```suggestion
/// * `impl Iterator`:  the iterator of edge references `G::EdgeRef` in the feedback arc set.
```

src/algo/floyd_warshall.rs
```diff
@@ -14,13 +14,19 @@ use crate::visit::{
...
 /// * `Err`: if graph contains negative cycle.
 ///
+/// # Complexity
+/// * Time complexity: **O(|V|³)**.
+/// * Space complexity: **O(|V|²)**.
+///
+/// where **|V|** is the number of nodes.
+///
 /// **Note**: If the graph is sparse (the number of edges is quite small),
 /// consider using the [`johnson`](fn@crate::algo::johnson), which is likely to show better performance.
```
RaoulLuque:
```suggestion
/// consider using [`johnson's algorithm`](fn@crate::algo::johnson), which is likely to show better performance in such cases.
```

src/algo/ford_fulkerson.rs
```diff
@@ -109,15 +109,30 @@ where
...
+/// # Arguments
+/// * `network`: a wieghted directed graph.
+/// * `source`: a stream *source* node.
+/// * `destination`: a stream *sink* node.
+///
+/// # Returns
+/// Returns a tuple of two values:
+/// * `N::EdgeWeight`: computed maximum flow;
+/// * `Vec<N::EdgeWeight>`: the flow of each edge.
+///   The vector is indexed by the graph's edge indices.
```
RaoulLuque: I really like this representation for tuple return types. Should we maybe also apply this to other algorithms? (For example, coloring). Other times, ` * (TypeA, TypeB): Explanation` was used instead. 

```suggestion
/// Returns a tuple of two values:
/// * `N::EdgeWeight`: computed maximum flow;
/// * `Vec<N::EdgeWeight>`: the flow of each edge. The vector is indexed by the graph's edge indices.
```

src/algo/k_shortest_path.rs
```diff
@@ -10,18 +10,28 @@ use crate::visit::{EdgeRef, IntoEdges, NodeCount, NodeIndexable, Visitable};
 /// \[Generic\] k'th shortest path algorithm.
 ///
 /// Compute the length of the k'th shortest path from `start` to every reachable
```
RaoulLuque:
```suggestion
/// Compute the length of the k-th shortest path from `start` to every reachable
```

src/algo/k_shortest_path.rs
```diff
@@ -10,18 +10,28 @@ use crate::visit::{EdgeRef, IntoEdges, NodeCount, NodeIndexable, Visitable};
...
+///   to compute path costs. Edge costs must be non-negative.
+///
+/// # Returns
+/// * `HashMap`: [`struct@hashbrown::HashMap`] that maps `NodeId` to path cost.
+///
+/// # Complexity
+/// * Time complexity: **O(k(|E| + |V|log(|V|)))**.
+/// * Space complexity: **O(|V| + k|E|)**.
+///
+/// where **|V|** is the number of nodes and **|E|** is the number of edges.
```
RaoulLuque: I think the complexity is slightly different:
```suggestion
/// # Complexity
/// * Time complexity: **O(k|E| log(k|E|))**.
/// * Space complexity: **O(|V| + k|E|)**.
///
/// where **|V|** is the number of nodes, **|E|** is the number of edges and **k** is the provided parameter.
```
For the space with the rationale that: In the worst case, put each of the |E| the maximum number of k times in the visit_next binary heap (first k |E| factor). Then each pop from the binary heap is log(#Entries) and the heap will then contain up to k |E| edges (second log ( k |E| ) factor).

src/algo/matching.rs
```diff
@@ -312,11 +324,23 @@ impl<G: GraphBase> PartialEq for Label<G> {
...
 /// **Panics** if `g.node_bound()` is `usize::MAX`.
 ///
+/// # Arguments
+/// * `graph`: an undirected graph.
+///
+/// # Returns
+/// * [`struct@Matching`]: computed maximum matching.
+///
+/// # Complexity
+/// * Time complexity: **O(|V|³)**.
```
RaoulLuque, line 339: I am not sure about this time complexity. I feel like it could also be something slightly different (slightly faster in most cases):
O(|V| |E| + |V|^2). But I am not certain and don't have enough time unfortunately to look into the details.

src/algo/maximal_cliques.rs
```diff
@@ -72,11 +72,17 @@ where
...
 /// if an edge (u, v) exists, then (v, u) also exists.
 ///
-/// ## Arguments
+/// # Arguments
 /// * `g`: The graph to find maximal cliques in.
 ///
-/// ## Returns
-/// A vector of sets making up the maximal cliques in the graph.
+/// # Returns
+/// * A vector of sets making up the maximal cliques in the graph.
```
RaoulLuque:
```suggestion
/// * `Vec<HashSet>`: A vector of [`struct@hashbrown::HashSet`] making up the maximal cliques in the graph.
```

src/algo/maximal_cliques.rs
```diff
@@ -72,11 +72,17 @@ where
...
+///
+/// # Complexity
+/// * Time complexity: **O(3^(|V|/3))**
+/// * Space complexity: **O(|V|²)** plus the space needed for sets R, P, and X.
+///
+/// where **|V|** is the number of nodes.
```
RaoulLuque: I would think that R, P and X are dominated by O(d|V|) where d is the number of maximal cliques (which can be up to O(3^ (n / 3)) many). So we could write something like:
```suggestion
/// * Space complexity: **O(|V|² + d|V|)** 
///
/// where **|V|** is the number of nodes and **d** is the number of maximal cliques (possibly up to **O(3^(|V|/3))** many).```

**Review by RaoulLuque — COMMENTED** [review]
This is how far I got for now (min_spanning_tree.rs - alphaetically) :)

**Comment by RaoulLuque** [comment]
> Is there a proof for this estimate for the execution time? The author of reference paper gives O(|E|).

Well, the + |V| term just makes sure that if there are not edges, the bound is still correct. I mean, we compute a good_node_sequence. And to actually compute the sequence, we will have to somehow iterate over each node at least once. So that's where I would imagine the + |V| term comes from.

File: src/algo/coloring.rs (hunk 12 above)
Comment (starovoid) [reply src/algo/coloring.rs]: Nice! Let it be O(|V| + |E|) for the case of graphs where |E| < |V|.

File: src/algo/floyd_warshall.rs (hunk 15 above)
Comment (starovoid) [reply src/algo/floyd_warshall.rs]: Maybe 
```markdown
/// consider using [`johnson`](fn@crate::algo::johnson), which is likely to show better performance in such cases.
```
?

starovoid:
> I'm not sure I'll be able to finish the review today. Until when would you need/like to have it?

There's no need to rush 👌, it's better to fix all the inaccuracies before release.

src/algo/floyd_warshall.rs
RaoulLuque: Not sure. It just sounds weird to me to use `johnson` as a noun. That way it sounds like johnson is something/someone you can use.
If you don't want to make everything a link, maybe instead:
```
/// consider using [`johnson`](fn@crate::algo::johnson)'s algorithm, which is likely to show better performance in such cases.
```

src/algo/matching.rs
starovoid: Are you saying this because you see the difference between the current implementation and Gabow's [article](https://dl.acm.org/doi/10.1145/321941.321942)? The O(V^3) estimate is taken from there. I haven't checked the implementation for compliance with the article yet either.

src/algo/maximal_cliques.rs
starovoid: Of course, the amount of memory used by clicks is greater |P|+|R|+|X|. But I was referring to the amount of auxiliary space. I just came up with the idea to replace "Space complexity" with "Auxiliary space", it will be more understandable.

starovoid: I think it's worth specifying the auxiliary space, since for algorithms that return iterators, the total space complexity cannot be calculated without additional conditions.

src/algo/ford_fulkerson.rs
starovoid: Changed the section in `dsatur_coloring` to:
```markdown
/// # Returns
/// * [`struct@hashbrown::HashMap`] that associates to each `NodeId` its color.
/// * `usize`: the number of used colors.
```

In other places, everything seems to look fine, except that we also can try to simplify `Some(a, b)`.

RaoulLuque: Nice :) Why not change it to:
/// # Returns
/// Returns a tuple of two values:
/// * [`struct@hashbrown::HashMap`] that associates to each `NodeId` its color.
/// * `usize`: the number of used colors.
To make it completely consistent?

src/algo/coloring.rs
```diff
@@ -6,16 +6,26 @@ use hashbrown::{HashMap, HashSet};
...
-/// It must implement `IntoEdges`, `IntoNodeIdentifiers` and `Visitable`
-/// Returns a tuple composed of a HashMap that associates to each `NodeId` its color and the number of used colors.
 ///
-/// Computes in **O((|V| + |E|)*log(|V|)** time
+/// # Arguments
+/// * `graph`: undirected graph without loops.
+///
+/// # Returns
+/// * [`struct@hashbrown::HashMap`] that associates to each `NodeId` its color.
+/// * `usize`: the number of used colors.
```
RaoulLuque: See comments in ford_fulkerson.rs
```suggestion
/// # Returns
/// Returns a tuple of:
/// * [`struct@hashbrown::HashMap`] that associates to each `NodeId` its color.
/// * `usize`: the number of used colors.
```

src/algo/matching.rs
RaoulLuque: I actually did not look at the original article, but just the code. But as I said, I am not sure :thinking: In any case we can leave it at O(V^3) I'd say, since that is an upper bound in any case, it may only not be as precise as possible :+1:

src/algo/mod.rs
```diff
@@ -126,6 +140,19 @@ where
...
+///
+/// # Arguments:
+/// `g`: an input graph that always treated as undirected.
+///
+/// # Returns
+/// `true`: if the input graph contains a cycle.
+/// `false`: otherwise.
+///
+/// # Complexity
+/// * Time complexity: amortized **O(|E|)**.
```
RaoulLuque, line 152: Why does it say amortized here? Because of the union find? In that case, it would have to say that for `connected_components` above too, right?

src/algo/simple_paths.rs
```diff
@@ -11,10 +11,31 @@ use crate::{
     Direction::Outgoing,
 };
 
+/// Calculate all simple paths with specified constraints from node `from` to node `to`.
+///
+/// The simple path is a path without repetitions.
```
RaoulLuque:
```suggestion
/// A simple path is a path without repeating nodes.
```

src/algo/simple_paths.rs
```diff
@@ -11,10 +11,31 @@ use crate::{
...
+/// The number of simple paths between a given pair of vertices almost always grows exponentially,
+/// reaching `O(V!)` on a dense graphs at `V` vertices.
```
RaoulLuque:
```suggestion
/// The number of simple paths between a given pair of vertices can grow exponentially,
/// reaching `O(|V|!)` on complete graphs with `|V|` vertices.
```

src/algo/simple_paths.rs
```diff
@@ -11,10 +11,31 @@ use crate::{
...
+///
+/// So if you have a large enough graph, be prepared to wait for the results for years.
+/// Or consider extracting only part of the simple paths using the adapter [`Iterator::take`].
```
RaoulLuque:
```suggestion
/// So if you have a large enough graph, be prepared to wait on the results for years.
/// Or consider extracting only part of the simple paths using the adapter [`Iterator::take`].
/// Also note, that this algorithm does not check that a path exists between `from` and `to`. This may lead to very long running times and it may be worth it to check if a path exists before running this algorithm on large graphs.
```
(Taken from the networkx page)

src/algo/simple_paths.rs
```diff
@@ -11,10 +11,31 @@ use crate::{
...
+/// * `to`: a target node of desired paths.
+/// * `min_intermediate_nodes`: the minimum number of nodes in the desired paths.
+/// * `max_intermediate_nodes`: the maximum number of nodes in the desired paths (optional).
+/// # Returns
 /// Returns an iterator that produces all simple paths from `from` node to `to`, which contains at least `min_intermediate_nodes` nodes
-/// and at most `max_intermediate_nodes`, if given, or limited by the graph's order otherwise. The simple path is a path without repetitions.
+/// and at most `max_intermediate_nodes`, if given, or limited by the graph's order otherwise.
+///
+/// # Complexity
+/// * Time complexity: for computing first **k** paths, the running time will be **O(k|V| + k|E|)**.
```
RaoulLuque:
```suggestion
/// * Time complexity: for computing the first **k** paths, the running time will be **O(k|V| + k|E|)**.
```

src/algo/spfa.rs
```diff
@@ -15,13 +15,20 @@ use alloc::{vec, vec::Vec};
 /// ## Arguments
 /// * `graph`: weighted graph.
 /// * `source`: the source vertex, for which we calculate the lengths of the shortest paths to all the others.
-/// * `edge_cost`: closure that returns cost of a particular edge
+/// * `edge_cost`: closure that returns cost of a particular edge.
```
RaoulLuque:
```suggestion
/// * `edge_cost`: closure that returns the cost of a particular edge.
```

src/algo/spfa.rs
```diff
@@ -15,13 +15,20 @@ use alloc::{vec, vec::Vec};
...
 /// ## Returns
 /// * `Err`: if graph contains negative cycle.
 /// * `Ok`: a pair of a vector of shortest distances and a vector
 ///   of predecessors of each vertex along the shortest path.
 ///
+/// ## Complexity
+/// * Time complexity: **O(|V||E|)**, but it's generally assumed that in the average case it is **O(|E|)**.
+/// * Auxiliary space: **O(|V|)**.
+///
+/// |V| is the number of nodes, |E| is the number of edges in the input graph.
```
RaoulLuque: For consistency:
```suggestion
/// where **|V|** is the number of nodes and **|E|** is the number of edges.
```

src/algo/steiner_tree.rs
```diff
@@ -140,7 +140,7 @@ where
 /// A `StableGraph` containing the nodes and edges of the Steiner tree.
 ///
 /// ## Complexity
-/// Time complexity is **O(|S| |V|²)**.
+/// Time complexity is **O(|S| |V|²)**,
```
RaoulLuque: For consistency:
```suggestion
/// Time complexity: **O(|S| |V|²)**.
```

src/algo/tred.rs
```diff
@@ -21,20 +21,29 @@ use crate::Direction;
...
-/// from [`toposort`]).
+/// # Arguments
+/// * `g`: a directed acyclic graph.
+/// * `toposort`: a topological order on the node indices of `g` (for example obtained from [`toposort`](fn@crate::algo::toposort)).
 ///
-/// [`toposort`]: ../fn.toposort.html
-///
-/// Returns a pair of a graph `res` and the reciprocal of the topological sort `revmap`.
+/// # Returns
+/// Returns a pair of a graph `res` represented with [`UnweightedList`](type@crate::adj::UnweightedList) and the reciprocal of the topological sort `revmap`:
```
RaoulLuque:
```suggestion
/// Returns a tuple of:
/// * [`UnweightedList`](type@crate::adj::UnweightedList) `res` graph.
/// * `Vec`: reciprocal of the topological sort `revmap`.
```

src/algo/tred.rs
```diff
@@ -93,23 +98,22 @@ where
...
 ///
-/// Runtime complexity: **O(|V| + \sum_{(x, y) \in Er} d(y))** where **d(y)**
-/// denotes the outgoing degree of **y** in the transitive closure of **G**.
-/// This is still **O(|V|³)** in the worst case like the naive algorithm but
-/// should perform better for some classes of graphs.
+/// # Complexity
+/// * Time complexity: **O(|V| + \sum_{(x, y) \in Er} d(y))** where **d(y)**
+///   denotes the outgoing degree of **y** in the transitive closure of **G**.
+///   This is still **O(|V|³)** in the worst case like the naive algorithm but
+///   should perform better for some classes of graphs.
```
RaoulLuque:
```suggestion
/// # Complexity
/// * Time complexity: **O(|V| + \sum_{(x, y) \in Er} d(y))** where **d(y)**
///   denotes the outgoing degree of **y** in the transitive closure of **G**
///   and **Er** the edge set of the transitive reduction.
///   This is still **O(|V|³)** in the worst case like the naive algorithm but
///   should perform better for some classes of graphs.
```

RaoulLuque (commented):
Alright, I am done now :) Thanks a lot again for the changes, seems like a great addition to make the petgraph more consistent and approachable for users :star2:

src/algo/mod.rs
starovoid: Yes, thanks for the catch.

src/algo/k_shortest_path.rs
starovoid: Makes sense, thanks!

PR #825, opened 2025-06-09
docs: Fix time and auxiliary space complexity of bron-kerbosch

RaoulLuque:
I think the auxiliary space and time complexity as recently updated in #815, are not quite right for Bron-Kerbosch.

For one, the auxiliary space complexity cannot be right. Since, we return a `Vec` with `HashSet`, the auxiliary space has to be at least equal to the number of maximal cliques in the graph. [Moon and Moser showed](https://link.springer.com/article/10.1007/BF02760024) that this can be up to `3 ^ (|V| / 3)` many. Thus, I suggest introducing `k` as a variable to denote the number of maximal cliques in the graph.

The time complexity might also be able to be written in terms of `k`, however this would probably also depend on this implementation of Bron-Kerbosch with pivoting. Instead one could write something like:

```
/// * Time complexity: **O(p(|V|)k)**
```
For some polynomial `p`, which makes clear that the runtime is not always exponential in the number of vertices, but is merely dominated by the number of maximal cliques in the graph.


starovoid:
The "Auxiliary" space means exactly the auxiliary space, not the size of output data. For greater clarity, it is really useful to specify `k`, but as a separate comment, not as a replacement for the current doc-line.

RaoulLuque:
I think might be using different definitions of auxiliary space and I am not sure which one is correct. I looked at the one from [wikipedia](https://en.wikipedia.org/wiki/Space_complexity#Auxiliary_space_complexity) which states:

```
The term auxiliary space refers to space other than that consumed by the input. 
```
Thus, the space used by the output would also be part of the auxiliary space. This is confirmed by the more precise definition using Turing Machines, which follows the above one and can be found using the same link:
```
Auxiliary space complexity could be formally defined in terms of a 
[Turing machine](https://en.wikipedia.org/wiki/Turing_machine) 
with a separate input tape which cannot be written to, only read, 
and a conventional working tape which can be written to. The 
auxiliary space complexity is then defined (and analyzed) via the 
working tape.
```

starovoid:
> Thus, the space used by the output would also be part of the auxiliary space. This is confirmed by the more precise definition using Turing Machines, which follows the above one and can be found using the same link:

Then I suggest this: in situations where the output size exceeds the size of the auxiliary space, describe both components.

This difference makes sense due to the existence of algorithms that return iterators, in which it is impossible to unambiguously specify the space complexity, since the output can simply be discarded.

RaoulLuque:
> Then I suggest this: in situations where the output size exceeds the size of the auxiliary space, describe both components.
> 
> This difference makes sense due to the existence of algorithms that return iterators, in which it is impossible to unambiguously specify the space complexity, since the output can simply be discarded.

Okay, I see the possible ambiguity with returned Iterators. However, in that case we should still be able to clearly compute the auxiliary space complexity. Take for example a function which would return an iterator of the numbers from 1 to n. If it is implemented storing only the previously outputted number then I agree that the auxiliary space used is O(1). If it instead all previously outputted number (would be inefficient of course), then it is O(n).

However, in this case we actually store the entire output (all maximal cliques) at at least one point in time in the function. Thus it is auxiliary space used by the function. Therefore, I would definitely lean towards including this in the auxiliary space complexity of the function itself for this case.

starovoid:
> [Moon and Moser showed](https://link.springer.com/article/10.1007/BF02760024) that this can be up to `3 ^ (|V| / 3)` many. Thus, I suggest introducing `k` as a variable to denote the number of maximal cliques in the graph.

Shouldn't we write **O(|V|^2 +|V|k)**, since each clique can contain up to **|V|** nodes?

RaoulLuque:
> > [Moon and Moser showed](https://link.springer.com/article/10.1007/BF02760024) that this can be up to `3 ^ (|V| / 3)` many. Thus, I suggest introducing `k` as a variable to denote the number of maximal cliques in the graph.
> 
> Shouldn't we write **O(|V|^2 +|V|k)**, since each clique can contain up to **|V|** nodes?

Yes and No ^^ I think this is not clear and depends on the author, but since k can be exponential, in that case the V term would vanish in the Landau Notation and thus some leave it out in general.

However, when looking at the complete graph with increasing n for example, the k would be constant which would not make a lot of sense of course. Thus, I think you have a point :) So yeah, that would be even more precise, good point 😁

RaoulLuque:
Just adapted it accordingly :sun_with_face:

PR #828, opened 2025-06-11
refactor: Remove unnecessary trait bounds from impls/methods

regexident:
The goal of this PR is to make the following scenario possible:

```rust
#[derive(Default, Clone)]
struct DummyWrapper<Node, Edge, Ix> {
    graph: StableDiGraph<Node, Edge, Ix>,
}
```

… which currently requires a manual implementation of both, `Default` and `Clone`:

```rust
struct DummyWrapper<Node, Edge, Ix> {
    graph: StableDiGraph<Node, Edge, Ix>,
}

impl<Node, Edge, Ix> Clone for DummyWrapper<Node, Edge, Ix>
where
    Node: Clone,
    Edge: Clone,
    Ix: Clone,
{
    fn clone(&self) -> Self {
        Self {
            graph: self.graph.clone(),
        }
    }
}

impl<Node, Edge, Ix> Default for DummyWrapper<Node, Edge, Ix>
where
    Node: Default,
    Edge: Default,
    Ix: Default,
{
    fn default() -> Self {
        Self {
            graph: Default::default(),
        }
    }
}
```

… due to the following trait bounds getting in the way of the derives:

```
error[E0277]: the trait bound `Ix: petgraph::adj::IndexType` is not satisfied
 --> web-api/crates/interactive-graph/src/lib.rs:3:5
  |
1 | #[derive(Default, Clone)]
  |          ------- in this derive macro expansion
2 | struct DummyWrapper<Node, Edge, Ix> {
3 |     graph: StableDiGraph<Node, Edge, Ix>,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `petgraph::adj::IndexType` is not implemented for `Ix`

error[E0277]: the trait bound `Ix: petgraph::adj::IndexType` is not satisfied
 --> web-api/crates/interactive-graph/src/lib.rs:3:5
  |
1 | #[derive(Default, Clone)]
  |                   ----- in this derive macro expansion
2 | struct DummyWrapper<Node, Edge, Ix> {
3 |     graph: StableDiGraph<Node, Edge, Ix>,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `petgraph::adj::IndexType` is not implemented for `Ix`
```

With this PR's changes the initial scenario compiles as is, using derives.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -407,10 +407,11 @@ pub type DiGraph<N, E, Ix = DefaultIx> = Graph<N, E, Directed, Ix>;
 pub type UnGraph<N, E, Ix = DefaultIx> = Graph<N, E, Undirected, Ix>;
 
 /// The resulting cloned graph has the same graph indices as `self`.
-impl<N, E, Ty, Ix: IndexType> Clone for Graph<N, E, Ty, Ix>
+impl<N, E, Ty, Ix> Clone for Graph<N, E, Ty, Ix>
 where
     N: Clone,
     E: Clone,
+    Ix: Copy,
 {
     fn clone(&self) -> Self {
         Graph {
@@ -524,11 +525,7 @@ impl<N, E> Graph<N, E, Undirected> {
     }
 }
 
-impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix>
-where
-    Ty: EdgeType,
-    Ix: IndexType,
-{
+impl<N, E, Ty, Ix> Graph<N, E, Ty, Ix> {
     /// Create a new `Graph` with estimated capacity.
     pub fn with_capacity(nodes: usize, edges: usize) -> Self {
         Graph {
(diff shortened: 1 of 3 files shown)


RaoulLuque:
Thanks for the PR :sauropod: The ability to derive would indeed be quite neat ( or even expected I guess ) ^^

LGTM :)

RaoulLuque:
*Just FYI, since I don't have merge rights yet, someone else will have to look at it as well, but I think that'll be soon :)

RaoulLuque:
What do you think @starovoid ?

The only possible issue I could see would be that we would require the "stronger" trait bounds in the future and reintroducing them would then constitue a breaking change. However, I cannot come up with a scenario where this might be necessary. In particular, since most of the changes are separated into individual impl blocks with different trait bounds

starovoid:
> What do you think @starovoid ?
> 
> The only possible issue I could see would be that we would require the "stronger" trait bounds in the future and reintroducing them would then constitue a breaking change. However, I cannot come up with a scenario where this might be necessary. In particular, since most of the changes are separated into individual impl blocks with different trait bounds

I apologize for the long wait. I see no reason to reject this change and find it very useful. If stricter trait boundaries are needed in the future, this will most likely affect the existing API, so the breaking change will happen anyway.

Issue #766, opened 2025-04-07
docs: Add new features to the README


ABorgna:
#758 added the new features to the main crate docs, but the README is still missing some.

The readme is rendered in the crates.io [landing page](https://crates.io/crates/petgraph), so it's an important place for feature visibility.

dlight:
Suggestion: maybe use [cargo readme](https://github.com/webern/cargo-readme) to add the main crate docs into the readme, keeping them in sync. Another plus is to show an usage example straight from crates.io, too.

RaoulLuque:
With the [new docs and README](https://github.com/petgraph/petgraph/pull/807), [cargo readme](https://github.com/webern/cargo-readme) does not seem that fit anymore. At least until it supports templating parts of the docs from docs into the README.

It would be very nice if we could take only the Example from the docs and put that into the README to keep it up to date and possibly the sections about the feature flags.

I opened a PR addressing the issue manually

PR #807, opened 2025-05-31
docs: Update docs.rs

RaoulLuque:
This PR updates the docs.rs by mainly replacing the existing example by several other ones, which go into more detail. It also restructures the main docs while leaving most of the information the same by adding an overview section.

The goal of this is to offer a more smooth introduction into petgraph, by increasing the clarity of the docs and adding a usage section to help people that are new to rust get started. With a similar intent, more examples are added to show off some of petgraphs' features. 

Among other things, this resolves #804. 

For a preview of the new docs, just run
```
cargo doc --open
```
after cloning this PR branch.

Note that for the revamp I took inspiration from the documentation style of several of [@BurntSushi 's crates](https://github.com/burntsushi).

PR #830, opened 2025-06-15
refactor: add examples for scc algorithms and reorganize into dedicated module

AnarchistHoneybun:
tasks:
- [x] Add examples to kosaraju_scc docs.
- [x] Add examples to tarjan_scc docs.
- [x] Move both algorithms into separate module (and re-export them).

PR #832, opened 2025-06-18
docs: Sync the crate feature flags in the README and docs

RaoulLuque:
This PR resolves https://github.com/petgraph/petgraph/issues/766.

With the [new docs and README](https://github.com/petgraph/petgraph/pull/807), using cargo readme as suggested in the issue seems unsuited, since the templating does not support including parts of the lib docs in the README.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -436,27 +436,32 @@ println!("Enhanced DOT format:\n{:?}", fancy_dot);
 
 # Crate features
 
+`petgraph` is built with these features enabled by default:
+
+* **graphmap** -
+  Enables [`GraphMap`](./graphmap/struct.GraphMap.html).
+* **stable_graph** -
+  Enables [`StableGraph`](./stable_graph/struct.StableGraph.html).
+* **matrix_graph** -
+  Enables [`MatrixGraph`](./matrix_graph/struct.MatrixGraph.html).
+* **std** -
+  Enables the Rust Standard Library. Disabling the `std` feature makes it possible to use `petgraph` in `no_std` contexts.
+
+Optionally, the following features can be enabled:
+
 * **serde-1** -
-  Defaults off. Enables serialization for ``Graph, StableGraph, GraphMap`` using
+  Enables serialization for ``Graph, StableGraph, GraphMap`` using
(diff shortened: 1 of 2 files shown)


RaoulLuque:
The breaking change seems to have something to do with https://github.com/petgraph/petgraph/pull/830

RaoulLuque:
Nvm, just created the PR at the wrong second while something else was being merged I guess and somehow conflicted with the current master

src/lib.rs
```diff
@@ -436,27 +436,32 @@ println!("Enhanced DOT format:\n{:?}", fancy_dot);
...
-  Defaults on. Enables the Rust Standard Library. Disabling the `std` feature makes it possible to use `petgraph` in `no_std` contexts.
-* **generate** -
-  Defaults off. Enables graph generators.
-* **unstable** -
-  Defaults off. Enables unstable crate features (currently only `generate`).
 * **dot_parser** -
-  Defaults off. Enables building [`Graph`](./graph/struct.Graph.html) and [`StableGraph`](./stable_graph/struct.StableGraph.html) from [DOT/Graphviz](https://www.graphviz.org/doc/info/lang.html) descriptions. Imports can be made statically or dynamically (i.e. at compile time or at runtime).
+  Enables building [`Graph`](./graph/struct.Graph.html) and [`StableGraph`](./stable_graph/struct.StableGraph.html) from [DOT/Graphviz](https://www.graphviz.org/doc/info/lang.html) descriptions. Imports can be made statically or dynamically (i.e. at compile time or at runtime).
+* **unstable** -
+  Enables unstable crate features (currently only [`generate`](./)).
```
starovoid:
```markdown
[`generate`](./)
```

What does it mean?

starovoid: Maybe just write **generate**?

RaoulLuque:
Regarding the `generate` comment in the review, thanks good catch :) I was thinking that it would be nice to link to the generate function in the docs, however the does not actually work since they are feature gated. Then I looked into generating docs with feature flags and that is something I will create another issue for :+1: 

Anyways, should be fixed now ^^

PR #834, opened 2025-06-20
chore: Add template for algorithms documentation

starovoid:
This PR adds a file describing the approach to documenting algorithms.

In the future, this file will be referenced by `CONTRIBUTING.md` and any other instructions for adding new algorithms.


RaoulLuque:
Looks great, thanks :sun_with_face:

assets/guide/algodocs.md
```diff
@@ -0,0 +1,114 @@
...
+4. `Returns` - a description of the output produced by the algorithm.
+5. `Complexity` - time complexity and the amount of auxiliary memory.
+6. `Examples` - sample usage to illustrate the algorithm and help users get started quickly.
+
+The description of the algorithm should also include a reference to the literature. It can be either a Wikipedia page or a link to a journal article, etc.
+
+## Example
+The `bellman_ford` algorithm documentation is given below as an example:
+```markdown
+/// \[Generic\] Compute shortest paths from node `source` to all other.
```
ABorgna: Should we have a definition/explanation for these `[Generic]` tags?
I'm inclined to drop them imo.

starovoid: Here I just copied it from the documentation. I think it's worth removing `Generic` from all documentation, since almost all algorithms are already "generic" and this label is losing relevance. I suggest opening a separate PR for this action.

RaoulLuque: I second this and can create a PR for that :+1:

ABorgna: Lets remove it from the example here then.
```suggestion
/// Compute shortest paths from node `source` to all other.
```

RaoulLuque: PR is [here](https://github.com/petgraph/petgraph/pull/835)

ABorgna (approved):

RaoulLuque:
Not sure what's going on with the rust stable job but I'm re-running it :thinking:

starovoid:
> Not sure what's going on with the rust stable job but I'm re-running it :thinking: 

I recently had a nightly test fail, so there must be a system-related reason.

RaoulLuque:
system-related as in GitHub CI runners or something in the crate that is flaky?

RaoulLuque:
I would merge this PR anyway, right? Since adding this markdown file should certainly not be the issue, right? :thinking:

starovoid:
> system-related as in GitHub CI runners or something in the crate that is flaky?

Maybe something changed in the testing env, and now the quickcheck test for the SPFA algorithm is failing. We need to change this test by not calling `.unwrap()` and comparing `spfa` results with `bellman_ford`.

```
thread 'test_spfa' panicked at tests/quickcheck.rs:1638:53:
called `Result::unwrap()` on an `Err` value: NegativeCycle(())
```

Issue #840, opened 2025-06-30
Feature Request: Add `into_nodes_edges` to `StableGraph` mirroring `Graph`


BloodStainedCrow:
Hello!

I recently (while trying to merge two `StableGraph`s) wanted to transform a `StableGraph` into an iterator (or a vec) of its NodeType. Unless I fully missed it, this is not possible with the current Api, unless one converts the `StableGraph` into a normal `Graph` first, which invalidates all `NodeIndex` and `EdgeIndex`.

Since anyone using a `StableGraph` is storing the indices somewhere (why else would they be using a `StableGraph`), this is just not an option.
My first solution included `clone`ing every Node, but that is not an option for everyone (and doubled the peak memory consumption of my program).

I then resorted to just forking `petgraph` and adding a simple implementation:

```rust
pub fn into_nodes_edges(self) -> (impl Iterator<Item = (NodeIndex<Ix>, N)>, impl Iterator<Item = (EdgeIndex<Ix>, NodeIndex<Ix>, NodeIndex<Ix>, E)>) {
    let (inner_nodes, inner_edges) = self.g.into_nodes_edges();

    (
        inner_nodes.into_iter().enumerate().filter_map(|(i, node)| node.weight.map(|node| (i, node))).map(|(index, node)| (NodeIndex::new(index), node)), 
        inner_edges.into_iter().enumerate().filter(|(_, edge)| edge.weight.is_some()).map(|(index, edge)| (EdgeIndex::new(index), edge.source(), edge.target(), edge.weight.unwrap()))
    )
}
```

This kind of API gives the user all the information that is stable and visible as a consumer of `StableGraph`, without exposing implementation details AFAIK.

starovoid:
Hello! 
Your proposed solution looks good in my opinion. Could you open a PR with this implementation?
I would also suggest naming the new function `into_nodes_edges_iters` so as not to be confused with [`Graph::into_nodes_edges`](https://docs.rs/petgraph/latest/petgraph/graph/struct.Graph.html#method.into_nodes_edges).

BloodStainedCrow:
@starovoid I have opened a PR: #841.

RaoulLuque:
Resolved in #841 🥳

PR #841, opened 2025-07-02
feat: Add `into_nodes_edges_iters` to `StableGraph`

BloodStainedCrow:
As discussed in #840.  

  This allows transforming a `StableGraph` into its components, without allocations (via `Graph::from::<StableGraph>`) and without losing the `NodeIndex` or `EdgeIndex`.

### DIFF
--- a/src/graph_impl/stable_graph/mod.rs
+++ b/src/graph_impl/stable_graph/mod.rs
@@ -817,6 +817,15 @@ where
         }
     }
 
+    pub fn into_nodes_edges(self) -> (impl Iterator<Item = (NodeIndex<Ix>, N)>, impl Iterator<Item = (EdgeIndex<Ix>, NodeIndex<Ix>, NodeIndex<Ix>, E)>) {
+        let (inner_nodes, inner_edges) = self.g.into_nodes_edges();
+
+        (
+            inner_nodes.into_iter().enumerate().filter_map(|(i, node)| node.weight.map(|node| (i, node))).map(|(index, node)| (NodeIndex::new(index), node)), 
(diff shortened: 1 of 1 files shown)


BloodStainedCrow:
Hm, should I turn the Iterator Items into a struct (instead of a tuple)? That would also make it clearer, which NodeIndex is source and target.

starovoid:
> Hm, should I turn the Iterator Items into a struct (instead of a tuple)? That would also make it clearer, which NodeIndex is source and target.

Nice idea! It'll significantly improve readability and make it easier to modify iterators later.

BloodStainedCrow:
Okay I have now packed the returned values into `StableGraphNode` and a `StableGraphEdge` structs

starovoid:
> Okay I have now packed the returned values into `StableGraphNode` and a `StableGraphEdge` structs

Nice! 
It's also worth adding a `derive(Debug)` and implement `Clone` for these structs if the associated types `N` and `E` are `Clone` and indices are `Copy`.

Like `Node` and `Edge` types:
```rust
impl<E, Ix> Clone for Node<E, Ix>
where
    E: Clone,
    Ix: Copy,
{...}

impl<E, Ix> Clone for Edge<E, Ix>
where
    E: Clone,
    Ix: Copy,
{...}
```

BloodStainedCrow:
Is the `Ix: Copy` bound necessary?
Since `Ix: IndexType` implies `Ix: Copy` right?

BloodStainedCrow:
Ah, I see. I currently have a `where Ix: IndexType` bound on the struct itself.
Since that is different, from all other petgraph structs (like `Graph` and `Node` etc) I will move the bound to the impl instead.

starovoid (approved):

PR #842, opened 2025-07-03
feat!: Add optional support for `proptest` with a bunch of composable strategies

regexident:
This also adds a bunch of property tests that test the strategies themselves, in order to ensure their correctness.

The `petgraph::proptest` module is marked `pub` so that users of `proptest` have easy access to them for use in downstream property tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,16 +21,16 @@ jobs:
     strategy:
       matrix:
         include:
-          - rust: 1.64.0  # MSRV
+          - rust: 1.66.0  # MSRV
             required_features: std
           - rust: 1.81.0  # no_std MSRV
           - rust: stable
-            features: unstable quickcheck rayon
+            features: unstable quickcheck proptest rayon
             test_all: --all
           - rust: beta
             test_all: --all
           - rust: nightly
-            features: unstable quickcheck rayon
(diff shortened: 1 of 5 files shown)


starovoid:
Looks amazing!
It will probably take some time to study and think about everything well, but I definitely like the idea.

regexident:
> Looks amazing!

Thanks!

> It will probably take some time to study and think about everything well, but I definitely like the idea.

There is one gap in the current set of provided strategies that I'd still like to fill (otherwise I'm quite happy with them):

There so far is no strategy for assigning a pre-defined set of edge weights to a randomized set of corresponding edge endpoints. The lack of such a strategy is annoying in scenarios where the edge weights are expected to be unique across the resulting graph.

RaoulLuque:
Added the breaking change label and adapted the PR name accordingly, since this constitutes a breaking change :+1:

regexident:
> There is one gap in the current set of provided strategies that I'd still like to fill (otherwise I'm quite happy with them):
> 
> There so far is no strategy for assigning a pre-defined set of edge weights to a randomized set of corresponding edge endpoints. The lack of such a strategy is annoying in scenarios where the edge weights are expected to be unique across the resulting graph.

Decided to approach this at the `proptest` level instead:

https://github.com/proptest-rs/proptest/pull/586

PR #843, opened 2025-07-03
feat: Implemented new largest maximal clique algorithm and function

michaelgoldenn:
Implemented the maximal clique algorithm as described in [this paper](https://doi.org/10.3390/a6040618) and added a new public function `largest_maximal_clique`. This works significantly faster if you are only interested in the largest maximal clique, and there is still room for parallelization which is currently not implemented in the algorithm. I needed to implement it for my current project which uses ~1M nodes and ~10M edges, and the time to compute went from ~6 hours to ~2.5 minutes.

This code is likely organized poorly, the code and tests are all fully functional AFAIK, but I would like some help regarding layout of the different functions.

### DIFF
--- a/benches/maximal_cliques.rs
+++ b/benches/maximal_cliques.rs
@@ -5,7 +5,7 @@ extern crate test;
 
 use test::Bencher;
 
-use petgraph::algo::maximal_cliques;
+use petgraph::algo::maximal_cliques::{maximal_cliques, largest_maximal_clique};
 
 #[allow(dead_code)]
 mod common;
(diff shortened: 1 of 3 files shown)


starovoid:
Hello! 

The tests and benches look good at first glance, all that remains is to add the `quickcheck` prop tests. The `maximal_cliques_mathces_ref_impl` from the `tests/quickcheck.rs` file can be helpful as an example.

michaelgoldenn:
Ok, I have added a quickcheck. Do you think it makes the most sense to keep the new `color_order` function in `maximal_cliques.rs` or instead should we move it to `coloring.rs` as it is a coloring algorithm? It's not very good as a coloring algorithm for most cases, but it might be good to move it there anyway in case it's useful for other algorithms?

starovoid:
> Ok, I have added a quickcheck. Do you think it makes the most sense to keep the new `color_order` function in `maximal_cliques.rs` or instead should we move it to `coloring.rs` as it is a coloring algorithm? It's not very good as a coloring algorithm for most cases, but it might be good to move it there anyway in case it's useful for other algorithms?

I think it's worth leaving it where it is.

michaelgoldenn:
I am having some trouble with this. I believe that since the newly implemented algorithm is recursive it fills up the stack when running the quicktest, causing an error (but has not had any issues running in any other circumstances that I've found). Does anyone know of a good solution or some ideas that could help?

starovoid:
> I am having some trouble with this. I believe that since the newly implemented algorithm is recursive it fills up the stack when running the quicktest, causing an error (but has not had any issues running in any other circumstances that I've found). Does anyone know of a good solution or some ideas that could help?

`quickcheck` generates graphs containing self-loops, which cause this stack overflow. 
I can suggest two solutions:
1. Handle self-loops in the algorithm;
2. Filter out self-loops in the test and write in the algorithm's doc comment that self-loops are prohibited.

The first option is preferable if it does not lead to a noticeable decrease in performance.

michaelgoldenn:
> `quickcheck` generates graphs containing self-loops, which cause this stack overflow

Aha, that was it, thank you! I will work on a good solution

PR #845, opened 2025-07-06
ci: Use just and adapt CONTRIBUTING.md and ci.yml accordingly

### LINKED ISSUE #844: Better documentation of clippy lints
Currently, if someone follows the instructions present in CONTRIBUTING.md for creating a new pull request, they can have fully working code on their machine but the code will still fail in Github actions because of the harsher clippy lints. If there already is an easy command to check if it will work on the dev's machine, then I think it should be documented in CONTRIBUTING.md

RaoulLuque:
This addresses the problem with CONTRIBUTING.md found in #844. 

Additionally, further details about CI actions which might be interesting to contributors were added such as miri, as well as running tests on different toolchains / rust versions.

Resolves #844

### DIFF
--- a/justfile
+++ b/justfile
@@ -34,6 +34,9 @@ fmt:
 clippy:
     cargo clippy --all-features --lib --bins --examples --tests -- -D warnings
 
+# Runs all linting checks that are run in CI
+lint: fmt clippy
+
 # Checks if no-std is working same as in CI. Requires the wasm32v1-none target to be installed
 check-no-std:
     cargo check --no-default-features -p petgraph --target wasm32v1-none --features graphmap,serde-1,stable_graph,matrix_graph,generate,unstable
--- a/justfile
+++ b/justfile
@@ -37,6 +37,7 @@ clippy:
 # Runs all linting checks that are run in CI
 lint: fmt clippy
 
+# Runs all tests and linting that are run in CI
 ci: fmt clippy test
 
 # Checks if no-std is working same as in CI. Requires the wasm32v1-none target to be installed
--- a/justfile
+++ b/justfile
@@ -4,7 +4,7 @@ default:
 
 # Builds the project in development mode
 build:
-    cargo build $args
+    cargo build
 
 # Tests with all features enabled
 test:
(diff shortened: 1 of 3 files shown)


RaoulLuque:
Ah, good point :thinking: I have to say that I am not the biggest fan of introducing non-cargo tooling, but since one can install it via cargo install, I'd be in favor as well :)

I'll rename this PR then ^^ I suppose I'd also change the CI and everything as well then, to use them everywhere, right?

RaoulLuque:
Okay, should be adapted accordingly now :) Let me know if there's anything you think we should include / change :sauropod:

justfile
```diff
@@ -0,0 +1,42 @@
+# Lists all available recipes
+default:
+    just --list
+
+# Builds the project in development mode
+build:
+    cargo build $args
```
starovoid: It seems that `$args` are wrong here. `just build` doesn't work on my Linux machine:
```bash
❯ just build
cargo build $args
sh: line 1: args: unbound variable
error: Recipe `build` failed on line 7 with exit code 127
```

```suggestion
    cargo build
```

starovoid (changes requested):
Everything is good and beautiful except `just build` command.

RaoulLuque:
Alright, thanks :) I felt like it could be nice to have one command which contributors can use which checks most of the things in one, so I added the `just ci` job, which just runs cargo test and the lints in one and adapted the CONTRIBUTING.md accordingly :)

Btw, I forgot to add your suggestion here in Github, so now it's in my commit instead, sry :o

RaoulLuque:
@starovoid do you accept the new version :) ? Because GitHub is still showing you as requesting changes and I don't want to just dismiss your request ^^

starovoid:
> @starovoid do you accept the new version :) ? Because GitHub is still showing you as requesting changes and I don't want to just dismiss your request ^^

Everything looks great, I just haven't had a chance to look at it yet ^^

RaoulLuque:
Ah, sry. I missunderstood then :sweat_smile:

PR #848, opened 2025-07-08
feat: add Louvain community detection algorithm

michaelgoldenn:
Implemented the louvain algorithm for community detection. I used a *lot* of code from @jpacold's implementation in rustworkx [here](https://github.com/Qiskit/rustworkx/pull/1277). As such, much of the code structure any style is based on their work, and therefore may not adhere to the styling standards common in petgraph. I've tried to go through everything that I can and design rigorous test cases, but please look over this PR with extra care and let me know if there is anything I've forgotten or if you think the structure should be changed at all. Thanks :)

### DIFF
--- a/src/algo/community/metrics.rs
+++ b/src/algo/community/metrics.rs
@@ -3,7 +3,7 @@ use crate::visit::{
     NodeCount, NodeIndexable,
 };
 use alloc::{vec, vec::Vec};
-use core::{error::Error, fmt, hash::Hash};
+use core::{fmt, hash::Hash};
 use hashbrown::HashSet;
 
 /// Traits for graphs to compute modularity
@@ -186,7 +186,6 @@ where
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct NotAPartitionError;
-impl Error for NotAPartitionError {}
 impl fmt::Display for NotAPartitionError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(


michaelgoldenn:
Does anyone have any idea on why this breaks the API? My guess is because the code is in a directory instead of directly inside of `src`? Because I don't change any existing code here besides adding the new directory to `src/mod.rs`

starovoid:
> Does anyone have any idea on why this breaks the API? My guess is because the code is in a directory instead of directly inside of `src`? Because I don't change any existing code here besides adding the new directory to `src/mod.rs`

The reason is probably a branch mismatch: after the rebase, semver-checks have passed.

starovoid:
@RaoulLuque Please look at the implementation, it seems we can include this in `0.8.3`.

RaoulLuque:
@starovoid I'll have time to look at this earliest end of July. In any case, this is definitely not my field of knowledge, so not sure whether I'll have enough time even then to look into the topic and then check the implementation :thinking:  (Since it's quite a lot of code as well)

So no guarantees, but will put it on my list :+1: Possibly someone else could also have a look, since I'll probably miss some things

michaelgoldenn:
I will likely be reworking some major parts of this soon, so it's probably best to not look over this yet anyway. I'll convert to a draft for now as I don't think it's ready for release yet

RaoulLuque:
> I will likely be reworking some major parts of this soon, so it's probably best to not look over this yet anyway. I'll convert to a draft for now as I don't think it's ready for release yet

I see, thanks for letting us know :) Just tag me or request a review once it's ready 🌞

michaelgoldenn:
I tried for a bit to rework the code into something more functional (as of the latest commit this code is _not_ suitable to be introduced to the crate), but I got overwhelmed and burnt out. I will likely try again if nobody has implemented it after the planned changes in the next major version, but currently I think that it is too difficult of a problem for me to accomplish on my own and will instead focus on hopefully contributing to the library in other places.

RaoulLuque:
> I tried for a bit to rework the code into something more functional (as of the latest commit this code is _not_ suitable to be introduced to the crate), but I got overwhelmed and burnt out. I will likely try again if nobody has implemented it after the planned changes in the next major version, but currently I think that it is too difficult of a problem for me to accomplish on my own and will instead focus on hopefully contributing to the library in other places.

I see, thanks for letting us know :) Although it would of course be a nice addition, there is of course no strict necessity on offering such features, so definitely don't feel stressed or anything ☺️ Do whatever brings you joy and whatever you feel like, and if that's contributing in other ways that'd be great 🫶

PR #849, opened 2025-07-18
refactor: replace uses of 'crate::util::zip' with 'core::iter::zip'

cactusdualcore:
The function 'zip' was introduced in #747 as a part of adding `no_std` support. However, there is already such a function in `core::iter`. The MSRV stated in the description of #747 was 1.64 at minimum, but the function in `core` is stable since Rust 1.59. Thus, it is an unnecessary duplication and can be removed without consequences.

### DIFF
--- a/src/csr.rs
+++ b/src/csr.rs
@@ -4,6 +4,7 @@ use alloc::{vec, vec::Vec};
 use core::{
     cmp::{max, Ordering},
     fmt,
+    iter::zip,
     iter::{Enumerate, Zip},
     marker::PhantomData,
     ops::{Index, IndexMut, Range},
@@ -15,8 +16,6 @@ use crate::visit::{
     IntoEdges, IntoNeighbors, IntoNodeIdentifiers, IntoNodeReferences, NodeCompactIndexable,
     NodeCount, NodeIndexable, Visitable,
 };
-
-use crate::util::zip;
 
 #[doc(no_inline)]
 pub use crate::graph::{DefaultIx, IndexType};
--- a/src/util.rs
+++ b/src/util.rs
@@ -6,11 +6,3 @@ where
 {
     iterable.into_iter().enumerate()
 }
-
-pub fn zip<I, J>(i: I, j: J) -> iter::Zip<I::IntoIter, J::IntoIter>
-where
-    I: IntoIterator,
-    J: IntoIterator,
-{
-    i.into_iter().zip(j)
-}


cactusdualcore:
There is, unfortunately, no `core::iter::enumerate` or `std::iter::enumerate` function. If it is still wanted to remove the `utils.rs` file, I'd be happy to do that in another PR.

RaoulLuque:
LGTM :+1: Let's see what the tests say :)

RaoulLuque:
> There is, unfortunately, no `core::iter::enumerate` or `std::iter::enumerate` function. If it is still wanted to remove the `utils.rs` file, I'd be happy to do that in another PR.

I see, what would be your idea then on how to get rid of the `utils.rs` file?

cactusdualcore:
> I see, what would be your idea then on how to get rid of the `utils.rs` file?

Removing the `utils.rs` file is synonymous with (re-)moving the definition of `enumerate` in it, because that's all it contains, right? I think there are three options, from imho worst to best:

1. Put it in every file that uses it. No imports, but otherwise pretty stupid.
2. Use the `IntoIter` postfix form. I think its a shame that this is such an ugly piece of code for what is a fairly common pattern. 
3. Put `enumerate` in `lib.rs`, or anywhere else than in this otherwise empty file.

I would implement the third option, i.e. moving it to wherever it fits best, and delete that file in my PR.

BONUS OPTION: open an issue on the rustlang repository to add it to the core lib. It feels like a fairly useful function with really no maintenance to include it. If `core::zip` exists, why shouldn't this?

cactusdualcore:
Done!

It's actually a fairly tiny change, because the `enumerate` function is only used in two files. Put it in the `graph_impl` module now.

Given how infrequent of a pattern this _actually_ was, you should consider just `.into_iter().enumerate()`-ing, I think.

I have renamed the PR to reflect this second change.

RaoulLuque:
I see, thanks for the explanation :) I think I'd also lean towards just removing the helper function altogether and just `.into_iter().enumerate()`-ing everything, as what is does is already quite self-explanatory and a reasonable amount of code :thinking:

cactusdualcore:
@RaoulLuque Do you want me remove the helper function in this PR, or would you a prefer a second PR to do so?

RaoulLuque:
As you wish, I'd say :) But if you split it into two, I would reduce this one to simply introducing the `zip` from the `std` and revert the removal of the `util.rs` file and move that into the other PR I think :)

cactusdualcore:
Ok. I think I will do two PRs then! Thanks for your fast and in-detail feedback!

I'll try getting this PR done today, but I might take a few days to start on the second.

cactusdualcore:
This PR should not include the more recent command anymore and be just the replacing of `zip`. Feel free to rerun the tests and then merge. I'll probably open a new PR in the next few days.

Have a nice evening!

RaoulLuque:
Okay, great, thanks a lot :)

RaoulLuque:
I think the title was reversed, so I switched it up ^^

Side note: Just noticed, you're also from RWTH :)

cactusdualcore:
> Ok. I think I will do two PRs then! Thanks for your fast and in-detail feedback!
> 
> I'll try getting this PR done today, but I might take a few days to start on the second.

The second PR is #872

PR #851, opened 2025-07-26
refactor: Fix clippy (and other) lints

RaoulLuque:
This PR fixes some clippy lints about lifetime annotations which occur when running clippy like it is run in CI using the nightly toolchain.

More precisely, when running:
```bash
cargo clippy --all-features --lib --bins --examples --tests -- -D warnings
```
or alternatively `just clippy` when #845 is merged, one gets multiple warnings (Errors because of `-D warnings`) of the following sort:
```bash
error: lifetime flowing from input to output with different syntax can be confusing
   --> src/adj.rs:314:25
    |
314 |     pub fn edge_indices(&self) -> EdgeIndices<E, Ix> {
    |                         ^^^^^     ------------------ the lifetime gets resolved as '_
    |                         |
    |                         this lifetime flows to the output
    |
    = note: -D mismatched-lifetime-syntaxes implied by -D warnings
    = help: to override -D warnings add #[allow(mismatched_lifetime_syntaxes)]
help: one option is to remove the lifetime for references and use the anonymous lifetime for paths
    |
314 |     pub fn edge_indices(&self) -> EdgeIndices<'_, E, Ix> {
    |                                               +++
```

This PR fixes these lints by exactly adding an anonymous lifetime to the return types and thus making the lifetime flow more explicit. Furthermore, a tiny change in the order of `std` and `no-std` generic parameters in `matrix_graph.rs` is made to make RustRover lints happy.

### DIFF
--- a/src/graph_impl/stable_graph/mod.rs
+++ b/src/graph_impl/stable_graph/mod.rs
@@ -680,7 +680,7 @@ where
     }
 
     /// Return an iterator over the node indices of the graph
-    pub fn node_indices(&self) -> NodeIndices<N, Ix> {
+    pub fn node_indices(&self) -> NodeIndices<'_, N, Ix> {
         NodeIndices {
             iter: enumerate(self.raw_nodes()),
         }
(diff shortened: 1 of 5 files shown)


src/graph_impl/stable_graph/mod.rs
```diff
@@ -753,10 +753,10 @@ where
         &self,
         a: NodeIndex<Ix>,
         b: NodeIndex<Ix>,
-    ) -> EdgesConnecting<E, Ty, Ix> {
+    ) -> EdgesConnecting<'_, E, Ty, Ix> {
         EdgesConnecting {
             target_node: b,
-            edges: self.edges_directed(a, Direction::Outgoing),
```
RaoulLuque, line 759: This `Direction::` is removed, as it is redundant since `Outgoing` is in scope.

PR #853, opened 2025-07-26
docs: Fix output in Dot docstring example

RaoulLuque:
Checking the actual output of the example of the [Dot struct](https://docs.rs/petgraph/latest/petgraph/dot/struct.Dot.html), I noticed that the actual output has brackets for edge labels after the edges, which are missing in the docstring.

Thus, this PR adds them.


RaoulLuque:
Closed in favor of #856

PR #855, opened 2025-07-27
feat: Add dijkstra::with_dynamic_goal

SOF3:
Allow the goal node to be dynamically and lazily evaluated.

This can be useful in graphs where testing whether a node is the goal is costly (e.g. "a node is a goal if its hash is equal to a list of possible values")

### DIFF
--- a/src/algo/dijkstra.rs
+++ b/src/algo/dijkstra.rs
@@ -89,12 +89,90 @@ pub fn dijkstra<G, F, K>(
     graph: G,
     start: G::NodeId,
     goal: Option<G::NodeId>,
-    mut edge_cost: F,
+    edge_cost: F,
 ) -> HashMap<G::NodeId, K>
 where
     G: IntoEdges + Visitable,
     G::NodeId: Eq + Hash,
     F: FnMut(G::EdgeRef) -> K,
+    K: Measure + Copy,
+{
+    with_dynamic_goal(graph, start, |node| goal.as_ref() == Some(node), edge_cost).scores
+}
+
+/// Return value of [`with_dynamic_goal`].
+pub struct AlgoResult<N, K> {
+    /// A [`struct@hashbrown::HashMap`] that maps `NodeId` to path cost.
+    pub scores: HashMap<N, K>,
+    /// The goal node that terminated the search, if any was found.
+    pub goal_node: Option<N>,
+}
+
+/// Dijkstra's shortest path algorithm with a dynamic goal.
+///
+/// This algorithm is identical to [`dijkstra`],
+/// but allows matching multiple goal nodes, whichever is reached first.
+/// A node is considered a goal if `goal_fn` returns `true` for it.
+///
+/// See the [`dijkstra`] function for more details.
+///
+/// # Example
+/// ```rust
+/// use petgraph::Graph;
+/// use petgraph::algo::dijkstra;
+/// use petgraph::prelude::*;
+/// use hashbrown::HashMap;
+///
+/// let mut graph: Graph<(), (), Directed> = Graph::new();
+/// let a = graph.add_node(()); // node with no weight
+/// let b = graph.add_node(());
+/// let c = graph.add_node(());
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Hey, first off thanks for your PR :star2: 

I very much like the feature you are adding in your PR and also like the ideas you bring to the table, with the Custom Struct as Return Type, as well as taking closures as function parameters. I will have a closer look in the coming days :)

The thing is that there is the desire to convert a lot of the algorithms to return Custom Struct Return Types and possibly take closures as input similar to what [Pathfinding](https://github.com/evenfurther/pathfinding) does. However, we still need to figure out what the general direction/design of that should look like. Nevertheless, I am not opposed to taking a first step with your PR and possibly changing/adapting it at a later point in time :) Especially since your current change is not a breaking one and a future one would probably be a breaking change.

The only thing I could imagine to make sense to do, while we are at it, to not have to break an API in the future is, instead of taking a graph as an input, taking an `edges` or `neighbors` closure, which would replace `graph.edges(node)`. Thus we would be more flexible :thinking: However, this could also be done at a later point in time, and I don't want to bother you with it if you are simply interested in your desired feature ^^

I would be curious to hear what you think @starovoid

starovoid:
@RaoulLuque Hey, sorry for the long wait

First of all, I would like to note that it is a really useful feature. The same thing has been requested for a long time for the A* algorithm.


---
> to return Custom Struct Return Types 

This approach has already been implemented in places (for example, `bellman_ford::Paths`). In theory, we could effectively generalize this approach to all shortest path algorithms. However, any significant change will be breaking, and this should be discussed with the wider community, because we don't want to create unnecessary difficulties for `petgraph` users pursuit of beauty :)


> take closures as input similar to what [Pathfinding](https://github.com/evenfurther/pathfinding) does

It seems to me that this does not correspond well with `petgraph` style, because we provide abstraction and generality in a different way.

starovoid:
Hi @SOF3 , please run `cargo fmt` to fix Lints CI.

starovoid:
Hi @SOF3 , an error occurs when compiling tests for Rust 1.64.0 (our MSRV). Please rewrite these fragment, and I think there will be no further problems.


```
failures:

---- src/algo/dijkstra.rs - algo::dijkstra::with_dynamic_goal (line 120) stdout ----
error[E0658]: use of unstable library feature 'is_some_with'
  --> src/algo/dijkstra.rs:163:23
   |
46 | assert!(res.goal_node.is_some_and(|n| n == d || n == f));
   |                       ^^^^^^^^^^^
   |
   = note: see issue #93050 <https://github.com/rust-lang/rust/issues/93050> for more information

error[E0277]: can't compare `&petgraph::prelude::NodeIndex` with `petgraph::prelude::NodeIndex`
  --> src/algo/dijkstra.rs:163:41
   |
46 | assert!(res.goal_node.is_some_and(|n| n == d || n == f));
   |                                         ^^ no implementation for `&petgraph::prelude::NodeIndex == petgraph::prelude::NodeIndex`
   |
   = help: the trait `PartialEq<petgraph::prelude::NodeIndex>` is not implemented for `&petgraph::prelude::NodeIndex`
   = help: the trait `PartialEq` is implemented for `petgraph::prelude::NodeIndex<Ix>`

error[E0277]: can't compare `&petgraph::prelude::NodeIndex` with `petgraph::prelude::NodeIndex`
  --> src/algo/dijkstra.rs:163:51
   |
46 | assert!(res.goal_node.is_some_and(|n| n == d || n == f));
   |                                                   ^^ no implementation for `&petgraph::prelude::NodeIndex == petgraph::prelude::NodeIndex`
   |
   = help: the trait `PartialEq<petgraph::prelude::NodeIndex>` is not implemented for `&petgraph::prelude::NodeIndex`
   = help: the trait `PartialEq` is implemented for `petgraph::prelude::NodeIndex<Ix>`
```

SOF3:
fixed, please run CI again

PR #856, opened 2025-07-27
docs: Collection of Doc fixes

RaoulLuque:
This PR implements two quick Doc fixes as well as different smaller fixes:
- The link in the `lib.rs` docs was to `functions` in the `algo` module, instead of the top level `index.html`
- The top-level `index.html` in the `algo` module seemed very bare-bone, therefore I added a little bit of text instead of none
- Adds examples for different functions like `map` and `filter_map` are added.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,22 +18,23 @@ contribute to petgraph. **Do not let this guide intimidate you.** It
 should simply be a reference for you, which you can refer to when
 contributing. The structure of this guide is as follows:
 
-- [Conduct](#🫂-conduct)
+- [Conduct](#-conduct)
 
-- [Where We Need Help](#🧩-where-we-need-help)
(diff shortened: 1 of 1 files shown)


starovoid:
Hi! I see such mini doc fixes occur too often, how about leaving this one PR before the release and adding all new fixes to it? I'm a little worried about overloading the changelog :)

RaoulLuque:
On the one hand I agree with you, however on the other I also think that it is fine to have a long list of doc fixes in the changelog. Since the changelog is split into sections, I feel like it doesn't really negatively impact the readability. However, I think it shows that we care about the docs and we put work into them :) But I am happy to use this PR to consolidate future doc changes :+1:

starovoid:
Hi @RaoulLuque , could you include #875 in the current PR? Of course, with the preservation of authorship.

PR #863, opened 2025-07-31
feat: Add `map_owned` and `filter_map_owned` for `Graph` and `StableGraph`

### LINKED ISSUE #542: Add another `map` method taking `self` instead of `&self`

RaoulLuque:
This PR continues the work done in PR #794. Thus, it resolves #542.

It adds `map_owned` and `filter_map_owned` methods for both `Graph` and `StableGraph` which take ownership of the respective graphs and their respective associated data in comparison to the `map` and `filter_map` counterparts which only take `&` references.

Appropriate tests for this have been added. If desired, I can also add quickchecks, but this did not seem necessary.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -1506,7 +1506,11 @@ where
     /// as `self`.
     ///
     /// If you want a non-consuming version of this function, see [`map`](struct.Graph.html#method.map).
-    pub fn map_owned<F, G, N2, E2>(self, mut node_map: F, mut edge_map: G) -> Graph<N2, E2, Ty, Ix>
+    pub fn map_owned<F, G, N2, E2>(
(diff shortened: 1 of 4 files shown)


RaoulLuque:
> I just have an additional suggestion to add doc examples, first of all it would help beginners in Rust. But this is up to you, I consider this task a low priority.

That is a very good idea : ) I just didn't have time yet to add this yet, so I just added it to the PR #856 :+1:

PR #865, opened 2025-08-02
feat: Fix self-loop bug in all_simple_paths and enable multiple targets

### LINKED ISSUE #864: Bug in all_simple_paths when source=target and feature request for multiple targets
# Bug Report: all_simple_paths generates self-looping cycles when source=target

## Description
When using `all_simple_paths` with `source` equal to `target`, the function incorrectly generates self-looping cycles instead of returning an empty result or handling this edge case appropriately.

An example is as follows:
```rust
use petgraph::Graph;
use petgraph::algo::all_simple_paths;
use petgraph::graph::NodeIndex;
use std::collections::hash_map::RandomState;

fn main() {
    // Create a directed graph
    let mut graph = Graph::new();

    let a = graph.add_node("A");
    let b = graph.add_node("B");
    let c = graph.add_node("C");

    // Add edges for complete graph (every node connected to every other node)
    graph.add_edge(a, b, ());
    graph.add_edge(a, c, ());
    graph.add_edge(b, a, ());
    graph.add_edge(b, c, ());
    graph.add_edge(c, a, ());
    graph.add_edge(c, b, ());

    // Find all simple paths from A to A
    let paths: Vec<Vec<NodeIndex>> = all_simple_paths::<Vec<NodeIndex>, _, RandomState>(
        &graph, a, a, 0, None
    ).collect();

    println!("Found {} paths from A to A:", paths.len());
    for (i, path) in paths.iter().enumerate() {
        print!("Path {}: ", i + 1);
        for (j, &node) in path.iter().enumerate() {
            if j > 0 { print!(" -> "); }
            print!("{}", graph[node]);
        }
        println!();
    }
}
```

## Expected Behavior
When `source` equals `target`, the function should either:
- Return an empty iterator (no simple paths from a node to itself)
- Return a single path containing only the source node
(See networkx/networkx#6690 for discussion on the design choice here. In summary, they chose the second approach for consistency with other APIs)

## Actual Behavior
The function generates self-looping cycles, which violates the definition of "simple paths" (paths without repeated vertices).

**Example output:**
```
Found 2 paths from A to A:
Path 1: A -> C -> A
Path 2: A -> B -> A
```

These are cycles, not simple paths, since they repeat the starting vertex `A`.

---

# Feature Request: Support multiple targets in all_simple_paths

## Description
Currently, `all_simple_paths` only supports finding paths from a source to a single target. Adding support for multiple targets would provide significant performance benefits when searching for paths to any of several possible destinations.

Instead of calling `all_simple_paths` multiple times for different targets, a single call could find paths to any of the specified targets. This reduces redundant graph traversal when the source and path exploration overlap. Reference: https://github.com/networkx/networkx/pull/3138

Both issues affect the usability and correctness of the path-finding functionality in petgraph.

lazyhope:
### Description

This pull request resolves #864.

1.  **Fixes a bug in `all_simple_paths`**: The original implementation could return a self-looping path when `from`=`to`, as described in the issue.

2.  **Adds `all_simple_paths_multi` for multiple targets**: A new function, [`all_simple_paths_multi`](src/algo/simple_paths.rs:144), has been implemented. This function finds all simple paths from a single source node to a `HashSet` of possible target nodes, in a single DFS traversal.

Unit tests have been added to validate the fix and to ensure the correctness of the new multi-target pathfinding functionality across various graph types and edge cases.

TODOs:
- [x] Add benchmarks to the functions in `src/algo/simple_paths.rs`
- [x] Add docstrings for the new `all_simple_path_multi` function.

Help needed: Finalise the design choice to exclude the single node path when `source = target`

### DIFF
--- a/benches/simple_paths.rs
+++ b/benches/simple_paths.rs
@@ -3,59 +3,18 @@
 extern crate petgraph;
 extern crate test;
 
+mod common;
+
 use std::collections::hash_map::RandomState;
 
+use common::ungraph;
 use hashbrown::HashSet;
 use petgraph::algo::{all_simple_paths, all_simple_paths_multi};
 use petgraph::prelude::*;
 use test::Bencher;
 
 #[bench]
-fn simple_paths_bench(bench: &mut Bencher) {
-    static NODE_COUNT: usize = 5;
-    let mut g = Graph::new_undirected();
-    let nodes: Vec<NodeIndex> = (0..NODE_COUNT).map(|_| g.add_node(())).collect();
-
-    // Create a complete undirected graph - every pair of nodes has an edge
-    for i in 0..NODE_COUNT {
-        for j in i + 1..NODE_COUNT {
-            g.add_edge(nodes[i], nodes[j], ());
-        }
-    }
-
-    let from = nodes[0];
-    let to = nodes[NODE_COUNT - 1];
-
-    bench.iter(|| {
-        let paths = all_simple_paths::<Vec<_>, _, RandomState>(&g, from, to, 0, None);
-        paths.count()
-    });
-}
-
-#[bench]
-fn simple_paths_multi_bench(bench: &mut Bencher) {
-    static NODE_COUNT: usize = 5;
-    let mut g = Graph::new_undirected();
-    let nodes: Vec<NodeIndex> = (0..NODE_COUNT).map(|_| g.add_node(())).collect();
-
-    // Create a complete undirected graph - every pair of nodes has an edge
-    for i in 0..NODE_COUNT {
-        for j in i + 1..NODE_COUNT {
-            g.add_edge(nodes[i], nodes[j], ());
-        }
-    }
-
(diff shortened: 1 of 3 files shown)


starovoid (approved):
Great!

```
test bipartite_multi_bench       ... bench:     660,654.35 ns/iter (+/- 13,496.46)
test bipartite_single_bench      ... bench:   3,642,804.80 ns/iter (+/- 69,981.22)
```

This difference is impressive, and the new algorithm will be really useful.

lazyhope:
> Great!
> 
> ```
> test bipartite_multi_bench       ... bench:     660,654.35 ns/iter (+/- 13,496.46)
> test bipartite_single_bench      ... bench:   3,642,804.80 ns/iter (+/- 69,981.22)
> ```
> 
> This difference is impressive, and the new algorithm will be really useful.


Thanks a lot for the review and approving it! Really glad to contribute to it.

Issue #799, opened 2025-05-23
Supporting `Acyclic<GraphMap>`


mkeeter:
Right now, it's possible to construct an `Acyclic<GraphMap>`, but we can't do much with it:

```rust
#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Ord, PartialOrd, Hash)]
struct MyIndex(u32);

fn main() {
    let mut g = petgraph::graphmap::DiGraphMap::<MyIndex, ()>::new();
    let mut a = petgraph::acyclic::Acyclic::try_from_graph(g).unwrap();
    a.try_add_edge(MyIndex(1), MyIndex(2), ()).unwrap();
    println!("{a:?}");
}
```
```
error[E0277]: the trait bound `MyIndex: IndexType` is not satisfied
   --> src/main.rs:7:7
    |
7   |     a.try_add_edge(MyIndex(1), MyIndex(2), ()).unwrap();
    |       ^^^^^^^^^^^^ the trait `IndexType` is not implemented for `MyIndex`
    |
    = help: the following other types implement trait `IndexType`:
              NodeIndex<Ix>
              u16
              u32
              u8
              usize
note: the method call chain might not have had the expected associated types
   --> src/main.rs:6:63
    |
6   |     let mut a = petgraph::acyclic::Acyclic::try_from_graph(g).unwrap();
    |                 --------------------------------------------- ^^^^^^^^ `GraphBase::NodeId` is `MyIndex` here
    |                 |
    |                 this expression has type `Result<Acyclic<GraphMap<MyIndex, (), Directed>>, Cycle<MyIndex>>`
note: required by a bound in `Acyclic::<G>::try_add_edge`
   --> /Users/mjk/.cargo/registry/src/index.crates.io-6f17d22bba15001f/petgraph-0.8.1/src/acyclic.rs:183:20
    |
175 |     pub fn try_add_edge(
    |            ------------ required by a bound in this associated function
...
183 |         G::NodeId: IndexType,
    |                    ^^^^^^^^^ required by this bound in `Acyclic::<G>::try_add_edge`
```

What would it take to make this functional?

After looking through the codebase, it seems like the main challenge is the use of `FixedBitSet` in the DFS; those would probably become `IndexSet` in this particular case, but we presumably don't want to sacrifice performance for normal graphs.

I would be willing to work on a PR, given some pointers in the right direction!

[...]

PR #867, opened 2025-08-04
fix!: support `Acyclic<GraphMap<..>>`

mkeeter:
#799 discusses the feasibility of using `Acyclic<GraphMap<..>>`, and suggested implementing `IndexType` on the custom keys.  However, this doesn't actually work with sparse (integer) keys:
 
- The map used for DFS iteration is a `FixedBitSet`
- The bitset is resized based on `node_bound`, which for `GraphMap` returns the number of nodes being stored.
- `impl VisitMap for FixedBitSet` calls `x.index()` to get shift, which is the key's value

If keys are not densely packed, then the key's value can exceed the bitset size; this causes a panic (e.g. https://github.com/petgraph/petgraph/issues/799#issuecomment-3151817135).

This panic can also be **triggered by safe code**, without writing `unsafe impl IndexType...`.  The panic is demonstrated by [a new unit test](https://github.com/petgraph/petgraph/blob/e055e4c9f82970443b3841d4069de7833ae0e68a/src/acyclic.rs#L908-L918), which fails in the same way because `IndexType` is implemented for `usize`.

--------------------------------------------------------------------------------

So, what's to be done?

It turns out that `Visitable` actually defines the kind of map that should be used for keys; however, it's hard-coded to always be a `FixedBitSet` for DFS traversal:

https://github.com/petgraph/petgraph/blob/6b676bf2da17891c91cf6ac5df676d0808178369/src/acyclic.rs#L74-L75

This PR switches to use the appropriate map type, then churns through all of the places that need changing.  It adds new methods to `VisitMap` to match our previous behavior – `ensure_capacity`, `is_empty`, and `clear_all` – so it shouldn't be any less efficient.

There are a few nice properties that we get for free after this change:

- `Acyclic<GraphMap<..>>` now works with non-integral keys, e.g. strings
- Trait bounds become looser: `where G::NodeId: IndexType` is removed in a few places, and downgraded to `core::fmt::Debug` elsewhere

--------------------------------------------------------------------------------

BREAKING CHANGE:

- Add new methods to existing `VisitMap` trait
- Types implementing `VisitMap` must also be `Default`

### DIFF
--- a/src/visit/mod.rs
+++ b/src/visit/mod.rs
@@ -420,7 +420,6 @@ pub trait VisitMap<N>: Default {
     fn ensure_capacity(&mut self, n: usize);
 
     /// Checks that the map is empty
-    #[cfg(debug_assertions)]
     fn is_empty(&self) -> bool;
 }
 
@@ -455,7 +454,6 @@ where
         }
     }
 
-    #[cfg(debug_assertions)]
     fn is_empty(&self) -> bool {
         FixedBitSet::is_clear(self)
     }
@@ -487,7 +485,6 @@ where
         self.clear()
     }
 
-    #[cfg(debug_assertions)]
     fn is_empty(&self) -> bool {
         HashSet::is_empty(self)
     }
@@ -520,7 +517,6 @@ where
         }
     }
 
-    #[cfg(debug_assertions)]
     fn is_empty(&self) -> bool {
         std::collections::HashSet::is_empty(self)
     }


RaoulLuque:
Hey again, thanks for putting together this PR :) The idea and work looks good in general imho :sauropod: 

Lately I haven't had much time, so I'm not sure when I'll be able to actually look into this PR.

On a more general note, breaking changes regarding the trait system are in a bit of a weird state currently (at least from my perspective). There is bound to be a trait rework, which would make changes like this one redundant probably. Thus, I for one, don't feel that motivated to put time into PRs (however worthwhile the change itself may be). Unfortunately, on top of that, the timeline on when the trait rework will land is very vague and the original author/maintainer of the rework has only been saying that they will get back to working on it after this summer (central european summer).

I hope this doesn't come across the wrong way, just wanted to be transparent about that :)

Thanks again for your work. Definitely appreciated :star2:

dimtpap:
Can `remove_node` and `remove_edge` be implemented as well?

dimtpap:
Can `remove_edge` and `remove_node` be implemented as well?

mkeeter:
@RaoulLuque I understand not wanting to review things pending the trait rework.  I want to push back a little, though, because the current code is **actively wrong**, not just unimplemented:

```rust
let mut graph = Acyclic::<GraphMap<usize, (), Directed>>::new();
graph.add_node(1);
graph.add_node(2);
graph.try_add_edge(2, 1, ()).unwrap();
```

This will compile fine, but will panic at runtime with a cryptic error:
```
thread 'main' panicked at /playground/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fixedbitset-0.5.7/src/lib.rs:486:9:
put at index 2 exceeds fixedbitset size 2
```
([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ff6cdb5ee44a301e93108391123dd4be))

Since this PR is fixing an actual issue – not just adding new functionality – I think it may be worthwhile to review, even with the trait rework pending.

Whatever you decide, I appreciate the work that you're doing on `petgraph`, and hope to keep using + contributing to it in the future!

PR #869, opened 2025-08-15
chore: fix `mismatched_lifetime_syntaxes` warnings

mkeeter:
Rust 1.89 includes [a new lint](https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/#mismatched-lifetime-syntaxes-lint)  for mismatched lifetime syntaxes.  This PR fixes those warnings!

PR #871, opened 2025-08-20
fix: Algos don't work on `UndirectedAdaptor` (#870)

### LINKED ISSUE #870: Algorithms like A* don't work with `UndirectedAdaptor`
The `UndirectedAdaptor`'s implementation of `IntoEdges` does not faithfully reproduce the set of edges that an equivalently constructed undirected graph would have. The result is that search and traversal algorithms cannot traverse adapted edges whose underlying directed edge points the "wrong way".

This example unit test has a directed linear graph going from node 0 to node 5 in order.
When adapted as an undirected graph, A* cannot find a path from 5 to 0.

```rust
static LINEAR_EDGES: [(u32, u32); 5] = [(0, 1), (1, 2), (2, 3), (3, 4), (4, 5)];

... 

#[test]
pub fn test_undirected_adapter_can_traverse() {
    // UNGraph
    {
        std::eprintln!("Testing with UNGraph");
        let og_ungraph = Graph::<(), (), crate::Undirected>::from_edges(LINEAR_EDGES);
        assert_eq!(og_ungraph.is_directed(), false);

        let mut nodes = og_ungraph.node_identifiers().collect::<Vec<_>>();
        nodes.sort();

        for i in 0..nodes.len() {
            let n = nodes[i];
            let edges = og_ungraph.edges(n).collect::<Vec<_>>();
            std::eprintln!("edges of {n:?}: {edges:?}");
        }
    }
    {
        std::eprintln!("Testing with Graph");
        let graph = DiGraph::<(), ()>::from_edges(LINEAR_EDGES);

        let mut nodes = graph.node_identifiers().collect::<Vec<_>>();
        nodes.sort();

        for i in 0..nodes.len() {
            let n = nodes[i];
            let edges = graph.edges(n).collect::<Vec<_>>();
            std::eprintln!("edges of {n:?}: {edges:?}");
        }
    }
    {
        std::eprintln!("Testing with adapted Graph");
        let graph = DiGraph::<(), ()>::from_edges(LINEAR_EDGES);
        let ungraph = UndirectedAdaptor(&graph);

        let mut nodes = graph.node_identifiers().collect::<Vec<_>>();
        nodes.sort();

        for i in 0..nodes.len() {
            let n = nodes[i];
            let edges = ungraph.edges(n).collect::<Vec<_>>();
            std::eprintln!("edges of {n:?}: {edges:?}");
        }
    }

    let graph = DiGraph::<(), ()>::from_edges(LINEAR_EDGES);
    let mut nodes = graph.node_identifiers().collect::<Vec<_>>();
    nodes.sort();
    let ungraph = UndirectedAdaptor(&graph);
    let path = astar(&ungraph, nodes[5], |n| n == nodes[0], |_| 1, |_| 0);

    let x = path.unwrap(); // Panics
}
```

I've included additional log output which shows the difference in the `edges()` output between an originally-constructed UnGraph and an adapted Ungraph.

```
Testing with UNGraph
edges of NodeIndex(0): [EdgeReference { index: EdgeIndex(0), node: [NodeIndex(0), NodeIndex(1)], weight: () }]
edges of NodeIndex(1): [EdgeReference { index: EdgeIndex(1), node: [NodeIndex(1), NodeIndex(2)], weight: () }, EdgeReference { index: EdgeIndex(0), node: [NodeIndex(1), NodeIndex(0)], weight: () }]
edges of NodeIndex(2): [EdgeReference { index: EdgeIndex(2), node: [NodeIndex(2), NodeIndex(3)], weight: () }, EdgeReference { index: EdgeIndex(1), node: [NodeIndex(2), NodeIndex(1)], weight: () }]
edges of NodeIndex(3): [EdgeReference { index: EdgeIndex(3), node: [NodeIndex(3), NodeIndex(4)], weight: () }, EdgeReference { index: EdgeIndex(2), node: [NodeIndex(3), NodeIndex(2)], weight: () }]
edges of NodeIndex(4): [EdgeReference { index: EdgeIndex(4), node: [NodeIndex(4), NodeIndex(5)], weight: () }, EdgeReference { index: EdgeIndex(3), node: [NodeIndex(4), NodeIndex(3)], weight: () }]
edges of NodeIndex(5): [EdgeReference { index: EdgeIndex(4), node: [NodeIndex(5), NodeIndex(4)], weight: () }]

Testing with Graph
edges of NodeIndex(0): [EdgeReference { index: EdgeIndex(0), node: [NodeIndex(0), NodeIndex(1)], weight: () }]
edges of NodeIndex(1): [EdgeReference { index: EdgeIndex(1), node: [NodeIndex(1), NodeIndex(2)], weight: () }]
edges of NodeIndex(2): [EdgeReference { index: EdgeIndex(2), node: [NodeIndex(2), NodeIndex(3)], weight: () }]
edges of NodeIndex(3): [EdgeReference { index: EdgeIndex(3), node: [NodeIndex(3), NodeIndex(4)], weight: () }]
edges of NodeIndex(4): [EdgeReference { index: EdgeIndex(4), node: [NodeIndex(4), NodeIndex(5)], weight: () }]
edges of NodeIndex(5): []

Testing with adapted Graph
edges of NodeIndex(0): [EdgeReference { index: EdgeIndex(0), node: [NodeIndex(0), NodeIndex(1)], weight: () }]
edges of NodeIndex(1): [EdgeReference { index: EdgeIndex(0), node: [NodeIndex(0), NodeIndex(1)], weight: () }, EdgeReference { index: EdgeIndex(1), node: [NodeIndex(1), NodeIndex(2)], weight: () }]
edges of NodeIndex(2): [EdgeReference { index: EdgeIndex(1), node: [NodeIndex(1), NodeIndex(2)], weight: () }, EdgeReference { index: EdgeIndex(2), node: [NodeIndex(2), NodeIndex(3)], weight: () }]
edges of NodeIndex(3): [EdgeReference { index: EdgeIndex(2), node: [NodeIndex(2), NodeIndex(3)], weight: () }, EdgeReference { index: EdgeIndex(3), node: [NodeIndex(3), NodeIndex(4)], weight: () }]
edges of NodeIndex(4): [EdgeReference { index: EdgeIndex(3), node: [NodeIndex(3), NodeIndex(4)], weight: () }, EdgeReference { index: EdgeIndex(4), node: [NodeIndex(4), NodeIndex(5)], weight: () }]
edges of NodeIndex(5): [EdgeReference { index: EdgeIndex(4), node: [NodeIndex(4), NodeIndex(5)], weight: () }]
```

The originally-constructed UnGraph has each edge referenced twice in each orientation. This allows A* and other algorithms to work successfully. The adapted UnGraph has each edge referenced twice, but in the same orientation two times. This prevents traversal of adapted edges when the underlying direct edge is oriented in such a way that it would prevent traversal.

I've got a fix ready. I'll make a PR.

CattleProdigy:
Closes #870

The `IntoEdges` implementation for `UnidirectedAdaptor` does not produce the same set of edges references as an equivalently-constructed undirected graph. This prevents search and traversal algorithms from working if the traversal or search would require traversing an adapted edge whose underlying directed edge is "incoming" (which would prevent traversal in the directed case).

The underlying reason is that we currently, correctly concatenate incoming and outgoing edges from the underlying directed graph but they're concatenated with the same orientation which produces a set of redundant `EdgeReferences`-- making the adapted graph equivalent to the underlying directed graph as far as the `IntoEdges` implementation is concerned.

The `IntoEdges` implementation for undirected graphs produces two `EdgeReference`s for each graph edge-- one for each orientation. This is what allows traversal of edges in either direction.

The solution is to make the `UndirectedAdaptor`'s `IntoEdges` behavior match that of an undirected graph by reversing the edges with `Incoming` orientation. That way there will be two `EdgeReferences` for each edge as there are in undirected graphs.


CattleProdigy:
All these lints are slated to be fixed here: https://github.com/petgraph/petgraph/pull/869

src/visit/undirected_adaptor.rs
```diff
@@ -90,6 +196,21 @@ mod tests {
...
+        nodes.sort();
+        let ungraph = UndirectedAdaptor(&graph);
+        let path = astar(&ungraph, nodes[5], |n| n == nodes[0], |_| 1, |_| 0);
+
+        let true_path = (0..=5).rev().map(|i| nodes[i]).collect::<Vec<_>>();
+
+        let (cost, path) = path.unwrap();
+        assert_eq!(cost, 5);
+        assert_eq!(path, true_path);
+    }
```
starovoid, line 213: I don't mind using other algorithms in testing (such `astar`), but there must be at least one independent assert.
I suggest to define a expected `HashSet` of edges for each node and compare it with actual iteration result (like in issue description).

CattleProdigy: fixed here: https://github.com/petgraph/petgraph/pull/871/commits/330b5b1d3924e258b29901b5a450c5c67307c87b

This ended up being surprisingly subtle. if you try to write a test oriented around `ungraph.edges()` or `ungraph.edge_references()`, it's pretty easy to write a test that would pass on unfixed master. That behavior is probably part of why this bug slipped in in the first place.

The test I pushed ensures that there's an edge whose target is one of the neighbors. That's what's important for traversal algorithms to work (see e.g. https://docs.rs/petgraph/latest/src/petgraph/algo/astar.rs.html#129). This test does fail on master and shows that the fix works. 

I also pushed a typo fixup here: https://github.com/petgraph/petgraph/pull/871/commits/26bca337a3f41c0e6c2d9902b78311ff659dbac8

Let me know if this looks good and I'll squash

Cheers

starovoid: Looks good to me, thanks.

starovoid (approved):

PR #872, opened 2025-08-21
refactor: replace `crate::util::enumerate` with `Iterator::enumerate`

cactusdualcore:
This was discussed in my earlier PR #849 and removes `util.rs` and the `enumerate` function in that file.

I didn't have time to check this passes the tests or even compiles, so this is a best effort, purely syntactial transform.

I, however, did use a global text search to make sure that there are no more standalone calls to `enumerate`, only method calls.

Hope this helps ^^


RaoulLuque:
Closed in favor of #881

PR #878, opened 2025-09-21
ci: Fix Hashbrown dependencies

RaoulLuque:
There seems to be an issue with different versions of Hashbrown dependencies:
([PR Job](https://github.com/petgraph/petgraph/actions/runs/17891631430/job/50872932623?pr=790))
```
Run cargo generate-lockfile
  cargo generate-lockfile
  cargo update -p hashbrown --precise 0.15.0
  cargo update -p once_cell --precise 1.20.3
  shell: /usr/bin/bash -e {0}
  env:
    CARGO_TERM_COLOR: always
    CARGO_INCREMENTAL: 0
    SCCACHE_GHA_ENABLED: true
    RUSTC_WRAPPER: sccache
    SCCACHE_PATH: /opt/hostedtoolcache/sccache/0.10.0/x64/sccache
    ACTIONS_RESULTS_URL: https://results-receiver.actions.githubusercontent.com/
    ACTIONS_RUNTIME_TOKEN: ***
    CARGO_HOME: /home/runner/.cargo
    Updating crates.io index
error: There are multiple `hashbrown` packages in your project, and the specification `hashbrown` is ambiguous.
Please re-run this command with `-p <spec>` where `<spec>` is one of the following:
  hashbrown@0.15.5
  hashbrown@0.16.0
Error: Process completed with exit code 101.
```
The version pinning was put in place because Rust 1.64 is actually only compatible with Hashbrown 0.15.0 and not 0.15.5 (the above error stems from the Rust 1.64 Test Job - see link).
This PR now pins both Hashbrown versions (ours should be 0.15.5) to 0.15.0. 

Hashbrown 0.16 should be another dependency pulled in due to indexmap, see this excerpt from `Cargo.lock`:
```
[[package]]
name = "hashbrown"
version = "0.16.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "5419bdc4f6a9207fbeba6d11b604d481addf78ecd10c11ad51e76c2f6482748d"

[[package]]
name = "indexmap"
version = "2.11.4"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "4b0f83760fb341a774ed326568e19f5a863af4a952def8c39f9ab92fd95b88e5"
dependencies = [
 "equivalent",
 "hashbrown 0.16.0",
 "rayon",
]
```

Probably due to some other change (possibly in Cargo?), the default version for the indexmap dependency changed from 0.15.5 to 0.16 which now induces the ambiguity error from above. The specified dependency in indexmap is:
```
[dependencies.hashbrown]
version = ">= 0.15.0, < 0.17.0"
```
So both 0.15.5 and 0.16 would be valid.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,8 @@ jobs:
         if: matrix.rust == '1.64.0'
         run: |
           cargo generate-lockfile
+          cargo update -p hashbrown@0.15.5 --precise 0.15.0
+          cargo update -p hashbrown@0.16.0 --precise 0.15.0
           cargo update -p once_cell --precise 1.20.3
 
       - name: Build with no features


RaoulLuque:
> I like it as a temporary fix, but apparently it's already worth raising the MSRV in the next breaking release.

I am not sure I properly understand your wording. What do you mean exactly? Technically I think it is not necessary to raise the MSRV in the next breaking release, as petgraph works with 1.64. 

On that note, I just tried creating a new project with cargo, adding indexmap as a dependency and that failed, so I'll head there and see whether that's maybe an issue on their side :+1:

PR #782, opened 2025-04-29
feat: Add bidirectional Dijkstra algorithm

ricohageman:
## Brief
Add [bidirectional Dijkstra's](https://en.wikipedia.org/wiki/Bidirectional_search) algorithm for the shortest path between two points.
```rust
pub fn bidirectional_dijkstra<G, F, K>(
    graph: G,
    start: G::NodeId,
    goal: G::NodeId,
    mut edge_cost: F,
) -> Option<K>
where
    G: Visitable + IntoEdgesDirected,
    G::NodeId: Eq + Hash,
    F: FnMut(G::EdgeRef) -> K,
    K: Measure + Copy,
   ```

## About the algorithm and implementation
The time complexity of bi-directional dijkstra is the same as normal dijkstra. However, in certain scenarios (e.g. grids and sparse road networks), it has to explore significantly less nodes to find the shortest path. https://www.ucl.ac.uk/~ucahmto/math/2020/05/30/bidirectional-dijkstra.html is a great resource on the correctness of the algorithm.

## Benchmarks
Under the right circumstances (large & sparse networks), it clearly shows that bidirectional dijkstra is faster than normal dijkstra.
```
test dijkstra_bench_grid                 ... bench:     469,582.30 ns/iter (+/- 12,469.78)
test dijkstra_bench_grid_bidirectional   ... bench:      47,145.54 ns/iter (+/- 1,139.01)
test dijkstra_bench_grid_with_target     ... bench:     118,156.60 ns/iter (+/- 3,168.78)
```

However, the existing benchmark also shows that it's not always beneficial.
```
test dijkstra_bench_random               ... bench:     745,014.06 ns/iter (+/- 22,892.76)
test dijkstra_bench_random_bidirectional ... bench:   1,556,466.68 ns/iter (+/- 77,702.84)
test dijkstra_bench_random_with_target   ... bench:     694,060.94 ns/iter (+/- 18,275.15)
```

On macbook pro M1

PR #879, opened 2025-09-21
refactor: New Dijkstra/General Algorithm API

RaoulLuque:
This PR suggests a revamp to the Dijkstra API. This revamp could (and should) similarly be applied to most of the other algorithms in `petgraph` as well.

A struct to store the state of the Dijkstra computation is introduced. This struct can be constructed and edited using the builder pattern. Furthermore, one can call `.run()` to run the actual algorithm, which returns a `DijkstraOutput`. This structure has three main benefits:

1) Storing the state of the Dijkstra computation: This way, one can reuse previous computations. Using `with_goal()` one can change the goal function after one computation and start a new one with a new goal reusing the results from the previous run of Dijkstra.

2) Having a custom return type: Because the custom return type does not expose its fields, we are more flexible to change internal functionality without having a breaking change.

3) The new API allows for Dijkstra to return Paths as well. We could decide on how we want to implement this specifically and possibly introduce different run functions which only compute distances or paths as well (for performance reasons). But this is only possible due to the new API.

Usage would now look something like this:
```rust
    let mut dijkstra = Dijkstra::new(&graph, start, edge_cost_fn).with_goal(goal_fn);
    let result = dijkstra.run().distance_to_node(&goal_node);

    let mut dijkstra = dijkstra.with_goal(new_goal_fn);
    // Previously computed distances are reused. I.e. if new_goal_node is already explored, no computation is necessary.
    let result_two = dijkstra.run().distance_to_node(&new_goal_node);
```

Of course, this is compatible with the new `with_dynamic_goal` (https://github.com/petgraph/petgraph/pull/855) or rather integrates it natively. This could be applied similarly to `bidirectional_dijkstra` (https://github.com/petgraph/petgraph/pull/782).

If we would accept / agree on these or similar changes before the new version, these would not be considered breaking and we would not have to re-do APIs again right after introducing them (referring to `with_dynamic_goal` and `bidirectional_dijkstra`).

Note that the actual docstrings and tests etc for this API are not done yet, as this is just a draft and suggestion for how it could look like.

Tagging you @starovoid.

### DIFF
--- a/src/algo/dijkstra.rs
+++ b/src/algo/dijkstra.rs
@@ -132,9 +132,9 @@ where
     ///     (e, 1),
     ///     (f, 2),
     /// ].iter().cloned().collect();
-    /// let res = dijkstra::with_dynamic_goal(&graph, b, |&node| node == d || node == f, |_| 1);
-    /// assert_eq!(res.scores, expected_res);
-    /// assert!(res.goal_node == Some(d) || res.goal_node == Some(f));
+    /// let mut dijkstra = dijkstra::Dijkstra::new(&graph, b, |_| 1).with_goal(|&node| node == d || node == f);
+    /// let res = dijkstra.run();
+    /// assert_eq!(res.distances(), &expected_res);
     /// ```
     pub fn run(&mut self) -> DijkstraOutput<'_, G::NodeId, K> {
         let mut visit_next = BinaryHeap::new();


RaoulLuque:
> This PR contains breaking changes to the public Rust API. Please deprecate the old API instead (if possible), or mark the PR with a `!` following the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format to indicate a breaking change.
> cargo-semver-checks summary
> 
> ```
> 
> --- failure function_missing: pub fn removed or renamed ---
> 
> Description:
> A publicly-visible function cannot be imported by its prior path. A `pub use` may have been removed, or the function itself may have been renamed or removed entirely.
>       ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
>      impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.44.0/src/lints/function_missing.ron
> 
> Failed in:
> function petgraph::algo::dijkstra::with_dynamic_goal, previously in file /home/runner/work/petgraph/petgraph/BASELINE_BRANCH/src/algo/dijkstra.rs:165
> 
> --- failure struct_missing: pub struct removed or renamed ---
> 
> Description:
> A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.
>       ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
>      impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.44.0/src/lints/struct_missing.ron
> 
> Failed in:
> struct petgraph::algo::dijkstra::AlgoResult, previously in file /home/runner/work/petgraph/petgraph/BASELINE_BRANCH/src/algo/dijkstra.rs:104
> ```

This PR is not actually a breaking change as the breaking changes that are being referred to are not included in a release yet and only present on GitHub in the master branch.

RaoulLuque:
Moved to 0.10, see https://github.com/petgraph/petgraph/issues/891

PR #886, opened 2025-09-21
crate: Update dependencies

RaoulLuque:
This PR bumps the versions used for dependencies - where possible. Most of them are only dev-dependencies anyways.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ debug = true
 
 [dependencies]
 fixedbitset = { version = "0.5.7", default-features = false }
-indexmap = { version = "2.5.0", default-features = false }
+indexmap = { version = "2.11.0", default-features = false }
 hashbrown = { version = "^0.15.0", default-features = false, features = ["default-hasher", "inline-more"] }
 quickcheck = { optional = true, version = "0.8", default-features = false }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -54,10 +54,9 @@ dot-parser-macros = { version = "0.5.1", optional = true }
 [dev-dependencies]
 bincode = "1.3.3"
 defmac = "0.2.1"
-itertools = { version = "0.12.1", default-features = false }
-odds = { version = "0.4.0" }
-rand = "0.5.5"
-ahash = "0.7.2"
+itertools = { version = "0.14.0", default-features = false }
+rand = "0.6.3"
```
atouchet, line 58: This version of rand is years out of date. Are you deliberately not choosing to update to 0.9 instead?

RaoulLuque: Unfortunately this is indeed deliberate /: It seems like newer versions have some collisions with our (ancient) version of quickcheck. However, bumping that version (quickcheck) is a breaking change (see https://github.com/petgraph/petgraph/pull/788), as we re-export some of the API of quickcheck. Thus, the bigger version bumps will only be possible with a breaking change. However, one should be coming soon, keep an eye out for an announcing issue on that :)

PR #890, opened 2025-09-27
fix: `StableGraph::reverse` breaks free lists

Jhuni0123:
StableGraph is implicitly using the outgoing lists as a free list, but `reverse` is also reversing those.


RaoulLuque (approved):
LGTM, thanks 🦕 !

Issue #862, opened 2025-07-31
Flaky Quickcheck 'test_spfa_undir'


RaoulLuque:
The test 'test_spfa_undir' seems to be flaky as it sometimes passes and sometimes does not. Here is an example of it failing:
https://github.com/petgraph/petgraph/actions/runs/16649666204/job/47118746688

Locally I experienced the following smaller example failing (for the same Quickcheck: test_spfa_undir):
```bash
src/tester.rs:176:28:
[quickcheck] TEST FAILED. Arguments: (Graph { Ty: "Directed", node_count: 9, edge_count: 35, edges: (0, 0), (0, 1), (0, 5), (0, 
8), (1, 0), (1, 3), (1, 6), (1, 7), (2, 3), (2, 4), (2, 5), (2, 6), (3, 2), (3, 4), (3, 6), (4, 0), (4, 1), (4, 2), (4, 4), (5, 0), (5, 2), (5, 3), (5, 4), (5, 7), 
(6, 1), (6, 4), (6, 6), (6, 7), (7, 1), (7, 2), (7, 4), (7, 7), (8, 2), (8, 5), (8, 6), edge weights: {0: 3.893753, 1: -91.912415, 2: -48.74549, 
3: 0.44462585, 4: -80.72861, 5: 42.997223, 6: -19.344208, 7: -76.26109, 8: 82.21585, 9: 38.435287, 10: 59.628128, 11: 
86.483505, 12: -64.717125, 13: 54.60756, 14: -61.73351, 15: -57.99315, 16: 64.286545, 17: -26.240067, 18: -27.858757, 19: 
67.37335, 20: -89.80439, 21: 16.432571, 22: -52.82638, 23: 13.36129, 24: -50.19555, 25: -33.011154, 26: 87.34978, 27: 
1.7368317, 28: 5.6419373, 29: -60.348175, 30: 3.3531647, 31: 23.152992, 32: -87.10389, 33: -86.35218, 34: -25.684097} })
```

starovoid:
Hi! ~~It seems to me that the problem is in comparing the results of `spfa` with `bellman_ford`. Apparently, sometimes there are several different shortest paths in the input graph.~~

`spfa` sometimes gives `Err`, while `bellman_ford` is `Ok`

RaoulLuque:
I can't seem to be able to reproduce the failing tests somehow. Neither the one linked above nor the one in the message above 🤔

Jhuni0123:
[...]

PR #893, opened 2025-09-28
fix: use a queue for SPFA 

Jhuni0123:
fixes https://github.com/petgraph/petgraph/issues/862

The stack-based implementation of SPFA produces false negatives when detecting negative cycles. When benchmarking without a negative cycle, it actually runs faster than Bellman-Ford.

Since the SPFA implementation uses a fixed-size(capacity) queue, no reallocation occurs in `VecDeque`, so it is expected to be sufficiently fast.

**Before**: 46082d680ac1b56ba19da61ec90572307480eb27
```
test bellman_ford_bench        ... bench:       1,813.35 ns/iter (+/- 81.06)
test spfa_bench                ... bench:      21,197.04 ns/iter (+/- 1,145.72)
```
  
**After**: 10051b28c81c0bf5c7f32ba0be3fc5e391b84303
```
test bellman_ford_bench        ... bench:       1,826.56 ns/iter (+/- 155.06)
test spfa_bench                ... bench:       1,371.46 ns/iter (+/- 110.51)
```

### DIFF
--- a/benches/bellman_ford.rs
+++ b/benches/bellman_ford.rs
@@ -24,7 +24,7 @@ fn bellman_ford_bench(bench: &mut Bencher) {
             let n2 = nodes[j];
             let mut distance: f64 = ((i + 3) % 10) as f64;
             if n1 != n2 {
-                distance -= 1.0
+                distance -= 0.5
             }
             g.add_edge(n1, n2, distance);
         }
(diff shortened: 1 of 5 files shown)


RaoulLuque:
Looks great, thank you so much :+1: Would you mind rebasing and moving the test into `tests/spfa.rs`?

starovoid:
Thanks for the PR! I forgot that the benchmark creates a negative cycle, that explains everything :)

Issue #829, opened 2025-06-13
Implement `Graph::try_map` to map a Graph with potential errors.


andriyDev:
Today we have `Graph::map` which allows you to create a new graph after processing it with some closures. This works, unless you can have errors in that conversion process. Then you have to manually build up the graph yourself.

It would be nice if we had something like `Graph::try_map` where its closures return Result and returns `Result<Graph>`.

For a concrete case, I am writing a graph to disk. Some values in the graph are not serializable, so returning an error from the closures allows me to handle the case where the graph cannot be converted to my "serialized form".

starovoid:
Looking at the large number of possible variations of `xxx_map`, the idea arises to develop a number of traits with default implementations.

andriyDev:
What other variants are there? The only one I can think of is Option, but that you could just do with `try_map().ok()`. A separate function isn't really necessary imo.

Personally I'm against traits for these kinds of methods. It adds additional importing to callers and makes it much more difficult to discover.

Issue #891, opened 2025-09-27
Roadmap (WIP)


indietyp:
This issues outlines the public roadmap of the next few versions. The idea here is to have a concrete set of tasks that need to be achieved to allow for consolidation of the `next` branch with the current petgraph codebase with minimal disruptions. Which is a departure of the previous approach.

# 0.8.x

The 0.8 series is going to continue under a dedicated 0.8 branch, we're only going to push fixes to it.

# 0.9.0 ("Port")

This will mark the first version in the transition, during it we will do the following tasks:

- [ ] Create `petgraph-core` crate
- [ ] Port over the core traits from `next` (some modifications likely need to occur surrounding secondary storage, where I am still unsure about the implementation)
- [ ] Port over the `dino` graph implementation (potentially as the default)
- [ ] Implement traits on the existing set of graphs
- [ ] Remove old traits, depend on 0.8, and expose, deprecate and keep those traits implemented (semver trick)
- [ ] Port algorithms over to the new core traits

# 0.10.0 ("Split")

- [ ] Refactor algorithms to be trait based and more generic
   - Visit traits and walkers would also become simple algorithms, instead of being special-cased
   - Where possible keep the old signature and deprecate it
- [ ] Factor out (where it makes sense) implementations into subcrates
   - This might mean (after some evaluation) that the graph implementations get their own crate, with the `petgraph` crate essentially becoming a simple re-export
   - `petgraph-algorithms` becoming an external crate, only depending on `petgraph-core`
   - additional functionality like graph read/write (`petgraph-io`), generators (`petgraph-generators`) and quickcheck/proptest as separate crates that work over the defined traits

# 0.11.0 ("Finish")

- [ ] Extend the set of traits available
- [ ] Remove old deprecated interfaces

With community feedback we're likely going to be changing the timeline, but these changes should enable us to make developing new graph algorithm interfaces a lot more straightforward and increase maintainability.

michaelgoldenn:
I think that it would be quite helpful to create and link issues for each of the points listed so that contributors can get a better idea about the main plans for each point and discuss how we should move forward with them. There are a lot of exciting changes in `next` and I want to help however I can with porting them and reworking algorithms.

RaoulLuque:
> I think that it would be quite helpful to create and link issues for each of the points listed so that contributors can get a better idea about the main plans for each point and discuss how we should move forward with them. There are a lot of exciting changes in `next` and I want to help however I can with porting them and reworking algorithms.

[...]

PR #897, opened 2025-10-05
feat: add try_map for `Graph`

Wuketuke:
as per #829, it would be nice if `Graph` had a fallible `try_map`, returning a `Result<Graph>`
I didnt write a `owning_try_map`, because the graph would be lost on an error, and no `try_filter_map` because its is too niche.

There could be still mappings where only nodes fail but not edges or vice versa, but thats again very niche. Maybe if `map` was generally split into `map_edges` and `map_nodes`, the user could just have that fine grained control, but i think its silly because the code is very simple anyway

Since this pr might be too niche and not get accepted, i didnt bother writing a doc example. In case it does get accepted, i will amend this pr with a doc example

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -1594,6 +1594,14 @@ where
         g
     }
 
+    /// Create a new `Graph` by trying to map node and
+    /// edge weights to new values.
+    ///
+    /// The resulting graph has the same structure and the same
+    /// graph indices as `self`.
+    ///
+    /// if a mapping should fail, an `MapError` gets returned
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Thanks for the PR. Please note that we are currently working on version 0.9 which will drastically change existing functionality :) (See https://github.com/petgraph/petgraph/issues/891)

RaoulLuque:
> Thanks for the PR. Please note that we are currently working on version 0.9 which will drastically change existing functionality :) (See #891)

Because I think the last message might have come across a bit rude: I think the feature addition is great and it's a great addition for 0.9 :)

PR #906, opened 2025-10-10
feat: added Recursive Largest First coloring heuristic

### LINKED ISSUE #903: Add recursive largest first coloring heuristic
The recursive largest first coloring heuristic seems to provide better solutions than DSatur on average, which is currently the only provided coloring heuristic, see [wikipedia](https://en.wikipedia.org/wiki/Recursive_largest_first_algorithm).

michaelgoldenn:
Implemented the Recursive Largest First coloring heuristic and a more generic return type and function for graph coloring. Closes #903.

I'd like to get some feedback before merging, I wasn't sure what to do about `get_colors` or `ColorReturn`, I think documentation on both should be improved but I was not sure how. Or if there is a different general structure I should follow instead, I am willing to discuss and adjust it. If anyone has any good ideas or examples I would be happy to make changes :)

### DIFF
--- a/src/algo/coloring.rs
+++ b/src/algo/coloring.rs
@@ -84,9 +84,9 @@ where
 /// // |               |
 /// // f ----- e------ d
 ///
-/// let return = recursive_largest_first_coloring(&graph);
-/// assert_eq!(return.color_count, 2);
-/// assert_ne!(return.nodes_to_colors[&a], return.nodes_to_colors[&b]);
+/// let return_val = recursive_largest_first_coloring(&graph);
+/// assert_eq!(return_val.color_count, 2);
+/// assert_ne!(return_val.nodes_to_colors[&a], return_val.nodes_to_colors[&b]);
 /// ```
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Hi again :) Glad to see you again with another PR ! As described in #891, the crate is facing large changes which will also impact how algorithms are implemented and tested exactly :)

Thus, my suggestion would be to leave this PR open until 0.9 is near (i.e. the initial transition to the `next` branch is complete) and then you/we can rebase the changes on top of that and thus reuse most of your work :)

michaelgoldenn:
Sounds good! I agree we should wait until 0.9 to pull this, 80% of this was just the tests and documentation, so it will probably be easy to rework for the next changes :)

RaoulLuque:
> Sounds good! I agree we should wait until 0.9 to pull this, 80% of this was just the tests and documentation, so it will probably be easy to rework for the next changes :)

Great :) Forgot to say that in the last comment, but thanks a lot for implementing the algorithm, will be a great addition to the crate :heart_hands:

PR #907, opened 2025-10-12
ci: added doctest to justfile

michaelgoldenn:
This PR simply adds a line to also test doctests when `just test` is run. Currently, running `just test` does not test doctests, so incorrect formatting on those can pass `just test`, but fail during CI/CD. [Example](https://github.com/petgraph/petgraph/actions/runs/18420112265/job/52542782846).

### DIFF
--- a/justfile
+++ b/justfile
@@ -6,9 +6,10 @@ default:
 build:
     cargo build
 
-# Tests with all features enabled
+# Runs tests and doctests with all features enabled
 test:
     cargo test --features all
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Good catch, thanks a lot : ) As described in #891, the crate is currently facing large changes and we thus currently have a stop of merging PRs, however once the initial scaffolding of the changes is done, we will get back to this PR, as it is not really concerned with the crate and its API :sauropod:

PR #911, opened 2025-10-28
feat: Implement `ExactSizeIterator` for `AllEdges`

lenianiva:
The [iterator for `IndexMap`s](https://docs.rs/indexmap/2.5.0/indexmap/map/struct.Iter.html) implement `ExactSizeIterator` and `FusedIterator`, so `AllEdges` can have these traits too.

### DIFF
--- a/src/graphmap.rs
+++ b/src/graphmap.rs
@@ -933,6 +933,9 @@ where
     E: 'a,
     Ty: EdgeType,
 {
+    fn len(&self) -> usize {
+        self.inner.len()
+    }
 }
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Hey there, thanks for the PR : ) Just wanted to let you know that as described in https://github.com/petgraph/petgraph/issues/891, the crate is currently facing large changes and we thus currently have a stop of merging PRs, however once the initial scaffolding of the changes is done, we will get back to this PR, as it should be a straigt-forward addition :sauropod:

Issue #908, opened 2025-10-12
Grouped/batched toposort?


tower120:
Looks like library misses `toposort` version with result, that can be used in parallel.
Like, the one that return sequence of groups, with elements that can be processed simultaneously (if look at DAG as a dependency graph).
Or there is one and I don't see it?

In case if there is actually no such thing, and such algorithm is desirable in lib - I think I made
working modification of Kahn toposort for this. I can make a PR with it.

etylermoss:
@tower120 I'm also looking for the same - could you post a PR or repo?

tower120:
@etylermoss I would love to post a PR with it. But first I want to be sure - it is actually needed in `petgraph`.
It is not just copy/pasting part of code - code need to be properly organized and tested.

PR #920, opened 2025-11-18
feat: Grouped topological sort algorithm

etylermoss:
Implementation for #908.

### DIFF
--- a/src/algo/mod.rs
+++ b/src/algo/mod.rs
@@ -304,6 +304,7 @@ where
     use alloc::collections::VecDeque;
     use ToposortGroupingStrategy::*;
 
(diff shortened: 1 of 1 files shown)


tower120:
As an error you return `Cycle` with "problem" node index (I guess any one of them):
```
    if let Some(cycle_index) = degree.iter().position(|&x| x > 0) {
        Err(Cycle(g.from_index(degree[cycle_index])))
```

How can you be sure that non-zero indegree is not just "ureachable" node?
I mean, unreachable, but not inside cycle. For example in A->B->C->D->E + D->B, E will be unreachable, but outside of cycle, If you're unlucky enough - you'll hit E before B,C,D.

PR #921, opened 2025-11-28
feat: Grouped toposort

tower120:
This is my version of grouped/batched toposort implementation for https://github.com/petgraph/petgraph/issues/908.

I believe this version is faster then https://github.com/petgraph/petgraph/pull/920 - because it does not use `Vec<Vec<_>>` as output, instead it use flat `Vec` with ordered results + flat `Vec` of group "spans".

Additionally all intermediate containers can be reused - resulting in zero-allocation action.

### DIFF
--- a/src/algo/grouped_toposort.rs
+++ b/src/algo/grouped_toposort.rs
@@ -1,4 +1,4 @@
-use std::ptr::slice_from_raw_parts;
+use core::ptr::slice_from_raw_parts;
 use alloc::vec::Vec;
 use crate::Direction;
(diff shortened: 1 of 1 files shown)


tower120:
Looks like all errors are "cosmetic". Your CI `indexmap` errors https://github.com/petgraph/petgraph/actions/runs/19783792277/job/56693261323?pr=921#step:6:52 - is the thing that actually causing the tests to fail. When you fix them, and if you chose this implementation - I'll update it to comply with your formatting, and other things.

PR #927, opened 2025-12-08
feat!: new trait layout

indietyp:
creates the new trait layout and a first implementation

BREAKING CHANGE: new traits that replace the old

### DIFF
--- a/crates/core/src/graph/mod.rs
+++ b/crates/core/src/graph/mod.rs
@@ -91,14 +91,17 @@ impl<G: Graph> Graph for &mut G {
     impl_methods!(G);
 }
 
+#[cfg(feature = "alloc")]
 impl<G: Graph> Graph for alloc::boxed::Box<G> {
     impl_methods!(G);
 }
 
+#[cfg(feature = "alloc")]
 impl<G: Graph> Graph for alloc::rc::Rc<G> {
     impl_methods!(G);
 }
 
+#[cfg(feature = "alloc")]
 impl<G: Graph> Graph for alloc::sync::Arc<G> {
     impl_methods!(G);
 }
--- a/crates/core/src/graph/undirected.rs
+++ b/crates/core/src/graph/undirected.rs
@@ -143,3 +143,134 @@ pub trait UndirectedGraph: Graph {
         self.node_count() == 0
     }
 }
+
+impl<G> UndirectedGraph for &mut G
+where
+    G: UndirectedGraph,
+{
+    #[inline]
+    fn density_hint(&self) -> DensityHint {
+        G::density_hint(self)
+    }
+
+    #[inline]
+    fn cardinality(&self) -> Cardinality {
+        G::cardinality(self)
+    }
+
+    #[inline]
+    fn node_count(&self) -> usize {
+        G::node_count(self)
+    }
+
+    #[inline]
+    fn edge_count(&self) -> usize {
+        G::edge_count(self)
+    }
+
+    #[inline]
+    fn nodes(&self) -> impl Iterator<Item = NodeRef<'_, Self>> {
+        G::nodes(self)
+    }
+
+    #[inline]
+    fn nodes_mut(&mut self) -> impl Iterator<Item = NodeMut<'_, Self>> {
+        G::nodes_mut(self)
+    }
+
+    #[inline]
+    fn isolated_nodes(&self) -> impl Iterator<Item = NodeRef<'_, Self>> {
+        G::isolated_nodes(self)
+    }
+
+    #[inline]
+    fn edges(&self) -> impl Iterator<Item = EdgeRef<'_, Self>> {
+        G::edges(self)
+    }
+
+    #[inline]
+    fn edges_mut(&mut self) -> impl Iterator<Item = EdgeMut<'_, Self>> {
+        G::edges_mut(self)
+    }
+
+    #[inline]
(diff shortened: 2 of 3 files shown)


crates/core/src/graph/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+    edge::{Edge, EdgeMut},
+    id::Id,
+    node::{Node, NodeMut},
+};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Cardinality {
+    pub order: usize,
+    pub size: usize,
+}
```
RaoulLuque, line 22: Probably should already add a brief explanation here, to avoid confusion ^^
```suggestion
pub struct Cardinality {
    /// Number of nodes in the graph
    pub order: usize,
    /// Number of edges in the graph
    pub size: usize,
}
```

crates/core/src/id.rs
```diff
@@ -0,0 +1 @@
+pub trait Id: Copy + PartialEq + Eq + PartialOrd + Ord {}
```
RaoulLuque: Should we maybe have some way to obtain contiguous numberings from `Id`s to be able to store them efficiently in Vectors?

RaoulLuque (commented):
Looks really good, thank you : ) Just two questions 👍

crates/core/src/id.rs
```diff
@@ -1 +1,95 @@
...
+            } => {
+                write!(fmt, "Value {value} is out of range [{min}..={max}]!")
+            }
+        }
+    }
+}
+
+impl core::error::Error for LinearIdTryFromIntError {}
+
+pub trait LinearId:
```
RaoulLuque: I would like to suggest the name `CountableId`, as that would neatly line up with both the mathematical definition of something being countable <=> There exists a bijection to the natural numbers (which is basically what the trait bound of `From` is trying to enforce here). And, in normal language unsigned integers can be described as being counting numbers quite well. 
Of course, `(n,m)` is also countable but probably not natively a `LinearId`, however in that sense `(n,m)` is also linear but still not a `LinearId`.

(Also imho linear is a broadly misunderstood word in general )': See affine function != linear function )

indietyp: Thank you I thought about Countable as an alternative but found that it clashed too much with Countable in the programming language sense. I searched for an alternative that isn't too confusing in the programming language sense. Open for suggestions

RaoulLuque: > Thank you I thought about Countable as an alternative but found that it clashed too much with Countable in the programming language sense. I searched for an alternative that isn't too confusing in the programming language sense. Open for suggestions

I see, which programming language meaning of Countable are you referring to exactly? The only connection I am able to find is the mathematical sense

indietyp: I personally associate countable in the programming sense with think that have a count method, aka containers. So having something that is CountableId evokes to me like it's something that has an upper limit, something that has a .count/.Len method so to say instead of something that's countable in a mathematical sense

RaoulLuque: Ah, I see 🤔 Yeah that makes sense 👍 Still, I'm not sure whether I'd prefer it anyways over Linear. Will think about further options

crates/core/src/graph/undirected.rs
```diff
@@ -0,0 +1,132 @@
...
+    // Edge iteration
+
+    fn edges(&self) -> impl Iterator<Item = EdgeRef<'_, Self>>;
+    fn edges_mut(&mut self) -> impl Iterator<Item = EdgeMut<'_, Self>>;
+
+    // Lookup
+
+    fn node(&self, id: Self::NodeId) -> Option<NodeRef<'_, Self>> {
+        self.nodes().find(|node| node.id == id)
+    }
```
RaoulLuque, line 51: Forgot to ask yesterday, would it make sense to hint `#[inline]` to the compiler for functions like this? I know we expect users to override the default implementations, but otherwise it should most of the time make sense to inline them, right?

indietyp: Yes good point

crates/core/src/id.rs
RaoulLuque: How about `NumericId`, `IndexId` or `IndexableId` ?

EDIT: Another one could be `CountingId` although that probably faces the same problems as `CountableId`. A bit over explicit, but we could also go for `UnsignedIntId`. But that might be too narrow if users implement the trait for their own types which might themselves not be unsigned integers.

indietyp: `IndexId` is agreat!

crates/core/src/id.rs
```diff
@@ -0,0 +1,95 @@
...
+    + TryFrom<usize, Error = IndexIdTryFromIntError>
+{
+    const MIN: Self;
+    const MAX: Self;
+
+    #[inline]
+    #[must_use]
+    fn from_u16(index: u16) -> Self {
+        Self::try_from(index).expect("Cannot create ID: value outside valid range")
+    }
```
RaoulLuque, line 47: This is just a question, not blocking or anything. Why do use `expect` here in the `from_` methods instead of returning the according result?

RaoulLuque (approved):
Looks really good. Thanks so much for your work 🦕

crates/core/src/id.rs
indietyp: because we have the `TryFrom` impl for that! The idea is that you use `from_u16` for truly trivial cases you know, and otherwise `TryFrom`

PR #928, opened 2025-12-08
ci: Split CI into separate crates

RaoulLuque:
This PR changes the CI to be compatible and fitted to the new multi crate layout of petgraph.

In particular, the Testing CI is now structured as follows:
- For every applicable Rust version (stable, beta, nightly) and every crate, runs tests with all and no-default features enabled (ci.yml)
- Runs lints on the entire workspace (ci.yml)
- Checks for no-std compatibility of the workspace (ci.yml)
- Runs every crate's tests with using cargo hack and --feature-powerset. This runs all possible combinations of features. However, due to the high number of features, we only check for combinations of size 1, i.e. all features individually. This may be increased in the future, once there are less features. This action is only run on pushes / merges to the master branch or on PRs with the `S-run-thorough-ci-tests` label (merge-check.yml).
- Runs cargo-semver checks and fails if there are breaking changes but the PR is not declared as breaking. The outputs can now be found in the workflow itself instead of posting them as comments (thus avoiding the previous `pull_request_target` security concerns) (semver-checks.yml)
- Checks for each PR, if the title is formatted correctly. If it contains a "!", it should also contain an explanation of the breaking change in the body of the PR. Errors are again redirected to the workflow output instead of posted as comments (pr-title.yml).
- There is a `miri.yml` which runs miri on the project, however this is currently disabled, as the running time for this has increased drastically. This needs to be investigated to be able to re-enable miri (see #935).

Notably this PR also fixes some tests and their import of `std` dependent objects. As a result, all tests should now pass with `std` and `no_std`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ defmac            = { version = "0.2.1", default-features = false }
 dot-parser        = { version = "0.5.1", default-features = false }
 dot-parser-macros = { version = "0.5.1", default-features = false }
 fixedbitset       = { version = "0.5.7", default-features = false }
+foldhash          = { version = "0.2.0", default-features = false }
 fxhash            = { version = "0.2.1", default-features = false }
 hashbrown         = { version = "^0.15.0", default-features = false }
 indexmap          = { version = "2.5.0", default-features = false }
@@ -56,5 +57,6 @@ pedantic = { level = "warn", priority = -1 }
 self_named_module_files = "warn"
 
 alloc_instead_of_core = "warn"
+min_ident_chars       = "warn"
 std_instead_of_alloc  = "warn"
 std_instead_of_core   = "warn"
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -8,6 +8,12 @@ repository.workspace   = true
 rust-version.workspace = true
 
 [dependencies]
+foldhash  = { workspace = true, optional = true }
+hashbrown = { workspace = true, optional = true }
+
+[features]
+utils = ["dep:foldhash", "dep:hashbrown"]
+alloc = []
 
 [lints]
 workspace = true
--- a/crates/core/src/edge.rs
+++ b/crates/core/src/edge.rs
@@ -1,0 +1,34 @@
+use crate::{graph::Graph, id::Id};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub enum Direction {
+    Incoming,
+    Outgoing,
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Edge<I, D, N> {
+    pub id: I,
+
+    pub source: N,
+    pub target: N,
+
+    pub data: D,
+}
+
+impl<I, D, N> Edge<I, D, N> {
+    pub const fn opposite_endpoint(&self, direction: Direction) -> N
+    where
+        N: Id,
+    {
+        match direction {
+            Direction::Incoming => self.source,
+            Direction::Outgoing => self.target,
+        }
+    }
+}
+
+pub type EdgeRef<'graph, G> =
+    Edge<<G as Graph>::EdgeId, <G as Graph>::EdgeDataRef<'graph>, <G as Graph>::NodeId>;
+pub type EdgeMut<'graph, G> =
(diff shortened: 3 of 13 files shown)


Cargo.toml
```diff
@@ -15,7 +15,7 @@ authors = [
 edition = "2024"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/petgraph/petgraph"
-rust-version = "1.91"
+rust-version = "1.90"
```
starovoid: Why was it decided to downgrade the version?

.github/workflows/ci.yml
```diff
@@ -9,74 +7,70 @@ on:
...
     steps:
-      - uses: actions/checkout@v4
-      - uses: mozilla-actions/sccache-action@v0.0.8
-      - uses: extractions/setup-just@v3
-      - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@stable
+      - name: Checkout code
+        uses: actions/checkout@v5
+      # Remove repo toolchain pin so we can specify toolchain manually
+      - name: Remove toolchain pin
```
indietyp: Do we need to lint in different versions? I would just remove this step and then use the specified nightly version inside of the toolchain. This makes sure that we don't have drift. We could then activate dependabot to update the nightly version occasionally instead.

crates/petgraph/tests/graph6.rs
```diff
@@ -10,6 +6,18 @@ use petgraph::{
     graph6::{FromGraph6, ToGraph6, from_graph6_representation, get_graph6_representation},
 };
 
+#[cfg(all(feature = "std", feature = "graphmap"))]
+type GraphMap<N, E, Ty> = petgraph::graphmap::GraphMap<N, E, Ty>;
```
indietyp, line 10: should the gate rather be a change in default for the generic? so having a default hasher of `fxhash` for no-std and `RandomHasher` for std?

Cargo.toml
RaoulLuque: I don't remember anymore actually but some part of CI was failing due to this - maybe there was a change in the cargo doc json formatting json and then Cargo Denver checks were playing up? Will try to upgrade the version again and see, maybe it's fixed now :)

.github/workflows/ci.yml
starovoid: A good suggestion

crates/petgraph/tests/graph6.rs
RaoulLuque: I guess, but at the same time it already is, as graphmap has a Default Hasher for `feature = "std"`. I guess we could overwrite it here again in case we change the default hasher on `graphmap` for `std`, but that would be a breaking change anyways

RaoulLuque:
Removed the `miri.yml` as it was commented out anyways and being so, it would cause a notification with a failed workflow each time someone triggered a workflow run on a PR. This is the file state as it was, for future reference, once we may want to add it back:
```
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ '*' ]
  workflow_dispatch: { }

name: Miri unsoundness checks

jobs:
  miri:
    name: Miri
    # This job runs on pushes to master or pull_request with the `S-run-thorough-ci-tests` label.
    if: (${{ github.event_name == 'pull_request' && contains(github.event.pull_request.labels.*.name, 'S-run-thorough-ci-tests') }}) || (${{ github.event_name == 'push' }})
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v5
      # Remove repo toolchain pin so we can specify toolchain manually
      - name: Remove toolchain pin
        run: rm -f rust-toolchain.toml rustup-toolchain.toml
      - name: Install nightly toolchain with Miri
        uses: dtolnay/rust-toolchain@nightly
        with:
          components: miri
      - name: Install cargo-nextest
        uses: taiki-e/install-action@v2
        with:
          tool: cargo-nextest

      - name: Build caching
        uses: Swatinem/rust-cache@v2

      - run: just miri
```

Cargo.toml
RaoulLuque: Seems to work now, so no reason to change it actually, thanks :)

.github/workflows/ci.yml
RaoulLuque: Good point, fixed :)

RaoulLuque:
Closed in favor of #942

Issue #198, opened 2018-01-17
Parallel iteration on immutabe node references


sanmai-NL:
Would it be possible to implement to [`IntoParallelIterator`](https://docs.rs/rayon/0.9.0/rayon/iter/trait.IntoParallelIterator.html) for [`stable_graph::NodeReferences`](https://docs.rs/petgraph/0.4.11/petgraph/stable_graph/struct.NodeReferences.html)? 

There was previous discussion of implementing mutable access to the raw `Node` collection in #117; in the end that was put off. Is there a particular stumbling block to implementing the immutable [`raw_nodes()`](https://docs.rs/petgraph/0.4.11/petgraph/graph/struct.Graph.html#method.raw_nodes) for `StableGraph`, as it was for `Graph`?

bluss:
I don't know about `IntoParallelIterator` for that stable graph iterator -- it is not an `ExactSizeIterator`, it doesn't have simple random access split, so it doesn't seem like a good candidate. Basically a collect-to-vec solution or parallel iterator for the index range (covering 0 to upper bound) is the best.

And yes, the path is closed. The design is as follows:

* `Graph` is nakedly confessing it is a `Vec<Node<..>>` and a `Vec<Edge<..>>` and it has some "low level" accessors to the details of the data structure, because it is simple. Yet it does not expose mutable versions of all that so that its invariants can not be harmed.
* `StableGraph` has a more complicated implementation and will not expose any low level details, that would harm its encapsulation.

PR #929, opened 2025-12-09
feat: Add parallel iterators for `Graph` and `StableGraph`

ytanimura:
I have implemented `ParallelIterator`s running on nodes and edges in `Graph` and `StableGraph`.

cf: #198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ All notable changes to this project will be documented in this file.
 
 ### New Features
 
+- Add parallel iterators to `Graph` and `StableGraph` ([#929](https://github.com/petgraph/petgraph/pull/929))
 - Add `into_nodes_edges_iters` to `StableGraph` ([#841](https://github.com/petgraph/petgraph/pull/841))
 - Add methods to reserve & shrink `StableGraph` capacity ([#846](https://github.com/petgraph/petgraph/pull/846))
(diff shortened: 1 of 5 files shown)


ytanimura:
Ah... it looks like an issue with the `rustdoc` version and `semver-check.` https://github.com/petgraph/petgraph/actions/runs/20053126643/job/57512971841?pr=929#step:9:40

RaoulLuque:
Hey, thanks a lot for your PR : ) 

Currently, petgraph is undergoing a lot of changes which are also affecting the graphs offered by petgraph, i.e. like `Graph` and `StableGraph`, see #891 . Thus, at this point it is unclear whether this PR will be able to be merged as is for the future setup of `petgraph`. However, the work will probably be able to be reused and the general idea of adding parallel iterators is very nice. Thus, I'd suggest leaving this PR open until such functionality was added :+1: 

May I ask what your use case for such iterators would be?

PS: CI is currently indeed failing independent of your changes ^^

ytanimura:
Thank you for your reply.

> May I ask what your use case for such iterators would be?

Simply put, it's because heavy computations are often performed on the weights of each node.
In my example, one application is in graphics. This is called instancing. For shapes used repeatedly, like screws or mob characters, it's common practice to store them not as separate shapes, but as a tuple of pointers to the shape and transformation functions. The design of industrial products with complex combinations of parts is represented by a DAG where shapes are leaves and transformation matrices are edges. While using graph structures optimizes data organization, heavy processing like tessellation is still required for terminal nodes.

PR #934, opened 2025-12-14
feat: try reserve for the graph

TheAwiteb:
Add `Graph::try_reserve_nodes`, `Graph::try_reserve_edges`, `Graph::try_reserve_exact_nodes` and `Graph::try_reserve_exact_edges`

### DIFF
--- a/crates/petgraph/src/graph_impl/mod.rs
+++ b/crates/petgraph/src/graph_impl/mod.rs
@@ -8,6 +8,7 @@ use core::{
     ops::{Index, IndexMut, Range},
     slice,
 };
+use std::collections::TryReserveError;
 
 use fixedbitset::FixedBitSet;
 
@@ -1363,7 +1364,7 @@ where
     /// Reserves capacity for at least `additional` more nodes to be inserted in
     /// the graph. Graph may reserve more space to avoid frequent reallocations.
     ///
-    /// **Panics** if the new capacity overflows `usize`.
+    /// **Panics** if the new capacity overflows `isize`.
     #[track_caller]
     pub fn reserve_nodes(&mut self, additional: usize) {
         self.nodes.reserve(additional);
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Hi, thanks for your contribution and also for you other PR :) !

Just wanted to let you know that as sketched out in Issue #891 , we are currently restructuring the crate a lot and it is not certain yet whether the `Graph`-graph, as it exists currently will be present in the future versions of the crate / whether it might get deprecated soon.

It's also not sure yet whether we'll merge feature additions like this or not for parts of the crate that are potentially being dropped.

TheAwiteb:
Hi @RaoulLuque, thanks for your fast response.

I know that there is a big refactor for version 0.9.0

But I think v0.8 needs those functions, because the existing one `reserve` will panic if it can't allocate the requested memory, so, we need a safe alternative that we can handle the error instead of panicking.

Will you guys release v0.8.4? or the next release will be v0.9?

If there is v0.8.4, I think those functions should be in it.

Thanks.

TheAwiteb:
> The 0.8 series is going to continue under a dedicated 0.8 branch, we're only going to push fixes to it. The 0.8.3. release will be the last planned release for 0.8. The versions >= 0.9 will then be based on 0.8.3..

Looks like there is no v0.8.4 :(

So, what is the state for this change? Will you ping me to port it in v0.9.0 later?

RaoulLuque:
We are not sure yet what the state of 0.8.4 is, there'll be probably be such a version at some point, but that would of course bind some of our resources, and the priorities currently lie on focusing on 0.9.

RaoulLuque:
> > The 0.8 series is going to continue under a dedicated 0.8 branch, we're only going to push fixes to it. The 0.8.3. release will be the last planned release for 0.8. The versions >= 0.9 will then be based on 0.8.3..
> 
> Looks like there is no v0.8.4 :(
> 
> So, what is the state for this change? Will you ping me to port it in v0.9.0 later?

We can definitely ping you to try to port it to `0.9` if that's applicable : ) However, since the `Graph` type as it currently is will slowly be deprecated, it's unclear whether such a method will be applicable in the future :+1:

PR #940, opened 2026-01-02
feat(algo): Lexicographical topological sort

### LINKED ISSUE #939: Lexicographical Topological Sort
Is there any plan to implement an algorithm like https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.lexicographical_topological_sort.html which is a stable version of topological sort?

lenianiva:
This PR adds a lexicographic topological sort based on the implementation in https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.lexicographical_topological_sort.html

Resolves #939

One problem with this is we cannot directly reuse the dfs space. I added a new `LfsSpace` which could be used for other operations where node traversal has to follow some particular key order in the case of ambiguity. `LfsSpace` space is slightly different from `DfsSpace`, in that it stores a remaining in-degree map and a heap, but not `visited`. Both of these data structures can have capacity drastically smaller than the graph itself. Although their sizes must sum to the graph node count.

- [x] Add `lexicographic_toposort`
- [x] Add `LfsSpace::<NodeId, Key>` for reusable spaces
- [x] Sorting by custom keys in `lexicographic_toposort_by_key`

### DIFF
--- a/crates/petgraph/src/algo/mod.rs
+++ b/crates/petgraph/src/algo/mod.rs
@@ -36,8 +36,7 @@ pub mod spfa;
 pub mod steiner_tree;
 pub mod tred;
 
-use hashbrown::HashMap;
-use alloc::{collections::{BinaryHeap}, vec, vec::Vec};
+use alloc::{collections::BinaryHeap, vec, vec::Vec};
 
 pub use astar::astar;
 pub use bellman_ford::{bellman_ford, find_negative_cycle};
@@ -46,6 +45,7 @@ pub use coloring::dsatur_coloring;
 pub use dijkstra::{bidirectional_dijkstra, dijkstra};
 pub use feedback_arc_set::greedy_feedback_arc_set;
 pub use floyd_warshall::floyd_warshall;
+use hashbrown::HashMap;
 pub use isomorphism::{
     is_isomorphic, is_isomorphic_matching, is_isomorphic_subgraph, is_isomorphic_subgraph_matching,
     subgraph_isomorphisms_iter,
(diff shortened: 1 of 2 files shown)


schneiderfelipe:
Is there an update on this? This would be a great addition to petgraph!

RaoulLuque:
Since this Issue is blocked by https://github.com/petgraph/petgraph/issues/891, you'd have to look there for updates :) (And with regard to that, the most recent updates can be found on the petgraph discord ^^)
Until then I'd suggest you depend on the branch by @lenianiva in your project :+1:

lenianiva:
> Is there an update on this? This would be a great addition to petgraph!

What features do you urgently need? I can code them in but I don't know what change will #891 bring.

RaoulLuque:
> I can code them in but I don't know what change will #891 bring.

I think I can somewhat sketch what change #891 will bring for this PR:
- The new trait bounds will have to be used for the generic graph argument (see https://github.com/petgraph/petgraph/tree/master/crates/core/src/graph `/directed.rs` and `undirected.rs` respectively).
- Probably there are going to be doc changes as we are going to try to unify the algo docs once more.
- Probably the return type / call structure of the algorithm might change. That is possibly have a custom return type instead of `Vec` and have a struct with which one can call the algorithm (while also offering the current one-function API which just uses the struct algorithm API under-the-hood.
In particular, no major changes to the algorithm logic should be necessary  ^^

Hope this helps in your considerations for expanding the algorithm : )

crates/petgraph/src/algo/mod.rs
```diff
@@ -262,6 +263,52 @@ where
...
+        let deg = g.neighbors_directed(n, Direction::Incoming).count();
+        if deg == 0 {
+            None
+        } else {
+            Some((n, deg))
+        }
+    })
+        .collect();
+    // collect all 0 in-degree nodes
+    let mut heap: BinaryHeap<G::NodeId> = g.node_identifiers().filter(|n| g.neighbors_directed(*n, Direction::Incoming).next().is_none())
```
nimral: BinaryHeap is a max heap, you might want to use Reverse:
https://doc.rust-lang.org/std/collections/struct.BinaryHeap.html#min-heap

crates/petgraph/src/algo/mod.rs
```diff
@@ -262,6 +263,52 @@ where
...
 
+/// Perform a lexicographic topological sort of a directed graph.
+///
+/// `toposort` returns `Err` on graphs with cycles.
+/// To handle graphs with cycles, use the one of scc algorithms or
+/// [`DfsPostOrder`](struct@crate::visit::DfsPostOrder)
+///   instead of this function.
+pub fn lexicographic_toposort<G>(
+    g: G,
+) -> Result<Vec<G::NodeId>, Cycle<G::NodeId>>
```
nimral, line 277: For my use case I modified this to accept a mapping NodeId → Key (similarly to https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.dag.lexicographical_topological_sort.html), but maybe I'm just not familiar enough with petgraph

lenianiva:
> > I can code them in but I don't know what change will #891 bring.
> 
> I think I can somewhat sketch what change #891 will bring for this PR:
> 
>     * The new trait bounds will have to be used for the generic graph argument (see https://github.com/petgraph/petgraph/tree/master/crates/core/src/graph `/directed.rs` and `undirected.rs` respectively).
> 
>     * Probably there are going to be doc changes as we are going to try to unify the algo docs once more.
> 
>     * Probably the return type / call structure of the algorithm might change. That is possibly have a custom return type instead of `Vec` and have a struct with which one can call the algorithm (while also offering the current one-function API which just uses the struct algorithm API under-the-hood.
>       In particular, no major changes to the algorithm logic should be necessary  ^^
> 
> 
> Hope this helps in your considerations for expanding the algorithm : )

Will the return type become an iterator like the implementation in networkx? It should be an easy change.

RaoulLuque:
> Will the return type become an iterator like the implementation in networkx? It should be an easy change.

Where possible that would preferable, yes :)

lenianiva:
> > Will the return type become an iterator like the implementation in networkx? It should be an easy change.
> 
> Where possible that would preferable, yes :)

How would the new interface interact with say https://github.com/petgraph/petgraph/blob/ed714652ab4576104e506c096b6ed9f5128613a7/crates/petgraph/src/visit/traversal.rs#L309 ?

PR #941, opened 2026-01-08
feat: Add Display Implementation for NodeIndex<Ix> and EdgeIndex<Ix> whenever Ix: Display.

RaoulLuque:
This PR implements the `Display` trait for `NodeIndex<Ix>` and `EdgeIndex<Ix>`, if `Ix: Display`.

During the implementation of the new Traits on the "old" graph types as described in https://github.com/petgraph/petgraph/issues/891, we noticed that these Impls would be useful.


RaoulLuque:
CI seems to be failing unrelated to this. I guess that's another TODO

PR #946, opened 2026-01-21
feat: Expose mutable Edge Weight iterators for MatrixGraph

RaoulLuque:
This PR adds `all_nodes`, `all_nodes_mut` and `all_edges`, `all_edges_mut` Methods to `MatrixGraph` (and adds corresponding tests).

During the implementation of the new Traits on the "old" graph types as described in https://github.com/petgraph/petgraph/issues/891, we noticed  that such methods would be useful, if not, necessary to implement the new traits on `MatrixGraph`.

This PR depends on the PR #941 being merged.

### DIFF
--- a/src/graph_impl/mod.rs
+++ b/src/graph_impl/mod.rs
@@ -125,6 +125,12 @@ impl<Ix: IndexType> NodeIndex<Ix> {
     }
 }
 
+impl<Ix: fmt::Display> fmt::Display for NodeIndex<Ix> {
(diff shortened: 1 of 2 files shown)


RaoulLuque:
The last commit 5bb3548e89ca06c8c5350cd2d769c016cb1f9d34 is just a drive-by as clippy seems to have failed unrelated to this. Not sure if it's new clippy lints or just because I touched the file in this PR :thinking:

PR #951, opened 2026-01-29
fix: correct edge_count for undirected graphs created using csr::from_sorted_edges

viniciusdutra314:
### Summary
This PR fixes an inconsistency in `Csr::from_sorted_edges` where `edge_count` was incorrectly calculated for `Undirected` graphs.

Previously, `from_sorted_edges` incremented `edge_count` for every entry in the input slice. For `Undirected` graphs, which requires deduplication of reciprocal entries `(u, v)` and `(v, u)`, this resulted in an `edge_count` that was roughly (not considering self loops) the double of logical edges. This behavior contradicted the `add_edge` implementation and the mathematical definition of an undirected graph.

### The Fix
A self_loop counter was added, the number of edges is simply half of the non self_loops edges (thus considering (u,v) as equal to (v,u) plus the self_loops
`edge_count = (total_entries-self_loops)/2 + self_loops`

This correctly deduplicates reciprocal pairs while preserving single self-loop entries.

### Unit test updates
* Updated `csr_from_undirected` test assertion to expect the correct logical edge count (5 instead of 8).
* Added small test called `csr_undirected_deduplication_check` to check for the simple case of  [(0,1),(1,0),(2,2)], which should yield 2

### DIFF
--- a/src/csr.rs
+++ b/src/csr.rs
@@ -207,6 +207,7 @@ where
         };
         let mut self_ = Self::with_nodes(max_node_id + 1);
         let mut iter = edges.iter().cloned().peekable();
+        let mut self_loops = 0;
         {
             let mut rows = self_.row.iter_mut();
 
@@ -245,6 +246,9 @@ where
                                 first_error: (n.index(), m.index()),
                             });
                         }
+                        if !Ty::is_directed() && n.index() == m.index() {
+                            self_loops += 1;
+                        }
                         last_target = Some(m);
                         self_.column.push(m);
                         self_.edges.push(weight);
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Hi there, thanks for the PR and the good catch ! Since the current graph Implementations are going to be phased out, we don't accept PRs for them on the main branch (see https://github.com/petgraph/petgraph/issues/891 for more info). So please, redirect your PR to the `petgraph:0.8` branch :sauropod:

viniciusdutra314:
Thanks for the feedback @RaoulLuque ! I've cherry-picked my changes into the 0.8 branch to resolve the merge conflicts. It should be ready for review now.

RaoulLuque:
I'm not sure why the checks are not running nor am I seeing a button to approve the workflows to run. I tried addressing this with the other PR I linked this one to, but not sure if that actually fixes it. GitHub does not seem to handle a non-default branch to develop on very well

PR #961, opened 2026-02-13
fix: fix the impl error in the page_rank algorithm

Avandar2024:
fix error and impove the performence


Avandar2024:
The test:
```rust
fn expected_ranks() -> Vec<f32> {
    vec![
        0.029228685,
        0.38176042,
        0.3410649,
        0.014170233,
        0.035662483,
        0.077429585,
        0.035662483,
        0.014170233,
        0.014170233,
        0.014170233,
        0.014170233,
        0.014170233,
        0.014170233,
    ]
}
```
encode a wrong result rather than compare with other impl

PR #964, opened 2026-02-18
feat!: Port and rewrite `MatrixGraph`

RaoulLuque:
This PR ports and rewrites the existing [`MatrixGraph`](https://docs.rs/petgraph/latest/petgraph/matrix_graph/index.html) to the new traits.

Some cleaning up was done, which resulted in a lot of code being removed, or moved to the `DirectedGraph` and `UndirectedGraph` trait impls. Separate modules for the `Directed` and `Undirected` impls and their helpers were created.

In particular, I used a new crate for `MatrixGraph` as this allows to import and use `MatrixGraph` in testing for `petgraph-algorithms` or other crates without creating a circular dependency (like it would be the case if `MatrixGraph` was part of the main `petgraph` crate).

Some of the existing tests were adapted, while most were deleted, as they are already contained in the `test_directed_graph!` and `test_undirected_graph!` macros (see https://github.com/petgraph/petgraph/pull/947 ).

Using the aforementioned tests, several bugs of the ported `MatrixGraph` implementation 

Breaking Change:
This PR introduces a new crate in the workspace: `matrix_graph` which contains the implementation of the MatrixGraph crate. This is not technically a breaking change.


RaoulLuque:
Semver checks seem to fail as a new crate is created and thus no comparison to the old one can be made.

RaoulLuque:
Converted to a draft since https://github.com/petgraph/petgraph/pull/1017 should be incorporated here as well.

PR #967, opened 2026-02-25
chore: Bump hashbrown to ^0.16

SuchAFuriousDeath:
Together with https://github.com/gfx-rs/wgpu/pull/9078, this should deduplicate hashbrown for wgpu and eventually for Ruffle.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ dot-parser-macros = { version = "0.5.1", default-features = false }
 fixedbitset       = { version = "0.5.7", default-features = false }
 foldhash          = { version = "0.2.0", default-features = false }
 fxhash            = { version = "0.2.1", default-features = false }
-hashbrown         = { version = "^0.15.0", default-features = false }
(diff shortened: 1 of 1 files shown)


RaoulLuque:
Thanks for the PR :) I'm wondering whether this should be considered a breaking change as `hashbrown` is included in the public API (e.g. [here](https://github.com/petgraph/petgraph/blob/ba97705952f2e69227a69ecbb2b5341069c64ebe/crates/petgraph/src/algo/dijkstra.rs#L108-L113) )

Also I want to note that the branch you are currently targeting is going to be `0.9`. If you would like this change to be included in `0.8.4.` (if it is not a breaking change), please change the target branch to be `0.8`

SuchAFuriousDeath:
This probably is a breaking change because of https://github.com/rust-lang/hashbrown/pull/643. I'm fine with 0.9.

SuchAFuriousDeath:
@RaoulLuque

PR #983, opened 2026-03-31
feat: Add unsafe node weight getter methods

mtreinish:
This commit adds new methods to the Graph and StableGraph types to get a node weight without bounds or validity checking. There are several use cases where the user can know with confidence that all the indices used when accessing the weights from the graph are valid and correct and having the overhead of bounds checking or whether the node has been removed from a stable graph. Having an unsafe method where the user asserts in their program the indices calling these methods with are valid and does it explicitly in an unsafe block is a valid can help address these use cases.

### DIFF
--- a/crates/petgraph/src/graph_impl/mod.rs
+++ b/crates/petgraph/src/graph_impl/mod.rs
@@ -614,6 +614,34 @@ where
         self.nodes.get_mut(a.index()).map(|n| &mut n.weight)
     }
 
+    /// Return a reference to a node without doing any bounds checking
+    ///
+    /// For a safe alternative see `node_weight`.
+    ///
+    /// # Safety
+    ///
+    /// Calling this method with an out-of-bounds index or an index that has
+    /// been removed from the graph is undefined behavior even if the
+    /// resulting reference is not used. You can think of this like calling
+    /// `.get(a).unwrap_unchecked()`
+    pub unsafe fn node_weight_unchecked(&self, a: NodeIndex<Ix>) -> &N {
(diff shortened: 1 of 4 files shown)


RaoulLuque:
Hi there,

that seems like a great addition, thanks ! I agree that this seem like a value-add to the existing functionality and I see the use case :+1:

Please note, however, the following issue which describes the current focus of the project: https://github.com/petgraph/petgraph/issues/891. In particular note the most recent comment on the issue. 
Probably `StableGraph` and `Graph` will be deprecated in their current form, however we could still consider merging this into the `0.8.` branch :thinking: Also, we could consider adding such methods to the new `DirectedGraph` and `UndirectedGraph` traits :thinking: 

I'll add this to the 0.9 milestone for now to not forget it. That might mean either of the above two or both, not sure yet ^^ Feel free to share your opinion as well : )

mtreinish:
Sorry I missed your comment your comment, sure I'm fine with whatever makes sense. I opened this to investigate some near term performance use case we have with very large graphs, so personally I'd prefer having access to it sooner rather than later. But I understand it might not fit into the next release plan so I'm happy with however it fits in.

I haven't looked into the replacement data structures described in the issue in too much depth but I think these methods will probably fits on the replacement structure too (assuming they're backed with Veca still). I did think about adding these to an trait too (although I haven't looked at the new traits described in the issue), but it didn't feel like the best fit across all the graph types that implement weights for nodes and edges. For example, something like graph map because there is no unsafe getter for a hash map. I also didn't want to force downstream graph impls to do this if it didn't fit.

PR #984, opened 2026-04-02
perf:  Kruskal's iterator creation 2x faster via O(E) bottom-up heapify

viniciusdutra314:
## Overview
This PR optimizes the initialization of the `BinaryHeap` in the `min_spanning_tree` (Kruskal's algorithm) implementation. By leveraging Rust's `FromIterator`  implementation for `BinaryHeap`, we transition from an $O(E \log E)$ construction to a linear $O(E)$ "bottom-up" heapify.

## Technical Details
Previously, the heap was populated by iterating over edge references and calling `.push()` for each element. This resulted in a total setup cost of $O(E \log E)$ as the heap re-balanced itself upon every insertion.

This change uses `BinaryHeap::from_iter()`. As documented in the [Rust Standard Library](https://doc.rust-lang.org/beta/std/collections/struct.BinaryHeap.html#impl-From%3CVec%3CT,+A%3E%3E-for-BinaryHeap%3CT,+A%3E), this organizes the elements in-place in $O(E)$ time.

## Benchmarks
To verify the impact, I added a benchmark specifically targeting the iterator creation (where the heapification occurs). For a **100x100 lattice graph**, the new implementation is approximately **2x faster** on my computer.
Screenshot_20260402_120242

## Changes
- Refactored `min_spanning_tree` to use ``BinaryHeap::from()`.
- Added a benchmark case for MST iterator initialization.

### DIFF
--- a/benches/min_spanning_tree.rs
+++ b/benches/min_spanning_tree.rs
@@ -15,6 +15,14 @@ use petgraph::{
     Graph, Undirected,
 };
 
+use crate::common::build_2d_grid;
+
+#[bench]
+fn min_spanning_tree_kruskal_create_iter(bench: &mut Bencher) {
+    let g = build_2d_grid(100, 100);
+    bench.iter(|| std::hint::black_box(min_spanning_tree(&g)));
+}
+
 #[bench]
 fn min_spanning_tree_kruskal_praust_undir_bench(bench: &mut Bencher) {
     let a = ungraph().praust_a();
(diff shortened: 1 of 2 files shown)


RaoulLuque:
Hi there, this is great, thanks so much ! :star2: 

Note the current focus of the maintainers of this repo described in this issue https://github.com/petgraph/petgraph/issues/891 . Nevertheless what the issue says I think we could merge this, as it's a relatively straightforward optimization with a great improvement. Makes me think if there's any other spots where there's these kind of easy wins to make :thinking: 

Would you maybe consider opening a PR to the main branch as well, or at least an issue to make sure we implement this change in main as well? : )

PR #991, opened 2026-05-05
chore: Bump hashbrown to 0.17

cuviper:
This follows the 0.16 upgrade from #967, which hasn't been released, so we can jump to the latest 0.17 instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ dot-parser-macros = { version = "0.5.1", default-features = false }
 fixedbitset       = { version = "0.5.7", default-features = false }
 foldhash          = { version = "0.2.0", default-features = false }
 fxhash            = { version = "0.2.1", default-features = false }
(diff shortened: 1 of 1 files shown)


konstin:
This change is blocking deduplicating the hashbrown dependency for us, would it be possible to have a release with this update?

RaoulLuque:
Hi there, I am currently relatively busy and I'm not sure there's anyone else around currently working on petgraph. Also I'll be on vacation in about 2 weeks for 6 weeks.

Since this is a breaking change, this can't be done in the 0.8 branch I'm afraid (if it could, then we could push a release before I go on vacation, no problem I'd say). However, since the 0.9 branch (i.e. main branch) is not really ready to be released, I'm unsure whether that's feasible right now (at least it's not for me, and I'm unsure whether there's someone else willing to take on the work).

Sorry to say this, but thought I didn't want to leave this question unanswered :o

Issue #960, opened 2026-02-12
Fix the error in the implementation of page_rank algorithm


Avandar2024:
In the implementation of the page_rank algorithm,  the original implementation incorrectly used if/else logic to choose between link propagation, dangling nodes, and teleportation, whereas these must be summed together for every node to satisfy the PageRank formula and prevent rank leakage. 
Below is my fix, optimize the code style too.I have tested it with networkx in python.
### code
the original: 
```rust
pub fn page_rank<G, D>(graph: G, damping_factor: D, nb_iter: usize) -> Vec<D>
where
    G: NodeCount + IntoEdges + NodeIndexable,
    D: UnitMeasure + Copy,
{
    let node_count = graph.node_count();
    if node_count == 0 {
        return vec![];
    }
    assert!(
        D::zero() <= damping_factor && damping_factor <= D::one(),
        "Damping factor should be between 0 et 1."
    );
    let nb = D::from_usize(node_count);
    let mut ranks = vec![D::one() / nb; node_count];
    let nodeix = |i| graph.from_index(i);
    let out_degrees: Vec<D> = (0..node_count)
        .map(|i| graph.edges(nodeix(i)).map(|_| D::one()).sum::<D>())
        .collect();

    for _ in 0..nb_iter {
        let pi = (0..node_count)
            .enumerate()
            .map(|(v, _)| {
                ranks
                    .iter()
                    .enumerate()
                    .map(|(w, r)| {
                        let mut w_out_edges = graph.edges(nodeix(w));
                        if w_out_edges.any(|e| e.target() == nodeix(v)) {
                            damping_factor * *r / out_degrees[w]
                        } else if out_degrees[w] == D::zero() {
                            damping_factor * *r / nb // stochastic matrix condition
                        } else {
                            (D::one() - damping_factor) * *r / nb // random jumps
                        }
                    })
                    .sum::<D>()
            })
            .collect::<Vec<D>>();
        let sum = pi.iter().copied().sum::<D>();
        ranks = pi.iter().map(|r| *r / sum).collect::<Vec<D>>();
    }
    ranks
}
```
my fix: 
```rust
pub fn page_rank<G, D>(graph: G, damping_factor: D, nb_iter: usize) -> Vec<D>
where
    G: NodeCount + IntoEdges + NodeIndexable,
    D: UnitMeasure + Copy,
{
    let node_count = graph.node_count();
    if node_count == 0 {
        return vec![];
    }
    assert!(
[...]

PR #959, opened 2026-02-11
perf: optimize PageRank implementation to reduce complexity and allocations

Avandar2024:
### Description

This PR improves the PageRank algorithm implementation by optimizing its computational complexity and aligning the code with idiomatic Rust patterns.

### Changes

- **Complexity Optimization**: Transitioned from an $O(N^2)$ node-to-node check to an $O(E)$ edge-based update by pre-calculating incoming edges or iterating over the graph's adjacency list.
- **Memory Efficiency**: Eliminated redundant `Vec` allocations inside the iteration loop. Used a buffer-swapping strategy (`std::mem::swap`) between `ranks` and `next_ranks`.
- **Idiomatic Rust**: 
  - Replaced manual indexing with safer and faster iterators (`iter()`, `iter_mut()`, `zip`).
  - Extracted constant teleportation terms $(1 - d) / N$ outside the inner loop.
- **Correctness**: Maintained the stochastic matrix condition and proper handling of dangling nodes (nodes with no out-degree).

### Performance Impact

For a graph with $V$ nodes and $E$ edges(according to the bench mark in benches/page_rank.rs):
- **Old**: test page_rank_bench ... bench:   3,933,410.45 ns/iter (+/- 87,098.60)
- **New**: test page_rank_bench ... bench:      14,219.46 ns/iter (+/- 778.38)
- Significant reduction in heap allocations and cache misses.

### Checklist

- [x] PR title follows Conventional Commits.
- [x] Tests updated/added to verify the new implementation.
- [x] Code formatted with `cargo fmt`.
- [x] No breaking changes (API remains the same).

PR #992, opened 2026-05-17
perf: compute PageRank in O(|V|+|E|) per iteration instead of O(|V|^2|E|)

RyanJamesStewart:
## What

`page_rank` computed each iteration by summing, for every node `v`, a term over **every** node `w`, where each `(v, w)` pair probed `w`'s out-edges to test whether `w -> v` (`graph.edges(w).any(|e| e.target() == v)`). That is `O(|V|^2 |E|)` per iteration. The function's own doc comment already advertised `O(n|V|^2|E|)`.

This rewrites the iteration to a single sparse pass over the edge set, `O(|V| + |E|)`, with identical semantics, and corrects the documented complexity to `O(n(|V| + |E|))`.

## Why it is equivalent

The per-`(v, w)` sum splits exactly into two per-iteration scalars plus a term that is non-trivial only for the real in-neighbours of `v`:

```
pi[v] = A + T + sum over distinct w with an edge w->v of
            ( d*r[w]/outdeg[w] - (1-d)*r[w]/nb )

A = sum over dangling w (outdeg 0) of  d*r[w]/nb       (rank sink)
T = sum over non-dangling w of        (1-d)*r[w]/nb    (random jump)
```

`A` and `T` are computed once per iteration in `O(|V|)`; the remaining term is applied by one pass over the edges in `O(|E|)`. Parallel edges `w -> v` collapse to a single contribution, matching the original boolean `any` semantics (a per-source dedup stamp). Dangling nodes and self-loops behave identically to before.

## Numerical note

This reorders the floating-point reduction, so output matches the previous implementation to within accumulation tolerance, not bit-for-bit:

- Verified equivalent to a max absolute delta of `6.7e-16` (f64) against the previous implementation over 3000 random graphs that deliberately include dangling nodes, self-loops, and parallel edges, across damping factors `0.0..=1.0` and `1..25` iterations.
- The `page_rank` doctest's f32 expected values shift by at most 2 ulps and are updated.
- `tests/page_rank.rs::test_page_rank` asserted exact f32 equality (`assert_eq!`) on a 100-iteration reduction, which is reduction-order dependent. It is changed to the same tolerance comparison that `test_par_page_rank` in the same file already uses against the same reference vector. The reference values themselves are unchanged.

## Benchmark

Serial `page_rank`, 30 iterations, random graph with 800 nodes / 6400 edges: previous implementation `~0.140 s`, this implementation `~0.00046 s` (about 300x), checksum difference `6e-16`. The factor grows with `|V|`.

## Scope

`parallel_page_rank` has the same `O(|V|^2 |E|)` characteristic but a correct sparse formulation there involves a parallel scatter that is a distinct piece of work. This PR keeps the change to the serial path; the parallel variant is unmodified and can follow separately.

### DIFF
--- a/crates/petgraph/src/algo/page_rank.rs
+++ b/crates/petgraph/src/algo/page_rank.rs
@@ -23,7 +23,7 @@ use crate::visit::{EdgeRef, IntoEdges, NodeCount, NodeIndexable};
 /// Otherwise, it panics.
 ///
 /// # Complexity
-/// * Time complexity: **O(n|V|²|E|)**.
+/// * Time complexity: **O(n(|V| + |E|))**.
 /// * Auxiliary space: **O(|V| + |E|)**.
 ///
 /// where **n** is the number of iterations, **|V|** the number of vertices (i.e nodes) and **|E|**
@@ -57,7 +57,7 @@ use crate::visit::{EdgeRef, IntoEdges, NodeCount, NodeIndexable};
 /// let damping_factor = 0.7_f32;
 /// let number_iterations = 10;
 /// let output_ranks = page_rank(&g, damping_factor, number_iterations);
-/// let expected_ranks = vec![0.14685437, 0.20267677, 0.22389607, 0.27971846, 0.14685437];
+/// let expected_ranks = vec![0.14685437, 0.20267676, 0.22389606, 0.27971846, 0.14685437];
 /// assert_eq!(expected_ranks, output_ranks);
 /// ```
 #[track_caller]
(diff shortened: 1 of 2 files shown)


RyanJamesStewart:
Closing this. After opening it I found the chain it conflicts with: the existing serial `page_rank` is not only O(|V|^2|E|) but mathematically incorrect (issue #960: link propagation, dangling redistribution and teleportation are chosen by an if/else, when the PageRank recurrence requires all three summed for every node and teleportation is unconditional; it diverges from networkx on a small example). #959 was a prior optimization attempt the author self-closed on finding the same correctness error, and #961 is an open fix for it.

This PR changed only the asymptotic cost while reproducing the previous output (verified to ~1e-15 against the prior implementation), so it preserves the #960 defect and would entrench it. Correctness should land before complexity, which #961 is already doing. Withdrawing in favor of #961.

The sparse O(|V|+|E|) reformulation here (the per-iteration split into a dangling-mass scalar, a teleport scalar, and one pass over distinct in-neighbours) is available for reuse on the performance half once the correctness fix lands.

PR #1001, opened 2026-07-14
fix: return zero colors for empty DSATUR graphs

### LINKED ISSUE #975: `dsatur_coloring` reports one used color for the empty graph
## Summary

`dsatur_coloring` tracks the highest assigned color in `max_color` and returns `max_color + 1` as the number of used colors.

When the input graph is empty, no node is ever colored, `max_color` stays at its sentinel value, and the function still returns `1`. And yet, an empty graph uses zero colors.

## Expected behavior

The color count should be `0` when no node was colored.

## Concrete regression test

```rust
use petgraph::{Graph, Undirected, algo::dsatur_coloring};

#[test]
fn dsatur_coloring_empty_graph_uses_zero_colors() {
    let graph: Graph<(), (), Undirected> = Graph::new_undirected();

    let (coloring, nb_colors) = dsatur_coloring(&graph);

    assert!(coloring.is_empty());
    assert_eq!(nb_colors, 0);
}
```

This currently fails because `dsatur_coloring(&graph)` returns `(empty_map, 1)` for an empty graph.

## Suggested fix

Special-case the empty graph or derive the count from whether any color assignment actually happened.

JSap0914:
## Summary
- Return zero colors when `dsatur_coloring` receives an empty graph.
- Add regression coverage for an empty undirected graph.

## Verification
- `cargo test -p petgraph --test coloring`

Fixes #975

### DIFF
--- a/crates/petgraph/src/algo/coloring.rs
+++ b/crates/petgraph/src/algo/coloring.rs
@@ -98,5 +98,7 @@ where
         }
     }
 
-    (colored, max_color + 1)
+    let nb_colors = if colored.is_empty() { 0 } else { max_color + 1 };
+
+    (colored, nb_colors)
 }
--- a/crates/petgraph/tests/coloring.rs
+++ b/crates/petgraph/tests/coloring.rs
@@ -1,6 +1,16 @@
 use petgraph::{Graph, Undirected, algo::dsatur_coloring};
 
 #[test]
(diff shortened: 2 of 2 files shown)


RaoulLuque:
Btw, if you want to make sure that this will be available in 0.8 as well, feel free to open an appropriate PR to the 0.8 branch.

PR #1011, opened 2026-08-15
test: Fix multiple small issues in `DirectedGraph` test macro

RaoulLuque:
This is a followup to #947. Fully qualified paths for calling the `DirectedGraph` traits are used instead of using the `.function_name()` syntax, to avoid ambiguity, if the to-be-tested graph also has a method called `function_name`.

Also fixes two test cases in which the graph implementation was expected to invalidate Node or Edge Ids of the Node or Edge that was added first into the graph. Instead we only expect this to be the case for the most recently added node or edge respectively.

At last, we use `.expect()` with an appropriate message instead of `.unwrap()` in the macro-generated tests, to make identifying the failure point of a test easier.


RaoulLuque:
Converted to draft since I want to add more precise messages in the test macro for when we unwrap etc. to be able to identify the exact location of a failure more easily.

PR #1016, opened 2026-08-16
fix: return zero flow when dinics source equals destination

### LINKED ISSUE #973: `dinics` does not terminate when `source == destination`
## Summary

`dinics` has no special handling for `source == destination`.

`build_level_graph` immediately labels the source as reachable, so the outer loop condition stays true, but the blocking-flow search makes no progress from that state and returns zero flow. The same state is then repeated forever.

## Current behavior

Calling `dinics(&g, s, s)` can loop indefinitely, even on a graph with a single node and no edges.

## Concrete regression test

This is small enough to add directly as a crate regression test in `crates/petgraph/tests/dinics.rs`:

```rust
use petgraph::{algo::dinics, prelude::Graph};

#[test]
fn test_dinics_source_equals_destination_regression() {
    let mut graph = Graph::<(), u32>::new();
    let source = graph.add_node(());

    let (max_flow, flows) = dinics(&graph, source, source);

    assert_eq!(0, max_flow);
    assert!(flows.is_empty());
}
```

This uses the smallest valid public input: one node, no edges, and the same node passed as both `source` and `destination`.
On the current implementation, the test hangs inside `dinics`.
With an early return for `source == destination`, it passes immediately and reports zero total flow.

## Expected behavior

The algorithm should either:

- return zero flow immediately for `source == destination`, or
- reject the input explicitly with a documented error/panic

## Suggested fix

Add an early return before the first level-graph construction for the `source == destination` case.

teddytennant:
`dinics` never returns when source and destination are the same node. `build_level_graph` labels the source as 1, the outer loop waits for the destination's level to drop to 0, and `find_blocking_flow` only succeeds when a *next* vertex is the destination, so it keeps adding zero flow forever.

Repro on `0.8`:

    let mut graph = Graph::<(), u32>::new();
    let s = graph.add_node(());
    dinics(&graph, s, s); // hangs

One-node, no edges is enough. I killed it after 15s.

Return zero flow (and a zeroed residual of `edge_count()`) when the terminals are the same node. That is the usual s-t definition, and it matches what the issue suggested.

`cargo test --all-features` is green on `0.8`. Added `test_dinics_source_equals_destination`.

Fixes #973

### DIFF
--- a/src/algo/maximum_flow/dinics.rs
+++ b/src/algo/maximum_flow/dinics.rs
@@ -79,6 +79,13 @@ where
     G: NodeCount + EdgeCount + IntoEdgesDirected + EdgeIndexable + NodeIndexable + Visitable,
     G::EdgeWeight: Sub<Output = G::EdgeWeight> + PositiveMeasure,
 {
+    if source == destination {
+        return (
+            G::EdgeWeight::zero(),
+            vec![G::EdgeWeight::zero(); network.edge_count()],
+        );
+    }
+
     let mut max_flow = G::EdgeWeight::zero();
     let mut flows = vec![G::EdgeWeight::zero(); network.edge_count()];
     let mut visited = network.visit_map();
--- a/tests/dinics.rs
+++ b/tests/dinics.rs
@@ -190,3 +190,12 @@ fn test_dinics_stable_graph() {
     let (max_flow, _) = dinics(&graph, source, sink);
     assert_eq!(5, max_flow);
 }
+
+#[test]
+fn test_dinics_source_equals_destination() {
+    let mut graph = Graph::<(), u32>::new();
+    let source = graph.add_node(());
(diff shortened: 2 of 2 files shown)


RaoulLuque:
Hi there, good catch and thanks for the fix :sauropod: Would you mind changing your PR target to `0.8` as the code you are targeting will most likely be deleted for the `0.9` release, see #891 .

teddytennant:
Retargeted to `0.8` and rewrote the branch on top of it. Same two-line guard, the file just moved from `crates/petgraph/` back to `src/`.

The hang reproduces there too, `test_dinics_source_equals_destination` sat at "running for over 60 seconds" before the fix. `cargo test --all-features` is green on the branch.

RaoulLuque:
Great, will merge this later, thanks again ☺️

PR #1035, opened 2026-09-06
fix!: Return an error instead of panicking on malformed graph6 input

### LINKED ISSUE #969: `graph6` decoder panics on malformed or truncated input
`graph6::from_graph6_representation` and the public `FromGraph6::from_graph6_string` entrypoints trust the input as if it were already structurally valid graph6. On malformed or truncated strings they panic instead of returning a recoverable error.

**Disclaimer**: this is not a security issue as there is no implication that these two `from_*` functions can or should take untrusted user inputs. This is more like a missing documentation issue. But nevertheless, it might be good to explicitly document this.

## Current behavior

The decoder performs unchecked operations on caller-controlled input:

- `bytes.first().unwrap()`
- `bytes[1..=3]` when the first decoded byte indicates the long-order form
- `adj_matrix_bits[i]` while reconstructing edges

That means malformed inputs can terminate the caller.

## Examples

The following classes of input all reach unchecked operations:

1. Empty input: `""`
2. Truncated long-order header: `"~"` (graph6 long-order prefix with the required extra header bytes missing)
3. Truncated adjacency payload: `"B"` (declares a non-empty graph but provides too few adjacency bits)

## Suggested fix

Introduce a checked parser that returns `Result<..., Graph6Error>` and validate:

- non-empty input before reading the first byte
- enough bytes for the long-order header before slicing `1..=3`
- enough adjacency bits for `n * (n - 1) / 2` before indexing

If the current infallible trait method must remain for compatibility, it can delegate to the checked parser and panic with an explicit message only at that outermost compatibility boundary.

teddytennant:
`from_graph6_representation` and `FromGraph6::from_graph6_string` treat their argument as already valid, so a malformed string kills the process instead of returning an error. There are four unchecked spots: `bytes.first().unwrap()` on an empty string, `(c as usize) - N` underflowing for any character below `'?'`, `bytes[1..=3]` when a `~` long-form header is cut short, and `adj_matrix_bits[i]` when the declared order needs more bits than the string carries. Two more inputs don't panic but decode to the wrong graph: a character above `'~'` gets masked down into a valid byte by `get_number_as_bits`, and the 8-byte `~~` order form (which the encoder never emits) is read as if it were the 4-byte `~` form.

graph6 is a file format, so a caller normally decodes lines it did not write. A truncated line taking the whole process down is a rough edge worth closing.

`get_order_bytes_and_adj_matrix_bytes` now checks each character is in `'?'..='~'` before the bias subtraction, checks the three order bytes are actually there before slicing, and rejects `~~` rather than mis-parsing it. `get_edges` computes the `n * (n - 1) / 2` bits the declared order needs and compares that against what it got before indexing anything; that product goes through `u64` so an order near the 18-bit maximum can't wrap a 32-bit `usize`.

I made the existing entry points fallible instead of adding a parallel checked parser next to them. A second `try_` API means the panicking one stays forever, and the caller who doesn't know it panics is exactly the caller who gets bitten by it. `Graph6Error` is shaped like `CsrError` and `GraphError`: a plain enum with `core::error::Error` and `Display`, so it still builds without `std`.

`cargo test --all-features`, `cargo fmt --all -- --check` on nightly, `cargo clippy --all-features --lib --bins --examples --tests -- -D warnings` and `cargo check --no-default-features -p petgraph --target wasm32v1-none --features graphmap,serde-1,stable_graph,matrix_graph,generate,unstable` all pass. The new test in `tests/graph6.rs` covers all six inputs above and every one of them panicked or decoded wrong on master.

BREAKING CHANGE: `from_graph6_representation` now returns `Result<(usize, Vec<(Ix, Ix)>), Graph6Error>` and `FromGraph6::from_graph6_string` returns `Result<Self, Graph6Error>`. Callers that know their input is well formed can add `.expect("valid graph6")`.

Closes #969

### DIFF
--- a/crates/petgraph/benches/graph6_decoder.rs
+++ b/crates/petgraph/benches/graph6_decoder.rs
@@ -33,5 +33,5 @@ fn from_graph6_str_63(bench: &mut Bencher) {
 }
 
 fn from_graph6_bench(bench: &mut Bencher, graph6_str: &str) {
-    bench.iter(|| (from_graph6_representation::<u16>(graph6_str.to_string())));
+    bench.iter(|| (from_graph6_representation::<u16>(graph6_str.to_string()).unwrap()));
 }
--- a/crates/petgraph/src/graph6/graph6_decoder.rs
+++ b/crates/petgraph/src/graph6/graph6_decoder.rs
@@ -5,6 +5,7 @@ use alloc::{
     vec,
     vec::Vec,
 };
+use core::fmt;
 #[cfg(any(feature = "graphmap", feature = "matrix_graph"))]
 use core::hash::BuildHasher;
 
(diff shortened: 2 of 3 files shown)


teddytennant:

teddytennant:
rebased onto master so Lints picks up #1037.

PR #1036, opened 2026-09-06
fix: Make steiner_tree return a tree

### LINKED ISSUE #922: Steiner tree sometimes reutrns non-tree structures
As the title says [steiner_tree](https://docs.rs/petgraph/latest/petgraph/algo/steiner_tree/index.html) sometimes returns non-tree structures. The example code runs the function until it finds a non-tree result.

```
use petgraph::{
	algo::steiner_tree,
	graph::{NodeIndex, UnGraph},
};

fn main() {
	loop {
		let edges = [(40, 13), (40, 17), (40, 26), (41, 0), (41, 1), (41, 9), (42, 6), (42, 7), (42, 11), (42, 20), (43, 3), (43, 12), (44, 7), (44, 12), (44, 21), (45, 5), (45, 6), (45, 10), (45, 19), (46, 4), (46, 8), (46, 17), (47, 15), (47, 16), (47, 20), (47, 29), (48, 0), (48, 8), (49, 22), (49, 23), (49, 27), (49, 36), (50, 24), (50, 25), (50, 29), (50, 38), (51, 31), (51, 35), (52, 1), (52, 2), (52, 10), (53, 4), (53, 5), (53, 9), (53, 18), (54, 34), (54, 39), (55, 33), (55, 34), (55, 38), (56, 13), (56, 14), (56, 18), (56, 27), (57, 16), (57, 21), (57, 30), (58, 2), (58, 3), (58, 11), (59, 25), (59, 30), (59, 39), (60, 31), (60, 32), (60, 36), (61, 32), (61, 33), (61, 37), (62, 14), (62, 15), (62, 19), (62, 28), (63, 22), (63, 26), (63, 35), (64, 23), (64, 24), (64, 28), (64, 37)];

		let mut graph = UnGraph::<(), i32>::new_undirected();
		for _ in 0..65 {
			graph.add_node(());
		}

		for edge in edges {
			graph.add_edge(NodeIndex::new(edge.0), NodeIndex::new(edge.1), 1);
		}

		let terminals = [
			NodeIndex::<u32>::new(40),
			NodeIndex::<u32>::new(41),
			NodeIndex::<u32>::new(42),
			NodeIndex::<u32>::new(43),
			NodeIndex::<u32>::new(44),
			NodeIndex::<u32>::new(45),
			NodeIndex::<u32>::new(46),
			NodeIndex::<u32>::new(56),
			NodeIndex::<u32>::new(58),
		];

		let mut maybe_tree = steiner_tree(&graph, &terminals);
		let copy = maybe_tree.clone();
		let mut nodes = 0;
		while nodes != maybe_tree.node_count() {
			nodes = maybe_tree.node_count();
			let indices: Vec<_> = maybe_tree.node_indices().collect();
			for index in indices {
				let neighbors: Vec<_> = maybe_tree.neighbors(index).collect();
				if neighbors.len() < 2 {
					maybe_tree.remove_node(index);
				}
			}
		}

		if maybe_tree.node_count() != 0 {
			println!("remaining: {maybe_tree:?}");
			println!("Stainer 'tree': {copy:?}");
			break;
		}

		println!("Next");
	}
}

```

teddytennant:
`steiner_tree` expands every edge of the metric closure's spanning tree back into the shortest path it stands for and unions those paths together. Nothing after that reduces the union back to a tree, so when two expanded paths overlap it comes out with a cycle in it. The leaf pruning can't take that back out, every node on a cycle has at least two neighbours, so the cycle survives into the returned `StableGraph`.

It takes a tie in the underlying graph, which is why the doc example never shows it. The smallest case I found is 7 nodes with every weight 1: edges a-b, a-d, b-d, c-d, c-e, d-g, f-g and terminals a, b, e, f. The metric closure there has four spanning trees of equal weight. Two of them pair e with a and f with b, and expanding those walks into d from both a and b, so the a-b-d triangle ends up whole in the result and you get 7 nodes with 7 edges. Which of the four comes out follows the iteration order of the metric closure `HashMap`, so a single call only hits it about half the time and the regression test repeats it.

Kou's algorithm has a step for exactly this, and so does the NetworkX implementation the docs point at: a minimum spanning tree over the subgraph of expanded paths, taken before the non-terminal leaves are pruned. `spanning_tree_edges` is that step. It's a plain Kruskal that hands back `EdgeIndex`es rather than `Element`s, because `min_spanning_tree` plus `from_elements` would rebuild the graph and renumber its nodes and the result currently keeps the input's node indices.

There's also a quickcheck property for the same invariant. It squashes the edge weights down to 1 and 2, since ties in the metric closure are what decides which spanning tree gets expanded.

Closes #922

### DIFF
--- a/crates/petgraph/src/algo/steiner_tree.rs
+++ b/crates/petgraph/src/algo/steiner_tree.rs
@@ -3,8 +3,6 @@ use core::{fmt::Debug, hash::Hash};
 
 use hashbrown::{HashMap, HashSet};
 
-#[cfg(feature = "stable_graph")]
-use crate::stable_graph::StableGraph;
 use crate::{
     Undirected,
     algo::{
@@ -17,6 +15,8 @@ use crate::{
         IntoNodeIdentifiers, IntoNodeReferences, NodeCompactIndexable, NodeIndexable, Visitable,
     },
 };
+#[cfg(feature = "stable_graph")]
+use crate::{graph::EdgeIndex, stable_graph::StableGraph, unionfind::UnionFind};
 
 type Edge<G> = (<G as GraphBase>::NodeId, <G as GraphBase>::NodeId);
 type Subgraph<G> = HashSet<<G as GraphBase>::NodeId>;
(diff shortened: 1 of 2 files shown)


teddytennant:

PR #1037, opened 2026-09-06
ci: Fix clippy

teddytennant:
Second half of the `Lints` fix. With #1033 in, the job gets past clippy and then dies at `cargo doc --no-deps --all-features`, which runs under `RUSTDOCFLAGS=-Dwarnings`:

```
error: redundant explicit link target
  --> crates/petgraph/src/visit/mod.rs:5:41
   |
 5 | //! Graph traits like [`IntoNeighbors`][in] create iterators and use the same
   |                        ---------------  ^^ explicit target is redundant
   |                        |
   |                        because label contains path that resolves to same destination
```

16 of those, in `visit/mod.rs`, `matrix_graph.rs`, `algo/matching.rs` and `algo/min_spanning_tree.rs`. Same story as the clippy one: nothing in the repo changed, `rustdoc::redundant_explicit_links` got stricter. The comment at `matrix_graph.rs:830` dates to June 2025 and was passing this job in January.

They are all leftovers from before intra-doc links, either ``[`MatrixGraph`][2]`` with a `[2]: struct.MatrixGraph.html` line further down, or the inline ``[`Zero`](trait.Zero.html)`` form. The label resolves on its own now, so the explicit target goes and the definition line goes with it. I only touched what rustdoc flags. The definitions whose labels don't resolve are still there and still needed, like `[1]: ../visit/trait.IntoNodeIdentifiers.html#tymethod.node_identifiers` in `matrix_graph.rs` and the `[bfs]`/`[dfspo]`/`[topo]` block at the top of `visit/mod.rs`.

Doc comments only, no code. `cargo doc --no-deps --all-features` with `-Dwarnings` is clean on `nightly-2026-09-05`, and `cargo test --all-features` passes including the 77 doc-tests.

`Lints` stays red on this PR until #1033 lands, because clippy runs before the doc step.

### DIFF
--- a/crates/petgraph/Cargo.toml
+++ b/crates/petgraph/Cargo.toml
@@ -1,7 +1,6 @@
 [package]
 name                   = "petgraph"
 version                = "0.8.3"
-readme                 = "README.md"
 rust-version.workspace = true
 license.workspace      = true
 authors.workspace      = true
--- a/crates/petgraph/src/algo/matching.rs
+++ b/crates/petgraph/src/algo/matching.rs
@@ -192,7 +192,7 @@ where
 /// guarantees about the output are given other than that it is a valid
 /// matching.
 ///
-/// If you require a maximum matching, use [`maximum_matching`][1] function
+/// If you require a maximum matching, use [`maximum_matching`] function
 /// instead.
 ///
 /// # Arguments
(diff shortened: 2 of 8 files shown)


teddytennant:
The red Lints here is the clippy half, `needless_bool` in the quickcheck test, which #1033 covers. This one only gets the `cargo doc` step.

Both are pre-existing on master and independent, so the job stays red until they've both landed.

RaoulLuque:
I just merged this with this https://github.com/petgraph/petgraph/pull/1033 and some more changes that seemed sensible to not have too many PRs ^^

Issue #164, opened 2017-09-10
Graph and StableGraph parity


bluss:
(List features that StableGraph is missing here)

- [ ] `.clear_edges()`
- (more..)

PR #1044, opened 2026-09-18
feat: add node_weights_mut, edge_weights_mut to Acyclic and Frozen

### LINKED ISSUE #1045: `node_weights_mut` on `Acyclic` and `Frozen` wrappers
I have a use case where I want to mutably borrow all the nodes in an `Acyclic<G>` at once, but currently the only way I can do this is to either:

- `graph.inner()` to recover the underlying graph, then rebuild the acyclic graph after the operation
- create a wrapper struct that contains an `Acyclic<Digraph<(), E>>` and a vector of the `N`s that I want, and keep them in sync manually.

I think it should be safe to expose this and the edge weights, because it can't change the topology of the graph. Same story with `Frozen`.

I've put together a [pull request](https://github.com/petgraph/petgraph/pull/1044) for this, but if I'm reading the contributing guidelines right, I'm also supposed to file an issue too.

HenrySwanson:
A caller who wants access to multiple simultaneous node/edge weights in an `Acyclic<G>` has no way of doing so right now, other than calling `inner()` and losing the guarantees on acyclicity. Similar story with `Frozen<G>`.

While I was adding `node_weights_mut` and `edge_weights_mut` to `Frozen`, I saw that `index_twice_mut` was implemented for `Frozen<Graph>` but not `Frozen<StableGraph>`, so I added that while I was at it.

These methods shouldn't allow the user to violate any of the invariants provided by the wrapper; both of them only allow mutable access to the weights, not the topology of the graph.

Closes [issue 1045](https://github.com/petgraph/petgraph/issues/1045).


HenrySwanson:
It would also close out one more checkbox from https://github.com/petgraph/petgraph/issues/164.

RaoulLuque:
Hi there : ) Thanks for the PR ! Please redirect your PR to the `0.8` branch, as this way the changes will take immediate effect. Also see #891 for more info on this ^^

HenrySwanson:
Whoops, I saw the warning in the README, but I interpreted it the wrong way around. Should be fixed now!

RaoulLuque:
All good, again, thanks for your work : )

RaoulLuque:
Not sure why the jobs are not running 🤔

HenrySwanson:
Yeah, I'm not sure what's up with that. Is there a way to restart them?

RaoulLuque:
I guess just push a commit and then revert that in another commit or smth maybe?

HenrySwanson:
Updated the commit message slightly and re-pushed

PR #1047, opened 2026-09-22
fix: restore free edge after failed stable graph insertion

### LINKED ISSUE #1046: `StableGraph::try_add_edge` occupies a place on the free edge list even when it fails
`StableGraph::try_add_edge` occupies a place on the free edge list via `self.free_edge = edge.next[0];` and only later checks whether the two end points actually exist. If they don't, it returns `Err`, but does not roll back the change it already did to `self.free_edge`.

Here's a reproducer of the bug:
```rust
let mut g = StableDiGraph::<&str, i32>::new();                                     
let a = g.add_node("a");                                                                                                                                  
let b = g.add_node("b");                                                                                                                                  
let e = g.add_edge(a, b, 10);                                                                                                                             
g.remove_edge(e);   // frees a slot -> reuse path                                                                                                         
g.remove_node(b);   // makes b vacant -> endpoint check will reject                                                                                       
assert!(g.try_add_edge(a, b, 99).is_err());   // correctly rejected                                                                                       
                                                                                                                                                          
assert_eq!(g.edge_count(), 0);                                                                                                                            
assert_eq!(g.edge_indices().count(), 1);      // ...but an edge exists                                                                                    
assert_eq!(g.edge_weight(EdgeIndex::new(0)), Some(&99)); 
```

sb123sb123:
Fixes https://github.com/petgraph/petgraph/issues/1046

## Problem
`StableGraph::try_add_edge` removes a slot from the free-edge list before validating both endpoint nodes. When validation fails, the slot can remain occupied by an edge that is not counted, corrupting the free list and losing the slot.

## Fix
Roll back a reused edge slot when endpoint validation returns `GraphError::NodeMissed`, and add a regression test proving the slot remains reusable.

## Tests
- `cargo test -p petgraph try_add_edge_missing_node_does_not_consume_free_edge`
- `cargo test -p petgraph stable_graph`
- `cargo fmt --all -- --check`
- `cargo clippy -p petgraph --lib --tests -- -D warnings`

AI assistance: OpenAI Codex (GPT-5.6 Luna)


RaoulLuque:
Hi there, thanks for your PR :) please redirect your PR to the 0.8 branch, see #891

